PR #6, opened 2021-07-22
Update documentation

bdlukaa:

Documentation update

There were barely no documentation on methods and classes.

## What is the new behavior?

Now, most of the methods and classes are documentated

---

The readme was also updated

### DIFF
--- a/lib/src/supabase.dart
+++ b/lib/src/supabase.dart
@@ -72,7 +72,7 @@ class Supabase {
       _instance._init(url, anonKey);
       _instance._authCallbackUrlHostname = authCallbackUrlHostname;
       _instance._debugEnable = debug ?? false;
-      _instance._localStorage = localStorage ?? LocalStorage();
+      _instance._localStorage = localStorage ?? const LocalStorage();
       _instance.log('***** Supabase init completed $_instance');
     }
 
@@ -91,7 +91,7 @@ class Supabase {
   bool _debugEnable = false;
 
   String? _authCallbackUrlHostname;
-  LocalStorage _localStorage = LocalStorage();
+  LocalStorage _localStorage = const LocalStorage();
 
   /// Dispose the instance
   void dispose() {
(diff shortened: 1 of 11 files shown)


lib/src/supabase.dart
```diff
@@ -108,12 +203,9 @@ class Supabase {
     _initialized = true;
   }
 
+  /// The [LocalStorage] instance used to persist the user session.
   LocalStorage get localStorage => _localStorage;
 
-  Future<bool> get hasAccessToken => _localStorage.hasAccessToken();
-
-  Future<String?> get accessToken => _localStorage.accessToken();
-
```
bdlukaa, line 116: These methods were removed because they were not being used

phamhieu: We should keep these helper methods. At least i used them in the flutter demo app 😄

phamhieu:
Nice work! @bdlukaa. Thank you

lib/src/supabase.dart
bdlukaa: Why?

bdlukaa: you can use `localStorage.accessToken`

phamhieu: true, but they are just more convenient.

phamhieu: FYI at the start, we didn't have localStorage class. We used `hasAccessToken` and `accessToken` methods. After introducing localStorage class, i think it's better to keep these getter methods to prevent breaking change.

bdlukaa: I'd rather remove it now instead of removing it in a future version

bdlukaa: but I'll add them back

bdlukaa: @phamhieu done!

phamhieu (approved):

Issue #11, opened 2021-07-30
Recovery session should happen automatically 


phamhieu:
# Bug report

## Describe the bug

Right now recovery session is not called under `Supabase.initialize` which is bad because users won't have access to session, user on app start

## Expected behavior

User should be able to access user and session after `Supabase.initialize` if there is persisted session info. 

## Additional context

Related https://github.com/supabase/supabase-flutter/issues/10

i-Senku:
```
auth.currentUser
auth.user()
auth.session()
```
+ The functions are confusing.
+ Instead, only the auth.session() function should be able to access the user value and session information from within this function.
+ session information should be saved with hive.

bdlukaa:
Fixed in #12

Issue #8, opened 2021-07-22
Use hive instead of shared_preferences


bdlukaa:
# Feature request

## Is your feature request related to a problem? Please describe.

Currently, the way to store the user session is made by the plugin `shared_preferences`. It is not a secure storage, making the user info vunerable.

## Describe the solution you'd like

Make use of a more secure storage solution, such as [hive](https://pub.dev/packages/hive), since it's lightweight, fast and has encryption built-in. This is the storage solution used by [firedart](https://pub.dev/packages/firedart), the pure-dart firebase implementation.

## Describe alternatives you've considered

Some other storage solutions, but they have some downsides:

- [shared_preferences](https://pub.dev/packages/shared_preferences): it's not secure, making the user info vunerable.
- [flutter_secure_storage](https://pub.dev/packages/flutter_secure_storage): it doesn't support all the platforms, it's not lightweight and is slow.
- [sqflite](https://pub.dev/packages/sqflite): it's not fast and doesn't support all the platforms.

With all the downsides presented above, hive is the best storage solution to store the user session: it's fast, lightweight and has built-in encryption.

dshukertjr:
Hi @bdlukaa, 

You raise a good point about hive being fast according to its bench mark test. 
One question, and I do apologize if I am wrong, but doesn't hive also require additional secure storage if you want to make it secure?
https://docs.hivedb.dev/#/advanced/encrypted_box?id=encrypted-box

bdlukaa:
@dshukertjr it does, but it shouldn't be a big deal.

Flutter apps are encrypted by default (and it's very hard to reverse engineer it) and the key will likely not leak, specially if the app is obfuscated using the `--obfuscate` param is used when building the app. See https://flutter.dev/docs/deployment/obfuscate#obfuscating-your-app for more info

bdlukaa:
Note that the key, once defined, can't be changed. If so, it'll result in wrong user information

PR #12, opened 2021-07-30
Move Auth implementation to a separated class

bdlukaa:

Feature and bug fix

Currently, the user session is not recovered from storage when the app is initialized.

## What is the new behavior?

Fetch the user session from storage when recovered

Fix for #11

## Additional context

This pull request also separates the auth logic into a separate singleton (`SupabaseAuth`), in order to make things easier for future PRs (for #8)

### DIFF
--- a/lib/src/supabase.dart
+++ b/lib/src/supabase.dart
@@ -1,138 +1,7 @@
 import 'package:flutter/foundation.dart';
-import 'package:shared_preferences/shared_preferences.dart';
 import 'package:supabase/supabase.dart';
-import 'package:url_launcher/url_launcher.dart';
 
-const supabasePersistSessionKey = 'SUPABASE_PERSIST_SESSION_KEY';
-
-Future<bool> _defHasAccessToken() async {
-  final prefs = await SharedPreferences.getInstance();
-  final exist = prefs.containsKey(supabasePersistSessionKey);
-  return exist;
-}
-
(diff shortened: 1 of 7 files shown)


bdlukaa:
@phamhieu can you review it, please?

lib/src/supabase.dart
```diff
@@ -179,10 +48,13 @@ class Supabase {
     );
     if (url != null && anonKey != null) {
       _instance._init(url, anonKey);
-      _instance._authCallbackUrlHostname = authCallbackUrlHostname;
       _instance._debugEnable = debug ?? kDebugMode;
-      _instance._localStorage = localStorage ?? const LocalStorage();
       _instance.log('***** Supabase init completed $_instance');
+
+      SupabaseAuth.initialize(
```
phamhieu, line 54: Do you think without await here, it can cause problem when user try to access session/user after calling `Supabase.initialize(...)`

phamhieu (commented):
Add a comment about await for `Supabase.initialize`. Other than that lgtm.

Great PR @bdlukaa :tada::tada::tada:. Thank you.

lib/src/supabase.dart
bdlukaa: indeed

bdlukaa: @phamhieu done

phamhieu (approved):

PR #13, opened 2021-08-03
Rework the LocalStorage implementation

bdlukaa:

`shared_preferences` is currently used to store the user session

## What is the new behavior?

`hive` is now used.

## Aditional information

Fix for #8

Reworked the LocalStorage implementation. `LocalStorage` is now an abstract class that can be extended by others. Some classes are already built-in: 

- `HiveLocalStorage`, that implements Hive as the storage solution to persist the session on the device.
- `EmptyLocalStorage`, that disabled persistence.

An example of how to extend LocalStorage:

```dart
class EmptyLocalStorage extends LocalStorage {
  /// Creates a [LocalStorage] instance that disables persistance
  const EmptyLocalStorage()
      : super(
          initialize: _initialize,
          hasAccessToken: _hasAccessToken,
          accessToken: _accessToken,
          removePersistedSession: _removePersistedSession,
          persistSession: _persistSession,
        );

  static Future<void> _initialize() async {}
  static Future<bool> _hasAccessToken() => Future.value(false);
  static Future<String?> _accessToken() => Future.value(null);
  static Future<void> _removePersistedSession() async {}
  static Future<void> _persistSession(_) async {}
}
```

I also added a new function to `LocalStorage`: `initialize`. There you can initialize anything you need before performing any search operation. On the `HiveLocalStorage` implementation, it is used to open the box:

```dart
  static Future<void> _initialize() async {
    HiveCipher? encryptionCipher;
    if (encryptionKey != null) {
      encryptionCipher = HiveAesCipher(base64Url.decode(encryptionKey!));
    }
    await Hive.initFlutter('auth');
    await Hive.openBox(_hiveBoxName, encryptionCipher: encryptionCipher);
  }
```

It could be useful to improve fetching time for a lot of packages. We can take `shared_preferences` as an example:

```dart

late SharedPreferences _sharedPreferencesInstance;

class SharedPreferencesLocalStorage extends LocalStorage {
  const SharedPreferencesLocalStorage ()
      : super(
          initialize: _initialize,
          hasAccessToken: _hasAccessToken,
          accessToken: _accessToken,
          removePersistedSession: _removePersistedSession,
          persistSession: _persistSession,
        );

  static Future<void> _initialize() async {
    _sharedPreferencesInstance = await SharedPreferences.getInstance();
  }
  static Future<bool> _hasAccessToken() async {
    return _sharedPreferencesInstance.containsKey(...);
  }
  static Future<String?> _accessToken() {...}
  static Future<void> _removePersistedSession() async {...}
  static Future<void> _persistSession(_) async {...}
}
```

PS.: this was NOT tested in any app in production.

### DIFF
--- a/lib/src/local_storage.dart
+++ b/lib/src/local_storage.dart
@@ -71,7 +71,7 @@ class HiveLocalStorage extends LocalStorage {
         );
 
   /// The encryption key used by Hive. If null, the box is not encrypted
-  /// 
+  ///
   /// This value should not be redefined in runtime, otherwise the user may
   /// not be fetched correctly
   ///
--- a/lib/src/local_storage.dart
+++ b/lib/src/local_storage.dart
@@ -32,36 +32,6 @@ Future<void> _defPersistSession(String persistSessionString) {
 /// By default, the package `hive` is used to save the user info on
 /// the device. However, you can use any other plugin to do so.
 ///
-/// For example, we can use `flutter_secure_storage` plugin to store
-/// user session in a secure storage.
-///
-/// ```dart
-/// final localStorage = LocalStorage(
-///   hasAccessToken: () {
-///     const storage = FlutterSecureStorage();
-///     return storage.containsKey(key: supabasePersistSessionKey);
-///   }, accessToken: () {
-///     const storage = FlutterSecureStorage();
-///     return storage.read(key: supabasePersistSessionKey);
-///   }, removePersistedSession: () {
-///     const storage = FlutterSecureStorage();
-///     return storage.delete(key: supabasePersistSessionKey);
-///   }, persistSession: (String value) {
-///     const storage = FlutterSecureStorage();
-///     return storage.write(key: supabasePersistSessionKey, value: value);
-///   });
-/// ```
-///
-/// To use the `LocalStorage` instance, pass it to `localStorage` when initializing
-/// the [Supabase] instance:
-///
-/// ```dart
-/// Supabase.initialize(
-///  ...
-///  localStorage: localStorage,
-/// );
-/// ```
-///
 /// See also:
 ///
 ///   * [SupabaseAuth], the instance used to manage authentication
--- a/lib/src/local_storage.dart
+++ b/lib/src/local_storage.dart
@@ -1,3 +1,5 @@
+import 'dart:convert';
+
 import 'package:hive_flutter/hive_flutter.dart';
 
 const _hiveBoxName = 'supabase_authentication';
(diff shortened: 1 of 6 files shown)


lib/src/supabase.dart
```diff
@@ -52,7 +53,7 @@ class Supabase {
       _instance.log('***** Supabase init completed $_instance');
 
       await SupabaseAuth.initialize(
-        localStorage: localStorage ?? const LocalStorage(),
+        localStorage: localStorage ?? const HiveLocalStorage(),
```
atnegrete, line 56: I haven't got much dart 2.0 experience, but could the default `?? const HiveLocalStorage()` be moved up to the static initialize method parameter? 

My thought process here is, am I able to disable persistence or will it be forced with this package?

atnegrete:
This is cool beans

lib/src/supabase.dart
bdlukaa: you can create a custom LocalStorage implementation that does nothing. That way, it'll never persist the session. It's a great idea to have it implemented by default on the package

bdlukaa: I added the `EmptyLocalStorage` class to disable session persistance

lib/src/local_storage.dart
```diff
@@ -0,0 +1,110 @@
...
+
+  /// The encryption key used by Hive. If null, the box is not encrypted
+  ///
+  /// This value should not be redefined in runtime, otherwise the user may
+  /// not be fetched correctly
+  ///
+  /// See also:
+  ///
+  ///   * <https://docs.hivedb.dev/#/advanced/encrypted_box?id=encrypted-box>
+  static String? encryptionKey;
```
bdlukaa, line 81: How to add a encryption key?

```dart
HiveLocalStorage.encryptionKey = MY_ENCRYPTION_KEY;
```

See <https://docs.hivedb.dev/#/advanced/encrypted_box?id=encrypted-box> for more info on how to store your encryption key

phamhieu:
I love the idea of `abstract LocalStorage class` and  `EmptyLocalStorage`. Nice work @bdlukaa 🚀🚀

> It could be useful to improve fetching time for a lot of packages

What do you mean `fetching time` here?

phamhieu (approved):

bdlukaa:
> What do you mean fetching time here?

On shared preferences, the instance is requested at every function. Using initialize, we store the instance in a variable and use it when necessary

dshukertjr (approved):
Amazing work 🎉 LGTM!

bdlukaa:
I'll merge after I do some testings

bdlukaa:
Testings done! Everything works as expected

PR #19, opened 2021-08-28
Null safety

### LINKED ISSUE #18: Migrate To Null Safety
# Need migrate to null safety

## Is your feature request related to a problem? Please describe.

Now Flutter new project by default using with null safety. To use it in a broad manner with a large audience, we need to migrate it to null safety.

## Describe the solution you'd like

Anybody can fork it, migrate to the null safety, and can send a pull request

## Describe alternatives you've considered

Mentioned Above

## Additional context

Null safety addition will be a good addition for current flutter projects

bdlukaa:

Bug fix (kinda)

---

Fixes #18

See https://dart.dev/go/null-safety-migration for more info

It all is already migrated to null-safety, no idea why it doesn't show on pub.dev.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -5,7 +5,7 @@ homepage: 'https://supabase.io'
 repository: 'https://github.com/supabase/supabase-flutter'
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: '>=2.13.0 <3.0.0'
   flutter: '>=1.17.0'
 
 dependencies:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: supabase_flutter
 description: Flutter integration for Supabase. This package makes it simple for developers to build secure and scalable products.
-version: 0.2.4
+version: 0.3.0
 homepage: 'https://supabase.io'
 repository: 'https://github.com/supabase/supabase-flutter'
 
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -1,11 +1,11 @@
...
 description: Flutter integration for Supabase. This package makes it simple for developers to build secure and scalable products.
-version: 0.2.4
+version: 0.3.0
 homepage: 'https://supabase.io'
 repository: 'https://github.com/supabase/supabase-flutter'
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: '>=2.13.0 <3.0.0'
   flutter: '>=1.17.0'
```
dshukertjr, line 9: @bdlukaa 
Yeah, this is a weird one. I did some research as well, but couldn't find what exactly gets you the `null-safety` badge. 

I noticed that [hive_flutter](https://pub.dev/packages/hive_flutter) also does not have the `null-safety` badge either even though the [current version](https://pub.dev/packages/hive_flutter/versions) is null safe. 

I was wondering if this line is the cause. We can get rid of this line, if I am correct, right?

dshukertjr:
With the [comment that I left on the related issue](https://github.com/supabase/supabase-flutter/issues/18#issuecomment-907589357), I think we can just wait for now and ignore the issue, but what do you think? It seems like there is not much we can do about it.

PR #22, opened 2021-09-14
chore: update supabase_dart version

dshukertjr:

This PR will update `supabase_dart` to v0.2.6, which includes the X-Client-Info header. 

Some of the small changes such as adding commas or reordering imports are done because of the new lint rule in dart 2.14.0. 

Closes https://github.com/supabase/supabase-dart/issues/45

`supabase_flutter` does not override `supabase_dart` header

## What is the new behavior?

`supabase_flutter` will override `supabase_dart` header

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [0.2.7]
+
+- chore: added `X-Client-Info` header
+
 ## [0.2.6]
 
 - fix: export local_storage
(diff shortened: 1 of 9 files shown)


bdlukaa:
I don't think this is necessary

bdlukaa:
We already have it on supabase_dart. There's no point on doing it here. It'll only complicate our lives in the future.

Just upgrade the supabase-dart version

dshukertjr:
@bdlukaa Thanks! Makes sense! Let me revert that change!

bdlukaa (approved):
LGTM!

Issue #30, opened 2021-10-03
Not able to import CountOptions enum


hareendranmg:
# Bug report

## Describe the bug

Not able to import CountOptions enum.
When clicking the Quick Fix in VS Code, it only showing create class, local variable and mixin.
It is not showing the library to import.  

## To Reproduce

```
await supabase
          .from(tableName)
          .select()
          .eq(column, value)
          .execute(count: CountOptions.exact);
```

## Screenshots
![image](https://user-images.githubusercontent.com/19310197/135762457-a777ec1d-f33a-4b82-82a3-ba7fbb68572b.png)
![image](https://user-images.githubusercontent.com/19310197/135762462-77f6acf3-a62f-45a4-91d9-243492429080.png)

## System information

- OS: Ubuntu 20.04
- Version of supabase_flutter: 0.2.8
- Version of flutter: 2.5.1

dshukertjr:
@hareendranmg This should now be fixed in the latest version of `supabase_flutter`!

PR #31, opened 2021-10-04
fix: export supabase package so that underlying symbols can be imported within a flutter project

dshukertjr:

Fixes a bug where users of this package cannot import underlying classes or Enums from `supabase` or `postgrest-dart` packages. 

Related https://github.com/supabase/supabase-flutter/issues/30

Users of this package cannot import underlying classes from underlying packages. 

## What is the new behavior?

Users can import classes or Enums from underlying packages. 

## Additional context

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [0.2.9]
+
+- fix: export supabase package
+
 ## [0.2.8]
 
 - chore: update supabase to v0.2.7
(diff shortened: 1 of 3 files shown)


dshukertjr:
@bdlukaa Do you think this would be a viable solution to [this](https://github.com/supabase/supabase-flutter/issues/30) issue?

bdlukaa:
Yes

bdlukaa (approved):

PR #32, opened 2021-10-05
Updated README example to reflect breaking change from v0.0.3. 

seenickcode:
If users do not 'await' 'Initialize', a Hive box exception will be thrown.

Docs update.

If users do not await initialize, an exception will be thrown.

## What is the new behavior?

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ import 'package:supabase_flutter/supabase_flutter.dart';
 void main() {
   WidgetsFlutterBinding.ensureInitialized();
 
-  Supabase.initialize(
+  await Supabase.initialize(
     url: SUPABASE_URL,
     anonKey: SUPABASE_ANNON_KEY,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ import 'package:supabase_flutter/supabase_flutter.dart';
 void main() {
```
bdlukaa, line 40:
```suggestion
void main() async {
```

seenickcode: @bdlukaa good catch, thanks. What about omitting `void` here for brevity as it is unnecessary? (Tho I'm not familiar with the conventions the repo owners may want to follow.)

bdlukaa: keep the `void` type. It's useful to make sure nothing is returned to `main`

seenickcode:
+1

bdlukaa (approved):
LGTM

PR #66, opened 2022-03-08
chore: update packages

DanMossa:
* Fix typo.
* Update hive for optimizations
* Update supabase for OIDC auth
* Update url_launcher for optimizations

Also, this new version of Hive has added a flush method to help with memory leaks. It shouldn't be run after every write due to performance issues. But we should eventually put a counter so that after X writes we flush.

### DIFF
--- a/lib/src/local_storage.dart
+++ b/lib/src/local_storage.dart
@@ -10,7 +10,7 @@ const supabasePersistSessionKey = 'SUPABASE_PERSIST_SESSION_KEY';
 /// See also:
 ///
 ///   * [SupabaseAuth], the instance used to manage authentication
-///   * [EmptyLocalStorage], used to disable session persistance
+///   * [EmptyLocalStorage], used to disable session persistence
 ///   * [HiveLocalStorage], that implements Hive as storage method
 abstract class LocalStorage {
   const LocalStorage({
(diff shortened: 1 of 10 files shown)


DanMossa:
@dshukertjr @bdlukaa

bdlukaa (approved):

dshukertjr:
@DanMossa 
It seems like the tests are not running because of some bug in `platform` package. 

We might need to add 
```yml
dependency_overrides:
  platform: ^3.1.0
```
to `pubspec.yml`. 

Related issue: https://github.com/flutter/flutter/issues/96711
Related Stack Overflow question: https://stackoverflow.com/questions/70363918/error-member-not-found-packageroot-how-to-solve-ignore-deprecated-member-u/70757512#comment125088312_70757512

dshukertjr:
@DanMossa 
Thanks for the quick fix! I can run it!

dshukertjr (commented):
@DanMossa It seems like we got some analysis errors as well. I think it would be great if we could get rid of them, but what do you think?

DanMossa:
@dshukertjr 
Yeah I agree! I went ahead and fixed them as well!

dshukertjr:
Hey @DanMossa , it seems like there are some files that were not formatted correctly. Could you take a look?

dshukertjr (approved):
LGTM! Amazing work!

PR #97, opened 2022-04-23
Change Hive.initFlutter to Hive.init

anggoran:
### Current Behavior
Using Hive.initFlutter to open directory named auth

### New Behavior
Using Hive.init to open directory like how `hydrated_bloc` do as a flutter-compatible package

### DIFF
--- a/lib/src/local_storage.dart
+++ b/lib/src/local_storage.dart
@@ -1,6 +1,6 @@
 import 'dart:convert';
 
-import 'package:hive_flutter/hive_flutter.dart';
+import 'package:hive/hive.dart';
 
 const _hiveBoxName = 'supabase_authentication';
 const supabasePersistSessionKey = 'SUPABASE_PERSIST_SESSION_KEY';
@@ -85,7 +85,7 @@ class HiveLocalStorage extends LocalStorage {
     if (encryptionKey != null) {
       encryptionCipher = HiveAesCipher(base64Url.decode(encryptionKey!));
     }
-    await Hive.initFlutter('auth');
+    Hive.init('auth');
     await Hive.openBox(_hiveBoxName, encryptionCipher: encryptionCipher);
   }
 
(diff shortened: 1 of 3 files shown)


bdlukaa:
Why? Since this is supposed to be running in a flutter environment, `initFlutter` seems to be the best option

anggoran:
I think `hive` is enough, since we already await `Supabase.initialize`, which means doesn't need to do async init again inside the method (from what I know, the difference between `hive` and `hive_flutter` is `hive_flutter` intended to be used to CRUD huge data, so more suited in direct app base code not package, cmiiw). Based on available packages on pub dev, there also aren't many packages that use `hive_flutter` compared to `hive`

DanMossa:
I think it's worth finding out the actually differences before changing the deps

anggoran:
The differences of `init` and `initFlutter` looks like lie on asynchronous method and subDir. I don't think subDir is needed, because if people import `hive` and `supabase_flutter` on their projects, they will set different homeDir and subDir on `Hive.initFlutter()`, so `Hive.init('auth')` seems to be fine
[Hive.initFlutter() line 8](https://github.com/hivedb/hive/blob/f370eaa76ddb383f4680e5cb9d9a7010eb608763/hive_flutter/lib/src/hive_extensions.dart) 
[Hive.init() line 8](https://github.com/hivedb/hive/blob/ae4e0a891204900f00bcfcf0ccb866cb968f90ba/hive/lib/src/hive.dart) 
[HiveImpl init line 57 (base code)](https://github.com/hivedb/hive/blob/c328cedd0ed2f84e4c334892a9fd3c37d4b52103/hive/lib/src/hive_impl.dart) 

[Hive.init() Implementation in hydrated_bloc package](https://github.com/felangel/bloc/blob/774d1c1733acd9a01b544b6f79813518361c0169/packages/hydrated_bloc/lib/src/hydrated_storage.dart)
(The instantiating of HiveImpl again seems to be not needed, since it is already declared as a global constant: [HiveImpl() as Hive](https://github.com/hivedb/hive/blob/ae4e0a891204900f00bcfcf0ccb866cb968f90ba/hive/lib/hive.dart))

PR #102, opened 2022-04-27
fix: added retries when session recovery fails due to network failure

dshukertjr:
This PR adds automatic retries of session recovery when session recovery fails due to network request failure. 

Fixes https://github.com/supabase-community/supabase-flutter/issues/33

* This PR is waiting for https://github.com/supabase-community/gotrue-dart/pull/71 to be released as well as subsequent supabase-dart release.

### DIFF
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -79,16 +79,25 @@ class SupabaseAuth {
     if (hasPersistedSession) {
       final persistedSession = await _instance._localStorage.accessToken();
       if (persistedSession != null) {
-        final response = await Supabase.instance.client.auth
-            .recoverSession(persistedSession);
-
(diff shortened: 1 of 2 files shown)


bdlukaa:
Wouldnt this be handled by the `gotrue_dart` package?

dshukertjr:
@bdlukaa 
Right. I feel like I have screwed up here. 

I was thinking since as of right now, `recoverSession()` method in gotrue-dart does not return a valid `GoTrueSessionResponse`, we needed to retry on supabase-flutter as well, but I think it's best to make gotrue-dart return a valid session response upon successful retry. 

 I have started another PR in gotrue-dart for this. Would be amazing to hear what you think about this route!
https://github.com/supabase-community/gotrue-dart/pull/75

Issue #112, opened 2022-05-15
Using incorrect version of url_launcher


DanMossa:
Hello!

I've updated my supabase_flutter package to `supabase_flutter: ^0.3.1`

Now when I go to run my app I get the following error:

```
The method 'launchUrl' isn't defined for the class 'GoTrueClient'.
 - 'GoTrueClient' is from 'package:gotrue/src/gotrue_client.dart' ('../../../../.pub-cache/hosted/pub.dartlang.org/gotrue-0.2.2/lib/src/gotrue_client.dart').
Try correcting the name to the name of an existing method, or defining a method named 'launchUrl'.
    final result = await launchUrl(url, webOnlyWindowName: '_self');
                         ^^^^^^^^^
```

And then inside of `lib/src/supabase_auth.dart`, `launchUrl` errors because it does not exist.

The current `pubspec.yaml` shows: `  url_launcher: ^6.0.10`
And this version of url_launcher does not have `launchUrl`.

DanMossa:
Fix in #113

dshukertjr:
Closing this as the fix has been merged.

PR #113, opened 2022-05-15
Update constraints and packages

DanMossa:
If we're going to be using new flutter versions then we need to update our constraints.
We should also not be using the latest stable version of Flutter for our tests. They should be a specific version that is within our constraints.

I was dealing with weird issues for a while.

Fixes:
#112 
#114 

@dshukertjr


DanMossa:
Also the tests run using stable.

They should be changed to using a specific version :/

dshukertjr (approved):
LGTM! Thanks for the quick fix!

PR #124, opened 2022-06-01
BREAKING: Remove SupabaseAuthRequiredState as well as overriding methods in SupabaseAuthState

dshukertjr:
## Comments are welcome!

This is just a proposal in an attempt to improve developer experience of this library.

## Current behavior
Currently, there are `SupabaseAuthState` and `SupabaseAuthRequiredState`. A lot of their functionalities overlap, and having two different classes might be confusing. 

Also, `SupabaseAuthState` have the following methods, but users don't necessary need all of them if they are using `SupabaseAuthState` just for auth state persistance.

```dart
  /// Callback when deeplink received and is processing. Optional
  void onReceivedAuthDeeplink(Uri uri) {
    Supabase.instance.log('onReceivedAuthDeeplink uri: $uri');
  }

  /// Callback when user is unauthenticated
  void onUnauthenticated();

  /// Callback when user is authenticated
  void onAuthenticated(Session session);

  /// Callback when authentication deeplink is recovery password type
  void onPasswordRecovery(Session session);

  /// Callback when recovering session from authentication deeplink throws error
  void onErrorAuthenticating(String message);
```

## New behavior

All of the classes that were inheriting `State` has been removed, and deeplinking and auth state persistence logic has been consolidated into `SupabaseAuth` class.  

Instead of relying on `onAuthenticated()` and `onUnauthenticated()`, users can utilize `Supabase.instance.client.auth.onAuthStateChange()` to navigate users to different routes depending on Auth state.


dshukertjr:
Also thinking that we could handle `startAuthObserver()` in a smarter way as well.

bdlukaa:
tests are failing

dshukertjr:
@bdlukaa I haven't changed the tests yet, but I wanted to see what you and other people think about the general direction of this PR.

dshukertjr:
@DanMossa @Vinzent03 
It would be great if I could get some feedbacks on the general direction of this PR. Would you agree that this is a good change?

Vinzent03:
I'm currently not using this library in my Flutter app. Instead I'm using supabase-dart. I don't know exactly why I didn't choose the flutter version, but I think the weird state classes were one reason. Using the listener directly is a good idea. To understand the new class a bit more:
Should I place the class multiple times in my widget tree or just one time at the top?

bdlukaa:
Imo we should get rid of all these classes and let the dev do it

DanMossa:
I think helper classes like this are only useful if they're basically required to be used.

Like, if 99% of people are going to use `onUnauthenticated` then that's awesome, but if there's other alternatives then I think it gets confusing.

My personal opinion is there should be 1 good way of doing something or else there's too much concern on which way the user should be doing it.

dshukertjr:
@Vinzent03 That is funny, because I do the same thing lol I use only supabase-dart in my app. 

>Should I place the class multiple times in my widget tree or just one time at the top?

Yeah, the currently the idea is to have only one of them alive in a widget tree at once. 

@bdlukaa 
>Imo we should get rid of all these classes and let the dev do it

I agree. Having these classes are counter intuitive for Flutter devs and causes confusion. One thing that I would want is automatic token refreshing though. It would be nice to provide a way where developers don't have to think about refreshing tokens themselves, and in order to do that, we need to listen to app's lifecycle. The way we do it currently is not the most elegant way of doing it, but do you think the developer should manage token refreshing by themselves too? Or should we explore better way of automatically handling token refresh without creating an extended class from `State`?

@DanMossa 
100% agree with what you are saying here. To me, this class that we provide is only good for automatically refreshing the auth token as well as handling deep links, but as I wrote earlier, I think we might be able to come up with a better way of handling automating token refresh without creating a class that extends `State`. 


@Vinzent03 @bdlukaa @DanMossa 
I think these are the questions that we want to ask ourselves.

- ### Should we provide a way to handle deeplinks and automatic token refresh, or should each developers take care of them on their own?
-> In my opinion, not having to think about token refresh is a nice developer experience, and quite frankly something that developers expect especially if they come from Firebase world, but open to suggestions. 
- ### If we are going to provide a way to automatically refresh token and handle deeplinks, how should we do it?
-> I think with this PR, all developers have to do is to have single `SupabaseAuthState` alive somewhere within their widget tree, and that is all they need to do, so it is considerably simpler than the current implementation, but ideal world would be if the user installs `supabase-flutter` package and the auth state is persisted with no additional configuration. I will do some research on how we could possibly do this, but would any of you have any suggestions on how that could be achieved? I think as long as there is a way of listening to app's lifecycle changes, we can automatically perform all the token refresh and stuff.

Vinzent03:
@dshukertjr I think automatic token refreshing is really important. Callbacks for `onAuthenticated` or `onErrorAuthenticating` should be removed and not handled in the widget tree.
Listening to `WidgetsBindingObserver` does not directly need context or a widget, so it could be integrated in the flutter client without an extra widget. Disposing and initializing must be handled manually then.

dshukertjr:
@Vinzent03 I did not know that WidgetsBindingObserver can be used without it being in the widget tree! That's great! 

Rewriting the PR as soon as I get back in front of a keyboard!

Vinzent03:
@dshukertjr I haven't tested it, but following this [Stack Overflow answer](https://stackoverflow.com/questions/61449993/flutter-how-to-track-app-lifecycle-without-a-widget) it should work.

Vinzent03:
@dshukertjr Looks great, but there is still `SupabaseAuthState` for deep linking, right? What about moving this out of the widget tree too?

dshukertjr:
@Vinzent03 Thanks for the quick checkup! Yup, will move the deep link stuff out of the widget tree too!

dshukertjr:
Actually testing out deeplink on Android and ios now...

lib/src/supabase_auth.dart
```diff
@@ -88,6 +96,8 @@ class SupabaseAuth {
         }
       }
     }
+    WidgetsBinding.instance?.addObserver(_instance);
+    _instance._startDeeplinkObserver();
```
dshukertjr, line 100: Added `_startDeeplinkObserver` here so that the user of this package don't explicitly have to call to start deep link observer. I think it shouldn't have much of an effect for those that do not use deeplinks, but what does everyone think about this?

dshukertjr:
I think most of the rework is done. 

One thing that I would want to add is a way to know when the initial authentication check is complete. 

In an app, it is pretty common to show different things to the users depending on whether the user is logged in or not. Currently with this rework, there is no way of knowing when the initial auth check has been complete. If the user is logged in, `onAuthStateChange` will run when the token refresh is complete, but if the user is not logged in, nothing will happen. There is no way to tell if the user is not signed in, or is just waiting on token refresh. 

What would be the best way to give developers a way to know when the initial  auth state? One idea that came to my mind was to have a completer name something like `hasReceivedInitialAuthState` or something, and users can await for the future of the completer, but I'm sure all of you have better ideas, so would love to hear them!

Vinzent03 (commented):
@dshukertjr Good point. I think using a `Completer` is a great way. I'm unsure about the naming, but I don't really have a better name. Is there any way the dev gets the error in case of a fail?

lib/src/supabase_auth.dart
```diff
@@ -96,6 +106,48 @@ class SupabaseAuth {
...
+    }
+
+    final String? jsonStr =
+        await SupabaseAuth.instance.localStorage.accessToken();
+    if (jsonStr == null) {
+      return false;
+    }
+
+    final response =
+        await Supabase.instance.client.auth.recoverSession(jsonStr);
```
DanMossa, line 144: Do we want to recover a session, (Which I think is refreshing the accessToken), on every state resume? Or only if the user is currently not logged in / if the access token is expired?

lib/src/supabase_auth.dart
```diff
@@ -151,6 +203,88 @@ class SupabaseAuth {
...
+
+  /// Handle the initial Uri - the one the app was started with
+  ///
+  /// **ATTENTION**: `getInitialLink`/`getInitialUri` should be handled
+  /// ONLY ONCE in your app's lifetime, since it is not meant to change
+  /// throughout your app's life.
+  ///
+  /// We handle all exceptions, since it is called from initState.
+  Future<void> _handleInitialUri() async {
+    if (!shouldHandleInitialDeeplink) return;
```
DanMossa, line 249: `shouldHandleInitialDeeplink` is a getter that checks to see the value of another variable, `_initialDeeplinkIsHandled`.

I think we can probably just use 1 variable for this.
Like, we can just check `_initialDeeplinkIsHandled` or set `shouldHandleInitialDeeplink` to true/false.

Is there a specific reason we wanted to separate them?

DanMossa (changes requested):

lib/src/supabase_auth.dart
```diff
@@ -151,6 +203,88 @@ class SupabaseAuth {
...
+    _handleIncomingLinks();
+    _handleInitialUri();
+  }
+
+  /// Stop deep link observer
+  ///
+  /// Automatically called on dispose().
+  void _stopDeeplinkObserver() {
+    Supabase.instance.log('***** SupabaseDeepLinkingMixin stopAuthObserver');
+    if (_deeplinkSubscription != null) _deeplinkSubscription?.cancel();
```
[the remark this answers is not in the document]
dshukertjr: Yeah, `hasReceivedInitialAuthState` is a pretty bad name 😂

Thanks for pointing out this `if` statement. Will fix it!

lib/src/supabase_auth.dart
dshukertjr: >I'd want _instance.isAuthCallbackDeeplink to be better at checking to see if it's actually a url that we want to recover a session from.

This is a great point! Will edit the logic to a more robust way of checking. 

>Do we want to recover a session, (Which I think is refreshing the accessToken), on every state resume? Or only if the user is currently not logged in / if the access token is expired?

For this one, I think we do want to recover session on every state resume (every time the app comes back to foreground from background), because the timer set to recover session stops when the app is in background, so we want to get a new token every time we come back to foreground, but do you think this can be optimized?

dshukertjr: About the session recovery on every time app coming back to foreground, I believe the js sdk has something similar implemented, and generally token refresh is not an expensive thing to do, so it's okay to over do it!

dshukertjr: @DanMossa 
>I'd want _instance.isAuthCallbackDeeplink to be better at checking to see if it's actually a url that we want to recover a session from.

Did you have any particular idea of what this could look like? I was going to implement a check to see if `#` is included in the URL or something, but started thinking that might not be a very good idea if the Gotrue server decided not to use `#` in the future. 

Also, even if we pass deeplinks that are actually not Supabase auth related, it will be rejected [here](https://github.com/supabase-community/gotrue-dart/blob/main/lib/src/gotrue_client.dart#L230), so right now, I am leaning towards leaving the way it is, but what do you think?

lib/src/supabase_auth.dart
dshukertjr: @DanMossa Great point! Fixing this now!

lib/src/supabase.dart
```diff
@@ -39,8 +39,8 @@ class Supabase {
   /// This must be called only once. If called more than once, an
   /// [AssertionError] is thrown
   static Future<Supabase> initialize({
-    String? url,
-    String? anonKey,
+    required String url,
+    required String anonKey,
```
dshukertjr, line 43: Just want to double check, but these are required, right?

lib/src/supabase_auth.dart
```diff
@@ -21,12 +27,23 @@ class SupabaseAuth {
   /// {@macro supabase.localstorage.accessToken}
   Future<String?> get accessToken => _localStorage.accessToken();
 
+  /// Returns when the initial session recovery is done.
+  /// Can be used to determine whether a user is signed in or not uplon
+  /// initial app launch.
+  Future<Session?> get initialSession => _initialSessionCompleter.future;
+  final Completer<Session?> _initialSessionCompleter = Completer();
```
dshukertjr, line 34: Added a completer that will complete upon initial session recovery with a session if it existed. 

Developers can await for `SupabaseAuth.instance.initialSession` to get the initial auth state of the signed in user. What do all of you think of this new variable?

Vinzent03: I like that name. But I would prefer returning a wrapper around `Session` to be able to return an error. Currently, you can't defer between a failed recovery and a missing persisted session.

lib/src/supabase.dart
Vinzent03: I think so. At least they are required in the js client: https://supabase.com/docs/reference/javascript/initializing

lib/src/supabase_auth.dart
dshukertjr: @Vinzent03 Great point! Let me change it to `Completer<GotrueSessionResponse>` right now!

dshukertjr: @Vinzent03 

I actually ended up not going with `Completer<GotrueSessionResponse>`, but keeping the type as is and using `completeError` when error occured. 

@bdlukaa has been working on amazing PRs that will throw error instead of returning them when error occurs, so wanted to align with it.

Vinzent03: Right, I remember the other ongoing reworks. Great.

lib/src/supabase_auth.dart
```diff
@@ -21,12 +27,23 @@ class SupabaseAuth {
   /// {@macro supabase.localstorage.accessToken}
   Future<String?> get accessToken => _localStorage.accessToken();
 
+  /// Returns when the initial session recovery is done.
+  /// Can be used to determine whether a user is signed in or not uplon
```
Vinzent03, line 31: Just a typo: `uplon` -> `upon`

dshukertjr: Thanks for this! I just fixed it!

dshukertjr:
I think this PR is pretty much complete on my end. Idealy, we will merge https://github.com/supabase-community/gotrue-dart/pull/74 and https://github.com/supabase-community/postgrest-dart/pull/69 and bump the major version of this package to release all of these breaking changes!

bdlukaa (changes requested):

lib/src/supabase_auth.dart
```diff
@@ -151,6 +203,88 @@ class SupabaseAuth {
...
+  Future<void> _handleInitialUri() async {
+    if (!shouldHandleInitialDeeplink) return;
+
+    try {
+      final uri = await getInitialUri();
+      if (uri != null) {
+        _handleDeeplink(uri);
+      }
+    } on PlatformException {
+      // Platform messages may fail but we ignore the exception
```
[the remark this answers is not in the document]
dshukertjr: Great question. I guess I copied that part from `supabase_deep_linking_mixin.dart` line 51

```dart
 Future<void> _handleInitialUri() async {
    if (!SupabaseAuth.instance.shouldHandleInitialDeeplink()) return;

    try {
      final uri = await getInitialUri();
      if (mounted && uri != null) {
        handleDeeplink(uri);
      }
    } on PlatformException {
      // Platform messages may fail but we ignore the exception
    } on FormatException catch (err) {
      if (!mounted) return;
      onErrorReceivingDeeplink(err.message);
    }
  }
```

lib/src/supabase_auth.dart
```diff
@@ -151,6 +203,88 @@ class SupabaseAuth {
...
+    } on FormatException catch (err) {
+      _onErrorReceivingDeeplink(err.message);
+    }
```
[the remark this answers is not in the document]
dshukertjr: Great call! Will add that. I will add handling PlatformException as well, since it seems relevant.

lib/src/supabase_auth.dart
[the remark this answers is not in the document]
bdlukaa: lmao! maybe it's because some platforms aren't supported

lib/src/supabase_auth.dart
```diff
@@ -62,40 +79,103 @@ class SupabaseAuth {
...
+      if (!_instance._initialSessionCompleter.isCompleted) {
+        // Complete with null if the user did not have persisted session
+        _instance._initialSessionCompleter.complete(null);
+      }
+      return _instance;
+    } catch (e) {
+      if (!_instance._initialSessionCompleter.isCompleted) {
+        _instance._initialSessionCompleter.completeError(e);
+      }
+      rethrow;
```
bdlukaa, line 129: It should complete with the stacktrace

```suggestion
    } catch (error, stacktrace) {
      if (!_instance._initialSessionCompleter.isCompleted) {
        _instance._initialSessionCompleter.completeError(error, stracktrace);
      }
      rethrow;
```

bdlukaa (approved):
LGTM overall!

rlee1990:
@dshukertjr how would this work with say flutter riverpods? I was using firebase before and had a Stream but I see this uses GotrueSubscription.

dshukertjr:
@rlee1990 
You should be able to use it very similar to how you are currently using Firebase!
By using [`auth.onAuthStateChange()`](https://supabase.com/docs/reference/dart/auth-onauthstatechange), you can listen to auth state changes just like listening to auth state change in Firebase. There is two values that you can listen to, the event and state, so there it is not using `Stream`, but the basic functionality is the same! Do you have any specific concerns?

rlee1990:
@dshukertjr Thanks for the quick reply. I currently use river pods flutter with a streamProvider. I will just need to see how I am going to use this in a different way with autoRouter and riverPods. Also was the deep link parts removed?

dshukertjr:
@rlee1990 
If you are using `StreamProvider`, you could probably yield a new stream every time the auth state changes on `auth.onAuthStateChange()` listener!

Deep link related logic are not removed, but have been moved inside `supabase_auth.dart` file!

PR #125, opened 2022-06-09
fix: re-initialize client

### LINKED ISSUE #65: Cannot re-initialize Supabase
## Steps to reproduce:

1. Call `Supabase.initialize(url, anonKey)`.
2. Later the app needs to change to a different URL (i.e., switching environments while the app is running), so it calls:
```dart
Supabase.instance.dispose();
Supabase.initialize(newUrl, newAnonKey);
```
## Actual Result:
On the second `initialize()` call, an error is thrown:
`Error: LateInitializationError: Field 'client' has already been initialized.`

## Expected Result:
I'm able to reinitialize the static Supabase singleton instance as often as needed since it's the only one I've got.

## Workaround:
I could create and manage my own `SupabaseClient` instance and just ignore `Supabase` from the `supabase_flutter` package. Note that this loses some functionality, like saving the auth state in Hive/local storage.

Vinzent03:
This allows the client to be re-initialized by removing the final.

close #65

### DIFF
--- a/lib/src/supabase.dart
+++ b/lib/src/supabase.dart
@@ -1,6 +1,5 @@
 import 'package:flutter/foundation.dart';
 import 'package:supabase/supabase.dart';
-
 import 'package:supabase_flutter/src/local_storage.dart';
 import 'package:supabase_flutter/src/supabase_auth.dart';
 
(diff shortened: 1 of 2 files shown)


bdlukaa (approved):

bdlukaa:
@dshukertjr

dshukertjr (approved):
Sorry for the late action. This looks amazing!

PR #131, opened 2022-06-16
Bugfix: OAuth should open in an external browser

Auxx:
OAuth is now blocked inside embedded browsers by Facebook, Google and others. Authentication page should always be opened in an external browser. This fix addresses that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [0.3.1+3]
+- fix: OAuth authentication page should open in an external browser
+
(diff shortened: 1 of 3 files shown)


dshukertjr (commented):
Thank you so much for this @Auxx ! Do you want to bump the package version to `0.3.1+3` and add changelog as well so that we can release it right away?

Auxx:
@dshukertjr done!

dshukertjr (approved):
LGTM! Thanks for the amazing work!

PR #165, opened 2022-08-06
chore: added a very basic example in example directory

dshukertjr:
Added very basic user profile management example app in example directory. 

Related https://github.com/supabase/supabase/issues/8123


example/lib/main.dart
```diff
@@ -0,0 +1,259 @@
+import 'package:flutter/material.dart';
+import 'package:supabase_flutter/supabase_flutter.dart';
+
+Future<void> main() async {
+  await Supabase.initialize(
+      url: 'https://ualrvfiametspbkkntpo.supabase.co',
+      anonKey:
+          'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6InVhbHJ2ZmlhbWV0c3Bia2tudHBvIiwicm9sZSI6ImFub24iLCJpYXQiOjE2NTk1ODU0MDIsImV4cCI6MTk3NTE2MTQwMn0.sMZmPO3ZKRK2-4J6SBc2k9ymoGYQPY7DxyYXxychX8I');
```
bdlukaa, line 8: won't this be leaked?

example/lib/main.dart
```diff
@@ -0,0 +1,259 @@
...
+  void initState() {
+    _getAuth();
+    super.initState();
+  }
+
+  Future<void> _getAuth() async {
+    setState(() {
+      _user = Supabase.instance.client.auth.currentUser;
+    });
+    Supabase.instance.client.auth.onAuthStateChange((event, session) {
```
bdlukaa, line 43: We should store the stream subscription and cancel it on dispose to avoid leaks.

Also, shouldn't we use SupabaseAuth? https://github.com/supabase-community/supabase-flutter/issues/170

example/lib/main.dart
```diff
@@ -0,0 +1,259 @@
...
+
+  @override
+  void dispose() {
+    _emailController.dispose();
+    _passwordController.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
```
bdlukaa, line 81: Maybe we could use supabase-ui?

bdlukaa (approved):

example/lib/main.dart
dshukertjr: Thanks, I always forget this 😂

About using SupabaseAuth, I left a comment here that I would appreciate it a lot if you could take a look!
https://github.com/supabase-community/supabase-flutter/issues/170#issuecomment-1208985193

example/lib/main.dart
dshukertjr: I thought about that as well. I might do that once Supabase Auth UI releases a stable version, but for now, there maybe some breaking changes, so I think without it is okay!

Issue #179, opened 2022-08-10
Passing custom headers to rpc


CodingSoot:
I want to pass custom headers to the rpc call, along with the auth headers.
I've digged in the source code and the implementation seems trivial.

bdlukaa:
You mean params?

CodingSoot:
No. The rpc call gets converted to an HTTP Post request, with the params in its body. I'm talking about the headers of that request.

dshukertjr:
@CodingSoot That is an interesting use case. Could I ask for more details on what you are trying to achieve ultimately here?

dshukertjr:
@CodingSoot Thanks for sharing your use case.

I am not sure if we want to add a option to set a custom headers on rpc call at the moment. 

You can instead have custom backend, which will take care of calling app check endpoints and if the verification success, you can call `rpc()` from the backend with your Supabase service role key. 

By default, `rpc()` functions are exposed, so you may want to [edit the `execute` permissions on it](https://stackoverflow.com/questions/52709231/how-postgresql-give-permission-what-execute-a-function-in-schema-to-user) so that anon key users cannot call it. 

Let me know what you think about this workaround!

Vinzent03:
@dshukertjr May I ask why not? Because the js SDK doesn't support it?

dshukertjr:
@Vinzent03 Yeah, that is the main reason. What do you think about it? Should we consider adding it?

For this particular use case though, since there is already cloud function being used to verify the token, I thought it might be better to let the cloud functions handle the verification process. 

BTW @CodingSoot do you not want to protect your other Postgrest routes like `select()` and `insert()`? Or is the plan to handle all of them in `rpc()`?

CodingSoot:
> You can instead have custom backend, which will take care of calling app check endpoints and if the verification success, you can call `rpc()` from the backend with your Supabase service role key.

If I want to call `rpc()` as if the client made the request, I can use [`setAuth()`](https://supabase.com/docs/reference/javascript/auth-setauth) in the backend before making the call, right ?


> BTW @CodingSoot do you not want to protect your other Postgrest routes like `select()` and `insert()`? Or is the plan to handle all of them in `rpc()`?

For now, I'm planning to handle of them in `rpc()`, but I guess it would be useful if other Postgrest routes were protected too.

dshukertjr:
@CodingSoot 
[...]

PR #181, opened 2022-08-11
feat: Accept custom headers and add add X-Client-Info header

dshukertjr:
We decided that we actually want to have separate `X-Client-Info` header for supabase_flutter, so adding that one as well.


Vinzent03:
I think for #179 it's better suited to pass custom headers at each rpc call, doesn't it?

Vinzent03:
So adding custom headers to the constructor is a nice side effect of this change, right?

dshukertjr:
Yeah, why not? JS SDK and all other SDKs allows for custom header to be set, so I think we can leave it!

Vinzent03 (approved):
Sure, just wanted to clarify.

PR #188, opened 2022-08-15
fix: add missing changelog entries for v1.0.0-dev.4

dshukertjr:


dshukertjr:
Merging this as it's only a changelog update!

PR #280, opened 2022-11-20
feat: add storageretryAttempts to `initialize()`

dshukertjr:

- Adds `storageRetryAttempts`. 
- Adds some docs.

### DIFF
--- a/lib/src/supabase.dart
+++ b/lib/src/supabase.dart
@@ -66,7 +66,7 @@ class Supabase {
     LocalStorage? localStorage,
     Client? httpClient,
     int storageRetryAttempts = 0,
-    bool debug = kDebugMode,
+    bool? debug,
   }) async {
     assert(
       !_instance._initialized,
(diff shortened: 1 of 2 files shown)


lib/src/supabase.dart
```diff
@@ -39,15 +39,34 @@ class Supabase {
...
+    String? schema,
+    Map<String, String>? headers,
     String? authCallbackUrlHostname,
-    bool? debug,
     LocalStorage? localStorage,
     Client? httpClient,
-    Map<String, String>? headers,
-    String? schema,
+    int storageRetryAttempts = 0,
+    bool debug = kDebugMode,
```
Vinzent03, line 69: This could be a breaking change

bdlukaa: Why? `headers` and `schema` were moved to the top of the parameters list

bdlukaa (approved):

lib/src/supabase.dart
Vinzent03: I meant the `debug` parameter. Previously you could assign `null`, but now it has to be non null bool.

dshukertjr: @Vinzent03 Reverted!

Vinzent03 (approved):

Issue #274, opened 2022-11-12
No access_token detected (orj. example)


yutech23:
I USE autH.signInWithPassword() 

I upgraded 1.1.0 version. I got a error. it's "No access_token detected. 
I have tried  example in git but I got same error. 
Please Help can

![image](https://user-images.githubusercontent.com/42283661/201445929-dcce2e81-3e2f-4f75-8683-0895cb738e79.png)

dshukertjr:
@yutech23 Could you follow the issue format and provide steps to reproduce?

d-e-h-i-o:
@dshukertjr I'm getting the same error as well. To reproduce just use the [official example](https://github.com/supabase/supabase/tree/master/examples/user-management/flutter-user-management), bump supabase_flutter to ^1.0.1 and open the app in a webbrowser.

It would be great if you could investigate that bug.

d-e-h-i-o:
Some more info: In my own codebase, I also get the error in version 1.0.1. It basically follows the tutorial from the [supabase website](https://supabase.com/docs/guides/with-flutter#set-up-a-login-page). When I go to my splash page `http://localhost:63771/#SplashPage` it works flawless, however, everytime I reload from a different screen, e.g. `http://localhost:63771/#LearningTab` the error occurs.

``***** SupabaseDeepLinkingMixin startAuthObserver     
***** SupabaseAuthState handleDeeplink http://localhost:63771/#LearningTab   
onReceivedAuthDeeplink uri: http://localhost:63771/#LearningTab   
No access_token detected.   
``
I hope that helps.

d-e-h-i-o:
@yutech23 Configuring [URL strategy](https://docs.flutter.dev/development/ui/navigation/url-strategies) to `usePathUrlStrategy()` helped me, if you are looking for a quick fix.

dshukertjr:
@d-e-h-i-o 
I just tried the example app you mentioned, and was able to sign in without any issue. 
https://www.loom.com/share/913ceb26e8294ab3a1a80330b518a6b6

>however, every time I reload from a different screen, e.g. http://localhost:63771/#LearningTab the error occurs.

What is the use case here? This example app is configured so that the user has to go through SplashPage in order to be redirected to the appropriate page.

yutech23:
> @yutech23 Configuring [URL strategy](https://docs.flutter.dev/development/ui/navigation/url-strategies) to `usePathUrlStrategy()` helped me, if you are looking for a quick fix.

[...]

PR #284, opened 2022-11-22
fix: only parse deep link if it has access token

dshukertjr:

With the current setup, Flutter web application will try to parse every URL as a auth call back URL, which results in `No access_token detected` error, causing confusion for developers getting started with Supabase and Flutter web. With this PR, the link will only be parsed, if it contains `access_token` in the query parameter. I remember @DanMossa suggesting implementing something like this, and I should have listened back then 😂

Fixes https://github.com/supabase/supabase-flutter/issues/274


DanMossa (approved):
Looks good to me!

PR #292, opened 2022-11-28
fix: auth callback not properly parsing callback URL

dshukertjr:

There was a bug with logic inside `_isAuthCallbackDeeplink`, where the link from Supabase does not contain query parameters, but rather passes parameters after `#`, and yet the library looks for `access_token` parameter resulting in no auth callback URL being parsed.

### DIFF
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -178,7 +178,7 @@ class SupabaseAuth with WidgetsBindingObserver {
   /// if _authCallbackUrlHost not init, we treat all deeplink as auth callback
   bool _isAuthCallbackDeeplink(Uri uri) {
     if (_authCallbackUrlHostname == null) {
-      return uri.queryParameters.containsKey('access_token');
+      return uri.toString().contains('access_token');
     } else {
       return _authCallbackUrlHostname == uri.host;
     }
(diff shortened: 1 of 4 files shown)


CodingArcher (approved):

CodingArcher (approved):

lib/src/supabase_auth.dart
```diff
@@ -178,7 +178,7 @@ class SupabaseAuth with WidgetsBindingObserver {
   /// if _authCallbackUrlHost not init, we treat all deeplink as auth callback
   bool _isAuthCallbackDeeplink(Uri uri) {
     if (_authCallbackUrlHostname == null) {
-      return uri.queryParameters.containsKey('access_token');
+      return uri.toString().contains('access_token');
```
Vinzent03, line 181: Instead of searching the whole URI, I would just check the `uri.fragment`. That should be the part after #

Vinzent03 (changes requested):

lib/src/supabase_auth.dart
dshukertjr: Sounds good.
```suggestion
      return uri.fragment.contains('access_token');
```

CodingArcher (approved):

Vinzent03 (approved):

PR #323, opened 2023-01-08
Feat/in app browser oauth

gabfeudo:

Added parameter `authScreenLaunchMode` to `signInWithOAuth` method because Apple explicitly asks for `LaunchMode.inAppWebView`, elsewhere they will not approve the app on App Store.

Actually, `LaunchMode.externalApplication` is the only way to launch auth screen browser.

## What is the new behavior?

Now developers can choose the launch mode for OAuth signin

### DIFF
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -283,6 +283,7 @@ extension GoTrueClientSignInProvider on GoTrueClient {
     Provider provider, {
     String? redirectTo,
     String? scopes,
+    LaunchMode mode = LaunchMode.externalApplication,
     Map<String, String>? queryParams,
   }) async {
     final res = await getOAuthSignInUrl(
@@ -294,7 +295,7 @@ extension GoTrueClientSignInProvider on GoTrueClient {
     final url = Uri.parse(res.url!);
     final result = await launchUrl(
       url,
-      mode: LaunchMode.externalApplication,
+      mode: mode,
       webOnlyWindowName: '_self',
     );
     return result;
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -283,7 +283,7 @@ extension GoTrueClientSignInProvider on GoTrueClient {
     Provider provider, {
     String? redirectTo,
     String? scopes,
-    LaunchMode mode = LaunchMode.externalApplication,
+    LaunchMode authScreenLaunchMode = LaunchMode.externalApplication,
     Map<String, String>? queryParams,
   }) async {
     final res = await getOAuthSignInUrl(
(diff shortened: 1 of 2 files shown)


Vinzent03:
Was there any particular reason for external browser? Isn't inAppWebView better on Android as well?

gabfeudo:
@Vinzent03 I don't know why originally it was externalApplication, but yes, inAppWebBrowser is better also for Android.

Maybe we can set it as default value but leave it as an optional parameter for customization, if someone would ever needs it

Vinzent03:
I haven't worked with this feature so far. I think @dshukertjr you have written that code. Despite, I think we need to export `LaunchMode` so that the dev doesn't have to add the dependency themselves, it looks good to me.

dshukertjr:
@gabfeudo This is great! We definitely should improve the user experience around OAuth login.

One thing I noticed is that when a user signs in with OAuth on iOS with `inAppWebBrowser`, the screen hangs at the attached screenshot. Although the user can tap the "done" button and go back to the app, with `externalApplication`, the user is automatically taken back to the app. Still `inAppWebBrowser` might be better user experience after all, but just wanted to flag it.

Anyway, agree that we should give developers the choice here. Should we at least keep the default value as `externalApplication` so that there are no surprises for those that are already using OAuth with Supabase?

supabase tips (8)

gabfeudo:
@dshukertjr done!
Regarding the loop, I need to investigate but I think I could find a solution

dshukertjr (commented):
@gabfeudo 
Nice! Yeah, if you could figure out the cause of the loop, that would be amazing! Either way, we can probably set the default value to `inAppWebBrowser` in the next major version release. 

Could we add the following to `lib/supabase_flutter.dart`?

```dart
export 'package:url_launcher/url_launcher.dart' show LaunchMode;
```

dshukertjr (approved):
Looks amazing! Thanks for the great work 💪

elliottetzkorn:
> @gabfeudo Nice! Yeah, if you could figure out the cause of the loop, that would be amazing! Either way, we can probably set the default value to `inAppWebBrowser` in the next major version release.
> 
> Could we add the following to `lib/supabase_flutter.dart`?
> 
> ```dart
> export 'package:url_launcher/url_launcher.dart' show LaunchMode;
> ```

Yeah I think for now Apple will still deny all Supabase Flutter apps w/ the webview logins with the current flow. Subscribing to this thread since I think my deploy will be blocked until the loop is resolved.

gabfeudo:
@elliottetzkorn I found out the looping cause. This weekend I’ll implement the fix

dshukertjr:
@gabfeudo That is amazing!

elliottetzkorn:
> @elliottetzkorn I found out the looping cause. This weekend I’ll implement the fix

Great thank you so much! That will make things feel so much better.

Issue #332, opened 2023-01-13
Auth throws error when running on Linux


bdlukaa:
**Describe the bug**
```
══╡ EXCEPTION CAUGHT BY SERVICES LIBRARY ╞══════════════════════════════════════════════════════════
The following MissingPluginException was thrown while activating platform stream on channel
com.llfbandit.app_links/events:
MissingPluginException(No implementation found for method listen on channel
com.llfbandit.app_links/events)

When the exception was thrown, this was the stack:
#0      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:294:7)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)
════════════════════════════════════════════════════════════════════════════════════════════════════
```

**To Reproduce**
Run the app on Linux

**Expected behavior**
App run with no errors

**Screenshots**
n/a

**Version (please complete the following information):**
 - [supabase_flutter 1.2.2]

**Additional context**
og issue: https://github.com/supabase/supabase-flutter/issues/46#issuecomment-1379267169

PR #333, opened 2023-01-14
fix: only start deep link observer on supported platforms

dshukertjr:

fixes https://github.com/supabase/supabase-flutter/issues/332

### DIFF
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -6,6 +6,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
 import 'package:url_launcher/url_launcher.dart';
+import 'dart:io' show Platform;
+import 'package:flutter/foundation.dart' show kIsWeb;
 
 // ignore_for_file: invalid_null_aware_operator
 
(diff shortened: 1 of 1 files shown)


lib/src/supabase_auth.dart
```diff
@@ -6,6 +6,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
 import 'package:url_launcher/url_launcher.dart';
+import 'dart:io' show Platform;
```
bdlukaa, line 9: I believe importing this breaks the web support (at least throw an error when running on the web). To fix it, we must do the stub-real thing. We do the same with isolates, which arent supported on the web

https://github.com/flutter/plugins/tree/main/packages/path_provider/path_provider_windows/lib/src

bdlukaa (changes requested):

lib/src/supabase_auth.dart
dshukertjr: @bdlukaa 
That is what I thought first too, but it seems like as long as we are [checking for web first](https://github.com/supabase/supabase-flutter/pull/333/files#diff-f0d0a37c250b72c9bc959f740b43aa12057188574466716183e4a2d4ce42168aR109) it should be fine as other lines containing `Platform` don't get executed.

[Here](https://zapp.run/edit/flutter-z0a06ta0b06?entry=lib/main.dart&file=lib/main.dart) is an example app working on Flutter web!

bdlukaa: cool!

bdlukaa (approved):

Issue #5, opened 2021-07-22
Native Sign In


bdlukaa:
# Feature request

## Is your feature request related to a problem? Please describe.

I want to sign in using a sign in provider, but the current implementation is not good. Launching the browser is not a good user experience.

## Describe the solution you'd like

Use the [`google_sign_in`](https://pub.dev/packages/google_sign_in) package to sign the user in using Google and [`sign_in_with_apple`](https://pub.dev/packages/sign_in_with_apple) to sign the user in using Apple.

Both of the packages return the user credentials, so, if I could sign in using the user credentials (just like Firebase does), it'd be great!

## Aditional context

If launching the brower will still be the current solution, we could have [in-app-handling](https://pub.dev/packages/url_launcher#browser-vs-in-app-handling).
By default, Android opens up a browser when handling URLs. You can pass `forceWebView: true` parameter to tell the plugin to open a WebView instead.

himeshp:
+1, I have been waiting for this implementation.

dshukertjr:
There needs to be some work done on the server side, but once that is done, we should be able to add this feature!
https://github.com/supabase/gotrue/issues/140

bdlukaa:
Depend on https://github.com/supabase/gotrue-js/issues/169

DanMossa:
Made a PR to bring this to Dart
https://github.com/supabase-community/gotrue-dart/pull/61

nextdev1111:
When will it be implemented

Aurion72:
> Hello, I've implemented Sign In with Apple in my App, but on iOS instead of opening the sign in sheet in the app, it opens safari with Sign In with Apple the then the user is redirect in the app, I'm calling the function like this:
> 
> supabase.auth.signInWithProvider( Provider.apple, options: AuthOptions( redirectTo: 'io.supabase.flutterquickstart://login-callback/'), );
> 
> Is it possible to open the sheet directly in the app and ket the user sign in, I've tried getting the credential and then passing to su-abase, but I always got an error of missing 'expires_in' parameters. Anyone has a solution?
> 
> I'm using su-abase 0.3.6 version on flutter

Hello, if you have any solutions for this i take it :)

hmarat:
@dshukertjr hey :) Do you have even a little progress? We all waiting for it )

dshukertjr:
[...]

PR #355, opened 2023-02-10
Added custom WebView for OAuth (LaunchMode.inAppWebView)

gabfeudo:

Implemented `webview_flutter` library (official Flutter library) to authenticate users when using `LaunchMode.inAppWebView` as `authScreenLaunchMode` for the `signInWithOAuth` method.

When using `signInWithOAuth` method (only with `LaunchMode.inAppWebView`), after the user completed the authentication flow there was a loop that was leading the users to not being automatically redirected back to their app.

## What is the new behavior?

https://user-images.githubusercontent.com/40292753/218302933-4310cabb-823b-4410-a0e8-50f02417a004.mp4

### DIFF
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -354,22 +354,28 @@ class _OAuthSignInWebViewState extends State<_OAuthSignInWebView> {
         NavigationDelegate(
           onPageStarted: (_) => setState(() => isLoading = true),
           onPageFinished: (_) => setState(() => isLoading = false),
-          onWebResourceError: (_) => Navigator.of(context).canPop()
-              ? Navigator.of(context).pop(false)
-              : null,
-          onNavigationRequest: (NavigationRequest request) {
-            if (widget.redirectTo != null &&
-                request.url.startsWith(widget.redirectTo!)) {
-              launchUrlString(request.url);
-              if (Navigator.canPop(context)) {
-                Navigator.of(context).pop(true);
-              }
-            }
-            return NavigationDecision.navigate;
-          },
+          onNavigationRequest: _onNavigationRequest,
+          onWebResourceError: _onWebResourceError,
         ),
       )
       ..loadRequest(widget.uri);
+  }
+
+  void _onWebResourceError(WebResourceError error) {
+    if (Navigator.canPop(context)) {
+      Navigator.of(context).pop(false);
+    }
+  }
+
+  FutureOr<NavigationDecision> _onNavigationRequest(NavigationRequest request) {
+    if (widget.redirectTo != null &&
+        request.url.startsWith(widget.redirectTo!)) {
+      launchUrlString(request.url);
+      if (Navigator.canPop(context)) {
+        Navigator.of(context).pop(true);
+      }
+    }
+    return NavigationDecision.navigate;
   }
 
   @override
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -309,14 +309,10 @@ extension GoTrueClientSignInProvider on GoTrueClient {
     if (authScreenLaunchMode == LaunchMode.inAppWebView &&
         context != null &&
         !kIsWeb) {
-      result = await showModalBottomSheet<bool>(
-        context: context,
-        isScrollControlled: true,
-        builder: (context) => _OAuthSignInWebView(
-          oAuthUri: uri,
-          redirectTo: redirectTo,
-        ),
-      );
+      Navigator.of(context).push(PageRouteBuilder(
+          pageBuilder: (context, insertAnimation, exitamiation) {
+        return _OAuthSignInWebView(oAuthUri: uri, redirectTo: redirectTo);
+      }));
     } else {
       result = await launchUrl(
         uri,
@@ -367,37 +363,28 @@ class _OAuthSignInWebViewState extends State<_OAuthSignInWebView> {
 
   @override
   Widget build(BuildContext context) {
-    return SizedBox(
-      height: MediaQuery.of(context).size.height * 0.9,
-      child: Stack(
-        alignment: Alignment.topCenter,
-        children: [
-          // WebView
-          WebView(
-            userAgent: 'random',
-            initialUrl: widget.oAuthUri.toString(),
-            javascriptMode: JavascriptMode.unrestricted,
-            onPageStarted: (_) => setState(() => isLoading = true),
-            onPageFinished: (_) => setState(() => isLoading = false),
-            navigationDelegate: _handleNavigationRequest,
-            onWebResourceError: _handleWebResourceError,
-          ),
-          // Handle
-          Container(
-            height: 4,
-            width: 40,
-            margin: EdgeInsets.symmetric(vertical: 4),
-            decoration: BoxDecoration(
-              color: Colors.grey,
-              borderRadius: BorderRadius.circular(8),
+    return Material(
+      child: SafeArea(
+        child: Stack(
+          alignment: Alignment.topCenter,
+          children: [
+            // WebView
+            WebView(
+              userAgent: 'Supabase OAuth',
+              initialUrl: widget.oAuthUri.toString(),
+              javascriptMode: JavascriptMode.unrestricted,
+              onPageStarted: (_) => setState(() => isLoading = true),
+              onPageFinished: (_) => setState(() => isLoading = false),
+              navigationDelegate: _handleNavigationRequest,
+              onWebResourceError: _handleWebResourceError,
             ),
-          ),
-          // Loader
-          if (isLoading)
-            const Center(
-              child: CircularProgressIndicator.adaptive(),
-            )
-        ],
+            // Loader
+            if (isLoading)
+              const Center(
+                child: CircularProgressIndicator.adaptive(),
+              )
+          ],
+        ),
       ),
     );
   }
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -91,15 +91,17 @@ class SupabaseAuth with WidgetsBindingObserver {
             if (!_instance._initialSessionCompleter.isCompleted) {
               _instance._initialSessionCompleter.complete(response.session);
             }
-          } on AuthException catch (error) {
-            Supabase.instance.log(error.message);
+          } on AuthException catch (error, stackTrace) {
+            Supabase.instance.log(error.message, stackTrace);
             if (!_instance._initialSessionCompleter.isCompleted) {
-              _instance._initialSessionCompleter.completeError(error);
+              _instance._initialSessionCompleter
+                  .completeError(error, stackTrace);
             }
-          } catch (error) {
-            Supabase.instance.log(error.toString());
+          } catch (error, stackTrace) {
+            Supabase.instance.log(error.toString(), stackTrace);
             if (!_instance._initialSessionCompleter.isCompleted) {
-              _instance._initialSessionCompleter.completeError(error);
+              _instance._initialSessionCompleter
+                  .completeError(error, stackTrace);
             }
           }
         }
@@ -218,8 +220,8 @@ class SupabaseAuth with WidgetsBindingObserver {
             _handleDeeplink(uri);
           }
         },
-        onError: (Object err) {
-          _onErrorReceivingDeeplink(err.toString());
+        onError: (Object err, StackTrace stackTrace) {
+          _onErrorReceivingDeeplink(err.toString(), stackTrace);
         },
       );
     }
@@ -241,13 +243,13 @@ class SupabaseAuth with WidgetsBindingObserver {
       if (uri != null) {
         await _handleDeeplink(uri);
       }
-    } on PlatformException catch (err) {
-      _onErrorReceivingDeeplink(err.message ?? err.toString());
+    } on PlatformException catch (err, stackTrace) {
+      _onErrorReceivingDeeplink(err.message ?? err.toString(), stackTrace);
       // Platform messages may fail but we ignore the exception
-    } on FormatException catch (err) {
-      _onErrorReceivingDeeplink(err.message);
-    } catch (err) {
-      _onErrorReceivingDeeplink(err.toString());
+    } on FormatException catch (err, stackTrace) {
+      _onErrorReceivingDeeplink(err.message, stackTrace);
+    } catch (err, stackTrace) {
+      _onErrorReceivingDeeplink(err.toString(), stackTrace);
     }
   }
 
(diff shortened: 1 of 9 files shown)


gabfeudo:
@dshukertjr one test is failing because the new `webview_flutter` dependency requires  Flutter SDK >= 2.17.0, while the SDK used to run the tests is 2.15.0. Is it possible to switch to 2.17.0?

dshukertjr:
@gabfeudo could we actually downgrade webview_flutter so that it supports v2.15.0?

gabfeudo:
@dshukertjr done! All tests passed. Let's go 🚀

lib/src/supabase_auth.dart
```diff
@@ -288,6 +287,7 @@ extension GoTrueClientSignInProvider on GoTrueClient {
   ///
   ///   * <https://supabase.io/docs/guides/auth#third-party-logins>
   Future<bool> signInWithOAuth(
+    BuildContext context,
```
dshukertjr, line 290: Adding a required argument would be a breaking change, so could we make this an optional named argument?

lib/src/supabase_auth.dart
```diff
@@ -300,12 +300,101 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
+  final Uri uri;
+  final String? redirectTo;
+
+  @override
+  State<_OAuthSignInWebView> createState() => _OAuthSignInWebViewState();
+}
+
+/// Modal bottom sheet with webview for OAuth sign in
+class _OAuthSignInWebViewState extends State<_OAuthSignInWebView> {
+  late final WebViewController controller;
```
dshukertjr, line 342: It seems like this isn't being used, but is it?

lib/src/supabase_auth.dart
```diff
@@ -300,12 +300,101 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
+      );
+    }
+    return result ?? false;
+  }
+}
+
+class _OAuthSignInWebView extends StatefulWidget {
+  const _OAuthSignInWebView({
+    Key? key,
+    required this.uri,
```
dshukertjr, line 329: Nothing critical, but personally I like a more descriptive variable name, maybe something like this:
```suggestion
    required this.oAuthUri,
```

lib/src/supabase_auth.dart
```diff
@@ -300,12 +300,101 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
     );
-    final url = Uri.parse(res.url!);
-    final result = await launchUrl(
-      url,
-      mode: authScreenLaunchMode,
-      webOnlyWindowName: '_self',
+    final uri = Uri.parse(res.url!);
+
+    bool? result;
+    if (authScreenLaunchMode == LaunchMode.inAppWebView) {
```
dshukertjr, line 306: After making `context` optional, we might need to do this to avoid breaking change for users that are already using `inAppWebView`. 
```suggestion
    if (authScreenLaunchMode == LaunchMode.inAppWebView && context != null) {
```

dshukertjr (commented):
@gabfeudo 
Thank you so much for working on this! This is great! Have you tried this on an Android device as well? Also, if possible, would you be able to attach a screen recording of how it looks?

lib/src/supabase_auth.dart
gabfeudo: Yeah I forgot it when i downgraded the library. I’ll delete it

Vinzent03:
I think the changes in the `examples` dir should be changed in another pr. And is there a reason for changing the quotes type in `pubspec.yaml`?

gabfeudo:
> I think the changes in the `examples` dir should be changed in another pr. And is there a reason for changing the quotes type in `pubspec.yaml`?

Ok for the examples. Regarding the quotes, nothing special, it was my Dart linter in VSCode

lib/src/supabase_auth.dart
gabfeudo: I think we should put an assertion error inside the if statement because it could be hard or non intuitive for developers to understand that passing an optional `BuildContext` can cause a different behavior.

Maybe the breaking change is mandatory?

gabfeudo:
> @gabfeudo Thank you so much for working on this! This is great! Have you tried this on an Android device as well? Also, if possible, would you be able to attach a screen recording of how it looks?

@dshukertjr I've added the video

dshukertjr (commented):
Thanks for adding a video. I haven't been able to check yet, but have you tested this on how it plays out on Android?

lib/src/supabase_auth.dart
dshukertjr: I agree that it would feel unintuitive, but we still cannot do a breaking change without a major version bump, which we don't plan to do just not quite yet. We can update the default behavior of the method to a more intuitive one in the next major launch.

gabfeudo: > I agree that it would feel unintuitive, but we still cannot do a breaking change without a major version bump, which we don't plan to do just not quite yet. We can update the default behavior of the method to a more intuitive one in the next major launch.

Ok, I’ll put the null check inside the if

gregorym:
Looking forward to seeing this PR merged 🚀

Vinzent03:
@gregorym Can you still please remove your changes in the example directory and your changes to pubspec.yaml?

gregorym:
> @gregorym Can you still please remove your changes in the example directory and your changes to pubspec.yaml?

I'm not the author, @gabfeudo is.

Vinzent03:
Oh sorry! Was on Mobile, which doesn't load all comments, so I've missed that.

gabfeudo:
> Oh sorry! Was on Mobile, which doesn't load all comments, so I've missed that.

@Vinzent03 done, even if I think there would be better how it was before because:

1. `pubspec.yaml` was following the Prettier standard for .yaml files
2. Files that were modified in the `example` folder are auto-generated every time someone runs `flutter pub get`

gabfeudo:
@dshukertjr @Vinzent03 any news about this implementation?

dshukertjr:
@gabfeudo Apologies for taking time on this one. I'm wondering if using bottom sheet is the best option here. I'm sure you have considered it as well, but I wonder if there is a way to display the webview full screen. 

Maybe something like this might work? I still want to test this one out thoroughly to make sure the user experience is smooth on all the platforms.
```dart
class _OAuthSignInWebViewState extends State<_OAuthSignInWebView> {
  bool _isLoading = true;

  void _handleWebResourceError(WebResourceError error) {
    if (Navigator.canPop(context)) {
      Navigator.of(context).pop(false);
    }
  }

  FutureOr<NavigationDecision> _handleNavigationRequest(
    NavigationRequest request,
  ) {
    if (widget.redirectTo != null &&
        request.url.startsWith(widget.redirectTo!)) {
      launchUrlString(request.url);
      if (Navigator.canPop(context)) {
        Navigator.of(context).pop(true);
      }
    }
    return NavigationDecision.navigate;
  }

  @override
    return MaterialApp(
      home: Scaffold(
        appBar: AppBar(),
        body: Stack(
          alignment: Alignment.topCenter,
          children: [
            // WebView
            WebView(
              userAgent: 'supabase_flutter',
              initialUrl: widget.oAuthUri.toString(),
              javascriptMode: JavascriptMode.unrestricted,
              onPageStarted: (_) => setState(() => _isLoading = true),
              onPageFinished: (_) => setState(() => _isLoading = false),
              navigationDelegate: _handleNavigationRequest,
              onWebResourceError: _handleWebResourceError,
            ),
            // Loader
            if (_isLoading)
              const Center(
                child: CircularProgressIndicator.adaptive(),
              )
          ],
        ),
      ),
    );
  }
}
```

gabfeudo:
> @gabfeudo Apologies for taking time on this one. I'm wondering if using bottom sheet is the best option here. I'm sure you have considered it as well, but I wonder if there is a way to display the webview full screen.
> 
> Maybe something like this might work? I still want to test this one out thoroughly to make sure the user experience is smooth on all the platforms.
> 
> ```dart
> class _OAuthSignInWebViewState extends State<_OAuthSignInWebView> {
>   bool _isLoading = true;
> 
>   void _handleWebResourceError(WebResourceError error) {
>     if (Navigator.canPop(context)) {
>       Navigator.of(context).pop(false);
>     }
>   }
> 
>   FutureOr<NavigationDecision> _handleNavigationRequest(
>     NavigationRequest request,
>   ) {
>     if (widget.redirectTo != null &&
>         request.url.startsWith(widget.redirectTo!)) {
>       launchUrlString(request.url);
>       if (Navigator.canPop(context)) {
>         Navigator.of(context).pop(true);
>       }
>     }
>     return NavigationDecision.navigate;
>   }
> 
>   @override
>   Widget build(BuildContext context) {
>     return MaterialApp(
>       home: Scaffold(
>         appBar: AppBar(),
>         body: Stack(
>           alignment: Alignment.topCenter,
>           children: [
>             // WebView
>             WebView(
>               userAgent: 'supabase_flutter',
>               initialUrl: widget.oAuthUri.toString(),
>               javascriptMode: JavascriptMode.unrestricted,
>               onPageStarted: (_) => setState(() => _isLoading = true),
>               onPageFinished: (_) => setState(() => _isLoading = false),
>               navigationDelegate: _handleNavigationRequest,
>               onWebResourceError: _handleWebResourceError,
>             ),
>             // Loader
>             if (_isLoading)
>               const Center(
>                 child: CircularProgressIndicator.adaptive(),
>               )
>           ],
>         ),
>       ),
>     );
>   }
> }
> ```

I saw that on all the big apps on the App Store, when you tap on Google Sign In they open a bottom sheet

dshukertjr:
@gabfeudo 
Is this the UI you are referring to?
Screenshot 2023-02-27 at 13 23 05


The bottom sheet UI might work, but I personally think it's safer to default to the full screen UI, just because the bottom sheet might not always work on all devices with all login providers. For example, Google login on Android uses a dialog like this:

Screenshot 2023-02-27 at 13 39 48

We will get these native signin UIs when this lands https://github.com/supabase/supabase-flutter/issues/5, so the UI that we are aiming for here should just be a good enough alternative until native login lands that will work fine for all login providers on all devices.

dshukertjr:
@gabfeudo Would you mind if I update this PR to use full screen navigation instead of a bottom sheet?

dshukertjr:
@shashankdaima That seems like a separate issue. I will follow up in the issue you have opened!

gregorym:
Can we get this merged? :D

gabfeudo:
> just because the bottom sheet might not always work on all devices with all login providers

Why shouldn’t it work?

dshukertjr:
@gabfeudo 
What I meant was that bottom sheet is not always the best UI choice for all login providers and platforms. You can see some examples in [my comment](https://github.com/supabase/supabase-flutter/pull/355#issuecomment-1445703860).

We won't go too wrong if it's a full screen dialog.

elliottetzkorn:
It would be nice to add a prop to swap between the sheet and a full screen dialog. The sheet that @gabfeudo has implemented is a better UI when it it does not break. I would prefer to use that for my oauth for sure.

elliottetzkorn:
That said if we are waiting for fully native log-in anyways I see how just using full screen dialog for now with no prop options for a sheet is more prudent.

dshukertjr:
@elliottetzkorn Thanks for the comment! Another thing about using bottom sheet is that it only works if the app uses `MaterialApp` as its root app, and will throw an `No Scaffold widget found.` exception. 

@gabfeudo Would you want to tackle this problem? One way to solve it is [this](https://github.com/supabase/supabase-flutter/pull/355#issuecomment-1445246944), but since bottom sheet seems pretty popular, I'm happy to keep the bottom sheet if we can find a solution for it!

elliottetzkorn:
Is it possible to know what the rough timeline is for merging this? I am waiting to publish my app to the app store until this is merged, since the current UI is so confusing to users.

lib/src/supabase_auth.dart
```diff
@@ -303,12 +318,92 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
-      mode: authScreenLaunchMode,
-      webOnlyWindowName: '_self',
+    final uri = Uri.parse(res.url!);
+
+    final willOpenWebview = authScreenLaunchMode == LaunchMode.inAppWebView &&
+        context != null &&
+        !kIsWeb && // `Platform.isIOS` throws on web, so adding a guard for web here.
+        Platform.isIOS;
+
+    if (willOpenWebview) {
```
dshukertjr, line 328: Updated the code so that webview will only launch on iOS as Android does not allow performing OAuth on webviwe.

https://developers.googleblog.com/2016/08/modernizing-oauth-interactions-in-native-apps.html

lib/src/supabase_auth.dart
```diff
@@ -303,12 +318,92 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
+      Navigator.of(context).push(PageRouteBuilder(
+          pageBuilder: (context, animation, secondaryAnimation) {
+        return _OAuthSignInWebView(oAuthUri: uri, redirectTo: redirectTo);
+      }));
+      return true;
```
dshukertjr, line 333: Opening the OAuth screen in full screen instead of bottom sheet so that Cupertino app users can take advantage of this nice feature as well!

dshukertjr (approved):
Thanks @gabfeudo for working on this. I did some minor cleanup, but it looks fantastic!

bertuz:
guys, one question: apparently it does not work properly on iOS:

consider this snippet:
```
  Future _googleSignupTapped(BuildContext context) async {
    try {
      final success = await Supabase.instance.client.auth.signInWithOAuth(
          Provider.google,
          redirectTo: dotenv.get('SUPABASE_AUTH_CALLBACK', fallback: ''),
          authScreenLaunchMode: LaunchMode.platformDefault,
          context: context);

      if (!success) {
        Navigator.of(context).pop(false);
      }
    } catch (e) {
      print(e.toString());
      context.showErrorSnackBar(
          message: 'Failed while signin in. Please retry');
      Navigator.of(context).pop(false);
    }
  }
```

- the login window is opened
- if logged in already, a blank page is shown. `done` must be pressed
- once pressed, an exception is thrown `flutter: PlatformException(Error, Error while launching https://<myproject>.supabase.co/auth/v1/authorize?provider=google&redirect_to=net.bertamini.myproject%3A%2F%2Flogin-callback%2F, null, null)
`
- if listening to the events, `Supabase.instance.client.auth.onAuthStateChange.listen((data){...})` a `signedIn` event is fired anyway, but with no `data.session` in it

is this expected? Am I misusing it in some way?

another question: is it using https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller/ for iOS? Just asking because I'm not seeing the login as a popup, which is usually the standard way of logging in with OAuth functionalities on iOS

dshukertjr:
@bertuz 
Could you open an issue? We can discuss it there!

bertuz:
yep, just wanted to make sure it's not an improper use of my brain in weekend mode 😜

lib/src/supabase_auth.dart
kkalisz: Well, I have tested that, and with the current code that this is not working on Android,
we got "Access blocked Error" but indeed when I allow opening WebBrowser on Android, WebBrowser is opened and login works fine.

dshukertjr: @kkalisz 
Could you give more details on this `Access blocked Error`? What provider did you use and if there are more error message, providing them would be helpful.

kkalisz: Hi @dshukertjr,

I'm experiencing exactly the same problem as in https://github.com/supabase/supabase-flutter/issues/357 
So google provider, and 403 disallowed error.
It looks like with the current code, this is expected because for android we are using default inApp browser.

I little bit changed the implementation locally, and it works correctly now. I will try to open PR Today
Thanks for your interest

PR #373, opened 2023-02-25
fix: show web support on pub.dev

Vinzent03:
Currently [pub.dev](https://pub.dev/packages/supabase_flutter/score) states that this package doesn't support web, because of the `dart:io` import. I've changed it to `universal_io`, which should fix the issue. This doesn't add a new dependency, as the sub libraries are already using it.


dshukertjr (approved):
Thanks for the fix!

PR #381, opened 2023-03-02
Add experimental signInWithIdToken for Apple, Google

DanMossa:


README.md
```diff
@@ -91,6 +91,31 @@ class _MyWidgetState extends State<MyWidget> {
 }
 ```
 
+#### Native Auth example
+```dart
```
dshukertjr, line 95: Could we add a warning here that says the current native auth implementation is experimental and there might be API changes in the future?

PR #392, opened 2023-03-08
fix: declare web support

Vinzent03:
The previous fix didn't work, because universal_io is still listed as not supported. I'm now hard coding the supported platforms as described [here](https://dart.dev/tools/pub/pubspec#platforms). Hopefully it works this time.

I added a section about linux support in the readme, because one might think we don't support linux at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,10 @@ Flutter Client library for [Supabase](https://supabase.com/).
 
 - Documentation: https://supabase.com/docs/reference/dart/introduction
 
+## Platform Support
+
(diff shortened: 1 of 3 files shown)


dshukertjr (commented):
Thanks for the fix! Sweet! I genuinely did not know that you can just declare what platform the package supports! Could we bump the package version and I can quickly release it!

Vinzent03:
What about using #381 for this?

dshukertjr:
@Vinzent03 Sounds great!

dshukertjr (approved):

PR #398, opened 2023-03-13
Update documentation to reflect correct code for Apple auth login

elliottetzkorn:

Doc update

The current doc code for Apple native auth is not correct.

## What is the new behavior?

The new behavior changes the provider from Google to Apple, and critically includes the nonce, non-hashed, in the constructor. 

## Additional context

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,9 @@ final AuthorizationCredentialAppleID credential = await SignInWithApple.getApple
 );
 
 return supabase.auth.signInWithIdToken(
-  provider: Provider.google, 
+  provider: Provider.apple, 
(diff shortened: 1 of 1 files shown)


DanMossa:
This is the exact code I'm using in my app and it works. What about it doesn't work for you?

elliottetzkorn:
You are passing an Apple auth credential into a Google sign in, in the documentation, without a nonce.

DanMossa (approved):

elliottetzkorn:
@DanMossa would you be willing to share the code you are using in your app? I would like to update this READM for the google native case as well, which I am struggling with.

Issue #184, opened 2022-08-12
Reset password generates 3 onAuthState changes. 


JulianSwales:
1.0.0-dev-3
I'm wondering if this is correct. When I generate a resetPasswordForEmail, it send me the reset deep link. 
When I click on that link, I get three onAuthStateChange events. 
signedIn followed by passwordRecovery followed by signedIn.
Why am I getting the signedIn states?  It's I'm watching for signed in as well as password recovery is getting tricky to react to a given state, unless this is expected? 
I'm watching client.auth.onAuthStateChange((event, session) {});

yukiyamazaki:
Has this problem been resolved?
I have the same problem.

Vinzent03:
I can't reproduce the issue on `v1.5.2`. I get an `signedIn` event followed by `passwordRecovery`. Depending on your JWT expiry limit and the time you open the link it may be caused by the jwt refreshing, but this should have an `tokenRefreshed` event before that.

Please comment or open a new issue if you still encounter this issue.

PR #401, opened 2023-03-15
Dont call recover session if deeplink is handled

### LINKED ISSUE #365: AuthChangeEvent signedIn called after passwordRecovery event
I am implementing password recovery feature and using `resetPasswordForEmail` method. After method is called, I am getting an email with **Reset Password** link. The deep linking is set up and after clicking on the link in email, my app is opened and I events are triggered in the following order: 

1. AuthChangeEvent signedIn 
2. AuthChangeEvent passwordRecovery
3. AuthChangeEvent signedIn

In my opinion it is expected that **signedIn event is triggered first followed by passwordRecovery event**, but **the signedIn event after passwordRecovery is unexpected**, since it does not allow me to handle the reset password that I am aiming to. 

To subscribe on events I use listen() method on supabaseClient.auth.onAuthStateChange.

I have noticed that after I do **RESTART** of an application I am not getting third sigendIn event, but this is not real life scenario and I can not expect that user opens 2 times app in order to reset password. 

**Version:**
 - [1.4.0]

Could you please advice?

rokk4:

fixes #365 #184 

Recover Session gets called when the deeplink was handled and the session
from the deeplink was saved. Resulting in emitting signIn when it should not have been emitted. 1. signedIn 2. recovery 3. signedIn

## What is the new behavior?

didChangeAppLifeCycle now checks if a deeplink is handled and if so it does not call recoverSupabaseSession when it should not. Resulting in  only 1. signedIn 2. recovery event beeing emitted.

### DIFF
--- a/lib/src/supabase_auth.dart
+++ b/lib/src/supabase_auth.dart
@@ -140,7 +140,9 @@ class SupabaseAuth with WidgetsBindingObserver {
   void didChangeAppLifecycleState(AppLifecycleState state) {
     switch (state) {
       case AppLifecycleState.resumed:
+        if (!_initialDeeplinkIsHandled) {
         _recoverSupabaseSession();
+        }
         break;
       case AppLifecycleState.inactive:
         break;


Vinzent03:
That won't work, because `_initialDeeplinkIsHandled` is always true after startup. You basically disable the recover session completely.

rokk4:
> because `_initialDeeplinkIsHandled` is always true after startup.

This is not correct. 

It is initialized `false` and never reassigned except in '_handleInitalUri'.
https://github.com/supabase/supabase-flutter/blob/98f5c981b102c3475f102ab9aa24d798b74951a7/lib/src/supabase_auth.dart#L44
https://github.com/supabase/supabase-flutter/blob/98f5c981b102c3475f102ab9aa24d798b74951a7/lib/src/supabase_auth.dart#L238-L240
Screenshot 2023-03-15 at 14 11 00



>  That won't work

It is working for me, please test it.

> You basically disable the recover session completely.

Nope, since  `_initialDeeplinkIsHandled` is always `false` after startup.
If it would, no initial deepDeepLink would ever be handled, look at:
https://github.com/supabase/supabase-flutter/blob/98f5c981b102c3475f102ab9aa24d798b74951a7/lib/src/supabase_auth.dart#L238-L240

Vinzent03:
> It is initialized false and never reassigned except in '_handleInitalUri'. 

Correct, but it's called in `_startDeeplinkObserver`, which is called in `initialize`. So it's always called and then set to true, to prevent multiple executions of the initial deeplink after a `dispose` and `initilize` of `SupabaseAuth`.

> Nope, since _initialDeeplinkIsHandled is always false after startup.
If it would, no initial deepDeepLink would ever be handled, look at:

As described, `_initialDeeplinkIsHandled` is there to ensure `_handleInitialUri` only runs once.

> >That won't work
> 
> It is working for me, please test it.

I agree that you don't get a third `signedIn` event, but that's just, because `_recoverSupabaseSession` is no more called after `resume`.

rokk4:
Why is `_recoverSupabaseSession` needed anyway?

I tested this on iOS/Android and Web with not calling `_recoverSupabaseSession` at all. Everything is working as expected.

Even with 'Don't keep activites' everything is fine. Because the persisted session is recovered in `initialize`.

When `_recoverSupabaseSession` is called, `_recoverSession` actually runs twice resulting in one `signedIn` event to much. 

What is the purpose of running `_recoverSession` twice?

Vinzent03:
The recover session method is used to handle eventually expired access token. This is needed, because the timer normally used to handle expired access token may not fire while the app is in background.
`initialize` is only called after a complete start/restart of the app.

rokk4:
Then `_recoverSupabaseSession` and/or `_recoverSession` and  `_callRefreshToken` should not emit signIn or tokenRefreshed events when called from resume.
`_recoverSupabaseSession` should only quietly refresh the token or emit signedOut if an necessary refresh fails so the app can _resume_ were it left of. 

Why is it necessary to emit these events from resume?

Vinzent03:
I don't know what the best way is. @dshukertjr Is there a particular reason to emit a sign in event if in realty nothing has changed, not even a token refresh? I'm open to use another method than `recoverSession` to handle an eventually expired access token, so that we don't have to read from the disk as well, because the state on disk and memory should be the same, right?

dshukertjr:
@Vinzent03 
Sorry for the delay on the response here. I don't think there is any particular reason for the sign in event being emitted. 

I agree that we should have another method for handling eventually expired access token.

> we don't have to read from the disk as well, because the state on disk and memory should be the same, right?

I think so, but there might be some edge cases in Flutter web context where the user has multiple tabs of the same app opened.

Vinzent03:
@dshukertjr The edge case flutter web is actually a good point. I've created #44 now for this.

rokk4:
@Vinzent03  @dshukertjr  could you kindly explain how #44 solves this issue?

PR #418, opened 2023-03-29
fix: catch initial session

Vinzent03:
Coming from supabase/gotrue-dart#126 I'm catching all erros here. You can still use the `initialSession` as before, but if you are not using it, you don't get spammed with the error. Since using the stream in the next release is recomended, right?


dshukertjr (approved):
Nice! I think the `onAuthStateChanged` will throw error here, but with https://github.com/supabase/supabase-flutter/pull/416 we should be good there as well!

Issue #421, opened 2023-03-30
Session not persisted on auth.refreshSession() in Flutter Web Release Mode


TommyLCox:
See PR for this.. #422 

**Describe the bug**
On calling auth.refreshSession(), the new refreshToken is not being persisted in Flutter Web Release Mode Only. It works in Flutter Web Debug mode. The reason it works in debug mode is that in debug mode on auth.refreshSession both the AuthChangeEvent.tokenRefreshed AND AuthChangeEvent.signedIn events fire. In release mode, only the AuthChangeEvent.tokenRefreshed fires on refreshSession(). Supabase Auth only persists the session on the signedIn event, not on the tokenRefreshed event.

**To Reproduce**
Short Version:
1) Authenticate in a flutter web app using email/pwd.
2) refresh session with auth.refreshSession()
3) refresh browser
4) refresh session again with auth.refreshSession()
5) throws invalid refresh token  error.

Long version to test/observe:
To test, you first login, execute something that calls auth.refreshSession () and does a debugPrint of the refreshToken before and after the call to auth.refreshSession(). One can see that first token is different than the second token which comes after auth.refreshSession() which is good. (The first token in my test was supabase.auth.currentSession!.refreshToken. Then called auth.refreshSession() The second token was in the AuthStateChange listener(data) as data.session!.refreshToken. I am running another test to also check supabase.auth.currentSession!.refreshToken in the AuthStateChange listener after the auth.RecoverSession as well.) The key point here is that you login, run the auth.refreshSession, note the value token1 (before refresh) and then note value of  token2 (after refresh). Next, you refresh your web browser. Then, you run auth.refreshSession again and note the value of token1 just prior to this second auth.refreshSession(). You will see that in debug mode, token1 equals the value of token2 that was new new value after the first auth.refreshSession just before the web browser was refreshed. That is the EXPECTED behavior. It meant that the first auth.refreshSession acquired a new refresh token and held that new value AFTER the browser refresh.  HOWEVER, in release mode, this is not the case. After a browser refresh, you run auth.refreshSession where you check for token1 (just before the refresh) and you will see that token1 is the same as the first token1 back before the very first refreshSession(). So, the new token after the first refreshSession was LOST after a browser refresh in Release mode.

**Expected behavior**
The exepected behavior is that if you refresh the session with auth.refreshSession() so that a new token is generated, then refresh the browser, then the new token should have been retained. 

**Version (please complete the following information):**
[...]

PR #422, opened 2023-03-30
Update supabase_auth.dart

TommyLCox:
_onAuthStateChange now persists the session on AuthChangeEvent .signedIn AND .tokenRefreshed

See #421 - It persists the token to localStorage on auth.refreshSession() which generates AuthChangeEvent.tokenRefreshed. 

The current behavior is that only persists to localStorage on AuthChangeEvent.signedIn.

See #421 

## What is the new behavior?
It persists the session to localStorage on BOTH AuthChangeEvent.signedIn AND AuthChangeEvent.tokenRefreshed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## [1.6.2]
 
-- fix: Update supabase_auth.dart [#422](https://github.com/supabase/supabase-flutter/pull/422)
+- fix: persist session to local storage on `onAuthStateChanged` event with a session [#422](https://github.com/supabase/supabase-flutter/pull/422)
 
 ## [1.6.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [1.6.2]
+
+- fix: persist session to local storage on `onAuthStateChanged` event with a session [#422](https://github.com/supabase/supabase-flutter/pull/422)
+
 ## [1.6.1]
 
 - fix: log errors from `onAuthStateChange` [#416](https://github.com/supabase/supabase-flutter/pull/416)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [1.6.2]
+
+- fix: Update supabase_auth.dart [#422](https://github.com/supabase/supabase-flutter/pull/422)
+
 ## [1.6.1]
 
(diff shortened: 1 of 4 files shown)


lib/src/supabase_auth.dart
```diff
@@ -186,7 +186,7 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   void _onAuthStateChange(AuthChangeEvent event, Session? session) {
     Supabase.instance.log('**** onAuthStateChange: $event');
-    if (event == AuthChangeEvent.signedIn && session != null) {
+    if (((event == AuthChangeEvent.signedIn) || (event == AuthChangeEvent.tokenRefreshed) )&& session != null) {
```
Vinzent03, line 189: I think you can leave the inner brackets.

dshukertjr: What about just removing the checks for AuthChangeEvent all together?
```suggestion
    if (session != null) {
```

I want to bring https://github.com/supabase/gotrue-js/pull/629 to gotrue-dart when it lands.

dshukertjr:
I will quickly fix this and publish a new version.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [1.6.2]
+
+- fix: Update supabase_auth.dart [#422](https://github.com/supabase/supabase-flutter/pull/422)
```
Vinzent03, line 3: What about `fix: properly persist auth session`

Vinzent03 (changes requested):

dshukertjr:
@Vinzent03 
I have reverted the failing tests. Let me tackle the failing test in another PR.

Vinzent03:
I'm currently looking into the failing tests. I think they come from gotrue, due to completer not handled. I would still release this as the issue from this pr is more urgent than the test fail. That should only affect where the refresh token is wrong.

dshukertjr:
Fair enough.

Vinzent03 (approved):

Vinzent03:
@dshukertjr Do you know the reason for the many completer in the refresh token login in gotrue? Methods like `_callRefreshToken` both return a future and complete a completer https://github.com/supabase/gotrue-dart/blob/c7f3febdd151da73c2ce2fdb9918fa61a4a22a21/lib/src/gotrue_client.dart#L654

dshukertjr:
@Vinzent03 
Those completers were added to handle token refresh retries.

dshukertjr:
@Vinzent03 but yeah, I see places that are not using the completer properly...

Issue #294, opened 2022-11-29
Allow HiveLocalStorage to accept custom directory path


kaboc:
**Is your feature request related to a problem? Please describe.**

`Hive.initFlutter()` is used in the default local storage in supabase_flutter. It uses the directory obtained with `getApplicationDocumentsDirectory()`, so the data is stored in the document folder on windows. I don't think it is a suitable place for app data. (I haven't checked on other platforms.)

Besides that, in desktop apps, I'd like to store the session data for debug build in another place and for release build in the same directory as the executable. It'll be great if it is possible to specify the directory.

**Describe the solution you'd like**

https://github.com/supabase/supabase-flutter/blob/ce2056f0804d152f65546ddcb843ae6f5c54b7a3/lib/src/local_storage.dart#L64-L71

```dart
class HiveLocalStorage extends LocalStorage {
  const HiveLocalStorage({this.directory})
      : super(
          initialize: _initialize,
          hasAccessToken: _hasAccessToken,
          accessToken: _accessToken,
          removePersistedSession: _removePersistedSession,
          persistSession: _persistSession,
        );

  final String? directory;
```

```dart
await Supabase.initialize(
  url: 'SUPABASE_URL',
  anonKey: 'SUPABASE_ANON_KEY',
  localStorage: const HiveLocalStorage(directory: 'path/to/dir'),
);
```

The property name may be other then `directory`, like `path` or `directoryPath`.

Vinzent03:
@dshukertjr  I think we should generally switch to `Application Support`(https://pub.dev/packages/path_provider#supported-platforms-and-paths), because current `Application Documents` stores the auth hive box on linux in `$HOME/documents/auth/..`, which I think is VERY strange. 
I've already written a migration code in my own application, which I could reuse for this package.

dshukertjr:
@Vinzent03 How would the migration script work? Maybe we can do it for v2.0
https://github.com/supabase/supabase-flutter/issues/287

Vinzent03:
@dshukertjr 
```dart
import 'dart:io';

import 'package:hive/hive.dart';
import 'package:path/path.dart' as path;
import 'package:path_provider/path_provider.dart';

Future<void> migrateAndInitHive() async {
  if (!kIsWeb) {
    final supportDirPath = (await getApplicationSupportDirectory()).path;
    final documentsDir = await getApplicationDocumentsDirectory();

    if (!await Directory(supportDirPath).exists()) {
[...]

Issue #287, opened 2022-11-23
supabase-flutter v2 feature tracker


dshukertjr:
Let's keep all the features that we want to add/ remove/ modify in supabase-flutter v2.

dshukertjr:
## Immutability

Instead of modifying the properties of Postgrest classes, create a copy every time calling filter/ modifier methods. 

For example, queries like the following should work. 
```dart
var fooQuery = myQuery.eq('some_col', 'foo');
var notFooQuery = myQuery.neq('some_col', 'foo');
```

https://github.com/supabase/supabase-flutter/issues/162#issuecomment-1324226510

dshukertjr:
### Minimal Dart SDK Version

It depends on when exactly v2 is released, but  2.18.0 might be a good one maybe?

dshukertjr:
Better type support on realtime. Currently, all properties of `ChannelFilter()` is string. Haven't put too much thoughts into it yet, but I feel like we can do better here.

Vinzent03:
## New Hive box location for authentication

Currently, the box is stored in the default location of [hive_flutter](https://pub.dev/packages/hive_flutter), which is the Application Documents directory. This corresponds on Linux to `$HOME/documents`, which is very strange. The Application Support directory suits much better and is supported on all platforms as well. (They come from the [path_provider](https://pub.dev/packages/path_provider#supported-platforms-and-paths) package)

I've commented an example migration code in #294

dshukertjr:
## Use `LaunchMode.inAppWebView` as the default value for doing OAuth sign-in.

Currently, it;s `LaunchMode.externalApplication`, but `LaunchMode.inAppWebView` should provide better user experience. 
Related https://github.com/supabase/supabase-flutter/pull/323

dshukertjr:
## Remove deprecated `returning` option on `.delete()`

Related https://github.com/supabase/postgrest-dart/pull/105#discussion_r1082396094

dshukertjr:
## Explorer using the Dart native isolate feature to replace yet another json isolate

We might be able to use the new isolate feature introduced in Dart 2.19.0 depending on which minimum SDK version that we decide on.
https://medium.com/dartlang/better-isolate-management-with-isolate-run-547ef3d6459b

TC3000:
Allow realtime joins. Currently we can only listen to realtime events from a single table. It would be really nice to join data from multiple tables, i.e fetch `Post` and `Comments` as a stream.

[...]

PR #432, opened 2023-04-04
feat: add pkce support

dshukertjr:

Add pkce support.

To be merged after https://github.com/supabase/supabase-dart/pull/190


lib/src/local_storage.dart
```diff
@@ -113,3 +116,37 @@ class HiveLocalStorage extends LocalStorage {
         .put(supabasePersistSessionKey, persistSessionString);
   }
 }
+
+/// local storage to store pkce flow code verifier.
+class HiveGotrueAsyncStorage extends GotrueAsyncStorage {
```
Vinzent03, line 121: You have stated [here](https://github.com/supabase/supabase-flutter/issues/287#issuecomment-1336246821) that you are open for using shared preferences as well. Is there a reason to pick hive again?

dshukertjr: Thanks for asking! No particular reason, but mostly only because we are currently using Hive, and using SharedPreferences will add dependencies, but would you prefer to just start using SharedPreferences right now?

We could embed a script similar to [yours for migrating hive path](https://github.com/supabase/supabase-flutter/issues/294#issuecomment-1334944076) to migrate the auth persistence string stored in hive to sharedPreferences and just start using shared preferences right now. 

I don't know what I was thinking during [this](https://github.com/supabase/supabase-flutter/issues/294#issuecomment-1334689312) discussion, but no need to wait until v2 to do this for sure. In fact, if we would to make the migration, we should do it early, and keep the script over multiple versions, right?

lib/src/local_storage.dart
```diff
@@ -113,3 +116,34 @@ class HiveLocalStorage extends LocalStorage {
...
+class SharedPreferencesGotrueAsyncStorage extends GotrueAsyncStorage {
+  SharedPreferencesGotrueAsyncStorage() {
+    _initialize();
+  }
+
+  static const pkceHiveBoxName = 'supabase.pkce';
+
+  late final SharedPreferences _prefs;
+
+  Future<void> _initialize() async {
```
Vinzent03, line 130: What if _initialize() is not yet finished, so `_prefs` is not set and `getItem()` is called? What about using a completer for `_prefs`?

lib/src/supabase.dart
```diff
@@ -70,27 +76,30 @@ class Supabase {
...
-    );
+    _instance._init(url, anonKey,
+        httpClient: httpClient,
+        customHeaders: headers,
+        schema: schema,
+        storageRetryAttempts: storageRetryAttempts,
+        realtimeClientOptions: realtimeClientOptions,
+        gotrueAsyncStorage:
+            pkceAsyncStorage ?? SharedPreferencesGotrueAsyncStorage(),
+        authFlowType: authFlowType);
```
Vinzent03, line 95:
```suggestion
        authFlowType: authFlowType,);
```

lib/src/supabase_auth.dart
```diff
@@ -145,6 +149,7 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   /// Dispose the instance to free up resources
   void dispose() {
+    _initialDeeplinkIsHandled = false;
```
Vinzent03, line 152: I guess this was added for the tests, but I don't think that's a good idea for normal apps. What about calling this only in tests?

Vinzent03 (changes requested):

lib/src/local_storage.dart
dshukertjr: Love it!

lib/src/supabase_auth.dart
dshukertjr: Makes sense. Let me do that.

lib/src/local_storage.dart
```diff
@@ -113,3 +117,40 @@ class HiveLocalStorage extends LocalStorage {
...
   }
 }
+
+/// local storage to store pkce flow code verifier.
+class SharedPreferencesGotrueAsyncStorage extends GotrueAsyncStorage {
+  SharedPreferencesGotrueAsyncStorage() {
+    _initialize();
+  }
+
+  static const pkceHiveBoxName = 'supabase.pkce';
```
Vinzent03, line 127:
```suggestion
```
I think this can be removed.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## [1.9.0]
+
+- feat: add pkce support #432 [#455](https://github.com/supabase/supabase-flutter/pull/432)
```
Vinzent03, line 3:
```suggestion
- feat: add pkce support [#432](https://github.com/supabase/supabase-flutter/pull/432)
```

test/widget_test_stubs.dart
```diff
@@ -68,16 +72,129 @@ class MockLocalStorage extends LocalStorage {
...
-      .setMockMethodCallHandler(channel, (call) => null);
+      .setMockMethodCallHandler(channel, (call) async => initialLink);
 
   TestDefaultBinaryMessengerBinding.instance?.defaultBinaryMessenger
-      .setMockMethodCallHandler(anotherChannel, (message) => null);
+      .setMockMethodCallHandler(anotherChannel, (message) async => null);
+}
+
+class MockAsyncStorage extends GotrueAsyncStorage {
+  static const pkceHiveBoxName = 'supabase.pkce';
```
Vinzent03, line 101:
```suggestion
```

Vinzent03 (changes requested):

lib/src/local_storage.dart
dshukertjr: Oh yeah. Great catch!

Vinzent03 (approved):

PR #437, opened 2023-04-12
feat: Add `signInWithApple` method

dshukertjr:

After the very long wait, Supabase is bringing proper native mobile auth! There is a `bundleId` field within the Supabase dashboard that you can add your iOS bundle ID to start using Apple login for both iOS and other platforms. This only allows a single iOS app to be used with a single Supabase instance, but we will take care of that in the later times. 

In the mean while, implementing Apple login could be kind of complicated. We have had the `signInWithIdToken` method thanks to @DanMossa for a while, but the process of signing a user in requires generating nonce and verifying it, and just a whole bunch of stuff that the developers don't really need to worry about. 

I propose to add a `signInWithApple()` method that allows supabase_flutter user to implement Apple login with just one line.

```dart
final AuthResponse response = await supabase.auth.signInWithApple();
```

Pros and cons of adding this method instead of allowing developers to import Apple login package themselves and add the login from scratch.

Pros:
- Simple developer experience. Literally one line to start using Apple Login.

Cons:
- Adds unnecessary extra dependencies for those that don't use Apple login

Would love to hear comments/ feedback about this approach! 

## What about other providers?

If we go this path, the plan is to add `signInWithGoogle` method next, and create each individual sign-in methods for different native auth providers. I don't think we will ever get to supporting every single OAuth providers that we support currently, and quite frankly, I think supporting Apple and Google login will be good enough for 90% of the use cases, but we will see and hear what the community wants after Google login.

To implement Google login, we will use the [AppAuth ](https://pub.dev/packages/flutter_appauth) package, because the official Google login package does not allow us to specify the nonce, and after talking with the auth team, we have decided that for security reasons, we have to make nonce a required parameter. With any other providers that we may support in the future, we will see if we can implement it using the AppAuth package as well to minimize the extra dependencies of supabase_flutter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,32 +96,21 @@ class _MyWidgetState extends State<MyWidget> {
 ```
 
 #### Native Sign in with Apple example
-##### The `signInWithIdToken` method is currently experimental and is subject to change. Apple sign in on iOS is working, as shown in the code below, but other combinations of Apple/Google sign in on iOS/Android may not be. Follow [this issue](https://github.com/supabase/supabase-flutter/issues/399) for platform support progress.
+
+Before you run the code, you need to [register your app ID with Apple](https://developer.apple.com/help/account/manage-identifiers/register-an-app-id/) with the `Sign In with Apple` capability selected. 
 
 ```dart
-import 'dart:convert';
-import 'package:crypto/crypto.dart';
-import 'package:sign_in_with_apple/sign_in_with_apple.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
 
 final supabase = Supabase.instance.client;
-final String nonce = uuidV4();
-final String hashedNonce = sha256.convert(utf8.encode(nonce)).toString();
-const String clientId = 'com.app';
 
-final AuthorizationCredentialAppleID credential = await SignInWithApple.getAppleIDCredential(
-  scopes: [
-    AppleIDAuthorizationScopes.email,
-  ],
-  nonce: hashedNonce,
-);
-
-return supabase.auth.signInWithIdToken(
-  provider: Provider.apple, 
-  idToken: credential.identityToken,
-  nonce: nonce
-);
+return supabase.auth.signInWithApple();
 ```
+
+`signInWithApple()` is only supported on iOS. Other platforms can use the `signInWithOAuth()` method to perform Apple login.
+
+The `signInWithApple` method is currently experimental and is subject to change. Follow [this issue](https://github.com/supabase/supabase-flutter/issues/399) for platform support progress.
+
 
 ### [Database](https://supabase.com/docs/guides/database)
 
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ final supabase = Supabase.instance.client;
 return supabase.auth.signInWithApple();
 ```
 
-`signInWithApple()` is only supported on iOS. Other platforms can use the `signInWithOAuth()` method to perform Apple login.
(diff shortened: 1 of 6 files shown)


lib/src/supabase_auth.dart
```diff
@@ -350,6 +355,43 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
+
+  /// Signs a user in using native Apple Login.
+  ///
+  /// This method only works on iOS. If you want to sign in a user using Apple
+  /// on other platforms, please use the `signInWithOAuth` method.
+  ///
+  /// This method is experimental as the underlying `signInWithIdToken` method is experimental.
+  @experimental
+  Future<AuthResponse> signInWithApple() async {
+    assert(Platform.isIOS, 'Please use signInWithOAuth for non-iOS platforms');
```
Vinzent03, line 367:
```suggestion
    assert(!kIsWeb && Platform.isIOS, 'Please use signInWithOAuth for non-iOS platforms');
```
I think otherwise the assert itself would throw and I don't know if the message would come through.

Vinzent03 (changes requested):

Vinzent03 (approved):

bdlukaa (commented):
It looks like macOS and Android are also supported in the package. Why do we limit it to iOS only? At least, macOS should be supported as well

https://pub.dev/packages/sign_in_with_apple#supported-platforms

dshukertjr:
@bdlukaa 
The Android setup just essentially the same web based login essentially, so we said people can just use the current flow that we already have to keep things simple. 

macOS, tbh I thought we could worry about it later, but it seems like there is no additional setup to make it work, so let me change the code now to accept macOS right now!

README.md
```diff
@@ -96,33 +96,22 @@ class _MyWidgetState extends State<MyWidget> {
...
 
-return supabase.auth.signInWithIdToken(
-  provider: Provider.apple, 
-  idToken: credential.identityToken,
-  nonce: nonce
-);
+return supabase.auth.signInWithApple();
 ```
 
+`signInWithApple()` is only supported on iOS. Other platforms can use the `signInWithOAuth()` method to perform Apple login.
```
bdlukaa, line 110:
```suggestion
`signInWithApple()` is only supported on iOS and on macOS. Other platforms can use the `signInWithOAuth()` method to perform Apple login.
```

bdlukaa (approved):

lib/src/supabase_auth.dart
```diff
@@ -350,6 +355,44 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
+      ],
+      nonce: hashedNonce,
+    );
+
+    final idToken = credential.identityToken;
+    if (idToken == null) {
+      throw AuthException('Could not find ID Token from generated credential.');
+    }
+
+    return signInWithIdToken(
```
DanMossa, line 385: We don't need a clientId anymore?

DanMossa:
This is phenomenal and I'm really happy to see this happening!

DanMossa (approved):

dshukertjr:
> I believe that Google is migrating their endpoints to include Nonce which should allow us to use the official sign in with google package.
> 
> https://github.com/flutter/flutter/issues/102243
> https://github.com/flutter/flutter/issues/101004

Thanks for sharing this! This is great news! BTW, while I was trying to make Google login work with AppAuth, I found your issues/ comments all over the place! Thank you so so much for working so hard to bring native auth to Supabase!

PR #469, opened 2023-05-16

Vinzent03:


dshukertjr (approved):

dshukertjr:
@Vinzent03 
If it's not broken, let's not try to fix it. We are not losing anything from it, so we will not touch it. All the repositories that I have seen have both pull request and push triggers. I'm sure there is a good reason for it.  I presume it's for a case if the head of main and the root of the branched out code for the PR is different, then the result of the new main branch would be different from what the PR code is.

PR #476, opened 2023-05-18
ci: cleaner test names

Vinzent03:
There were many duplicates in the test names. Sould look much cleaner now.

Additionally I added that the formatter only runs on stable sdk to prevent failing test just because the formatter changed.

### DIFF
--- a/.github/workflows/functions_client_ci.yml
+++ b/.github/workflows/functions_client_ci.yml
@@ -46,6 +46,7 @@ jobs:
           melos bootstrap --ignore="supabase_flutter"
 
       - name: dartfmt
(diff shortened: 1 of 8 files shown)


.github/workflows/functions_client_ci.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           melos bootstrap --ignore="supabase_flutter"
 
       - name: dartfmt
+        if: ${{ matrix.sdk == 'stable'}}
```
dshukertjr, line 49: Love this update!

dshukertjr (approved):
This would make our lives so much better! Thanks for this one!

Issue #104, opened 2022-05-04
Order() and Limit() not working as expected with foreign table


Certinax:
## PostgrestError/failing query with working equal implementation in Javascript

**Disclaimer:** I'm sorry if this has already been raised as an issue or if this isn't the correct place to raise the issue.

I'm working on a query that works as expected in javascript, but when using the same implementation in Flutter I get a PostgrestError. It seems that the issues that are causing the error are the order and limit functions. I'm using a foreign nested table which, from my understanding, is padded with URL-encoded quotes in the final URL that is being sent. These are not added when used with the javascript library.

I have added the query string, both implementations, and the final URL's for comparison. 

**Query**
<details>
  <summary>Show entire query string</summary>

```javascript
`
    grazing_group:grazingGroup(
      id,
      name,
      farms:farm(
        id,
        name,
        color
      ),
      trips:trip(
        id,
        grazing_group_id,
        distance,
        start_time,
        end_time,
        owner:profiles(
          first_name,
          last_name
        ),
        waypoints:location(
          waypoint(
            timestamp,
            latitude,
            longitude
          )
        ),
        events:event(
          id,
          title,
          category,
          description,
          waypoint(
            timestamp,
            latitude,
            longitude
          ),
          images:image(
            img_url
          )
        ),
        flocks:flock(
          id,
          total_sheep,
          note,
          registerPosition:waypoint!flock_register_position_id_fkey(
            timestamp,
            latitude,
            longitude
          ),
          flockPosition:waypoint!flock_flock_position_id_fkey(
            timestamp,
            latitude,
            longitude
          ),
          sheep(
            id,
            color,
            tie_color,
            farm_id,
            note,
            is_injured
          ),
          lambs:lamb(
            id,
            color,
            farm_id,
            note,
            is_injured
          ),
          images:image(
            img_url
          )
        ),
        locations:location(
          waypoint(
            timestamp,
            latitude,
            longitude
          )
        )
      ),
      members:member(
        id,
        admin,
        pending,
        user_id,
        profile:profiles(
          first_name,
          last_name,
          email,
          picture
        )
      )
    )
    `
```

</details>

## Javascript
`Node v17.4.0`
```js
// package.json
"@supabase/supabase-js": "^1.31.1"
```

```javascript
const response = await supabase
  .from("member")
  .select(query)
  .eq("user_id", "myuserid")
  .eq("pending", false)
  .eq("grazing_group.id", "grazingGroupId")
  .order("start_time", {
[...]

PR #477, opened 2023-05-21
fix(postgrest): Remove qoutations on foreign table transforms on 'or'

roneylf:
small correction of an error similar to #104, where the character " generated a bug when converted to the url.
the previous bug fix fixed **Order**, **Limit** and **Range**, but had not fixed **Or**.


dshukertjr (approved):
LGTM! Thanks for the fix!

PR #480, opened 2023-05-24
ci: use melos for example

Vinzent03:
The workflow failes for supabaes_flutter, if the version of supabase changes. https://github.com/supabase/supabase-flutter/actions/runs/5040611117/jobs/9039549791

Instead I'm now using a normal version and include it in melos

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -3,7 +3,6 @@ repository: https://github.com/supabase/supabase-flutter
 
 packages:
   - packages/*
-  - packages/supabase_flutter/example
 
 command:
   version:
--- a/melos.yaml
+++ b/melos.yaml
@@ -3,6 +3,7 @@ repository: https://github.com/supabase/supabase-flutter
 
 packages:
   - packages/*
+  - packages/supabase_flutter/example
 
 command:
   version:
--- a/packages/supabase_flutter/example/pubspec.yaml
+++ b/packages/supabase_flutter/example/pubspec.yaml
@@ -11,8 +11,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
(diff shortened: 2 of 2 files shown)


dshukertjr:
Nice one! Yeah, this was a bit annoying. It seems like a bug on pub. https://github.com/dart-lang/pub/issues/2447

melos.yaml
```diff
@@ -3,6 +3,7 @@ repository: https://github.com/supabase/supabase-flutter
 
 packages:
   - packages/*
+  - packages/supabase_flutter/example
```
dshukertjr, line 6: I tried it locally, and it seemed like just by specifying a version in pubspec.yaml instead of a path was enough, hence we might be able to get rid of this line. Would you be able to confirm and if that's the case, remove this line?

Vinzent03: You are right, it works without melos as well. I inspired from [flutterfire melos.yaml](https://github.com/firebase/flutterfire/blob/master/melos.yaml). The way it currently is, I think melos would update the version of `supabase_flutter` in the example pubspec when we update `supabase_flutter`. Though I don't know if it occurs in the changelog or similar as well, so I'm fine with deleting it as well and time to time update the version manually.

Vinzent03:
By the way, did you manually publish the latest versions? I can't find the actions that should have run. At least they should have been triggered.

dshukertjr:
@Vinzent03 Yup, I manually published the versions. I was going to look into it when I have time, but melos makes it easy to publish all the packages anyway, so it was low priority to me. If you want to look into it, that would be great, but again, I don't  it's really not a high priority.

dshukertjr (approved):

PR #482, opened 2023-05-25
docs(supabase): update comment docs on stream

dshukertjr:

After seeing [this issue](https://github.com/supabase/supabase/issues/14579), I thought it might make sense to add some more details on the comment docs of `.stream()`.

### DIFF
--- a/packages/supabase/lib/src/supabase_query_builder.dart
+++ b/packages/supabase/lib/src/supabase_query_builder.dart
@@ -32,7 +32,7 @@ class SupabaseQueryBuilder extends PostgrestQueryBuilder {
 
   /// Returns real-time data from your table as a `Stream`.
   ///
-  /// Realtime is disabled by default for new tables. You can turn it on by [managing replication](/docs/guides/realtime/extensions/postgres-changes#replication-setup).
+  /// Realtime is disabled by default for new tables. You can turn it on by [managing replication](https://supabase.com/docs/guides/realtime/extensions/postgres-changes#replication-setup).
   ///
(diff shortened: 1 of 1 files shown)


packages/supabase/lib/src/supabase_query_builder.dart
```diff
@@ -30,18 +30,20 @@ class SupabaseQueryBuilder extends PostgrestQueryBuilder {
           isolate: isolate,
         );
 
-  /// Notifies of data at the queried table
+  /// Returns real-time data from your table as a `Stream`.
   ///
-  /// [primaryKey] list of name of primary key column(s).
+  /// Realtime is disabled by default for new tables. You can turn it on by [managing replication](/docs/guides/realtime/extensions/postgres-changes#replication-setup).
```
Vinzent03, line 35:
```suggestion
  /// Realtime is disabled by default for new tables. You can turn it on by [managing replication](https://supabase.com/docs/guides/realtime/extensions/postgres-changes#replication-setup).
```

Vinzent03 (changes requested):

packages/supabase/lib/src/supabase_query_builder.dart
dshukertjr: Oh yearh! Thanks for this!

Vinzent03 (approved):

Vinzent03:
Isn't the type `docs` better  than `fix` here? https://melos.invertase.dev/guides/automated-releases#conventional-commits

dshukertjr:
@Vinzent03 Sure!

PR #492, opened 2023-05-29

dshukertjr:
- functions_client@1.3.1
 - gotrue@1.8.1
 - postgrest@1.3.1
 - storage_client@1.4.1
 - supabase@1.9.1
 - supabase_flutter@1.10.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,53 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2023-05-29
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - There are no breaking changes in this release.
+
+Packages with other changes:
+
+ - [`functions_client` - `v1.3.1`](#functions_client---v131)
+ - [`gotrue` - `v1.8.1`](#gotrue---v181)
+ - [`postgrest` - `v1.3.1`](#postgrest---v131)
+ - [`storage_client` - `v1.4.1`](#storage_client---v141)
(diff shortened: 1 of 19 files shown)


dshukertjr:
Also, for the next release I was thinking about updating the minimum Flutter version to v3.0.0 and minimum Dart version to 2.17.0 since it has been a year since they were released, but do we have any objections/ concerns?

phamhieu:
>Also, for the next release I was thinking about updating the minimum Flutter version to v3.0.0 and minimum Dart version to 2.17.0 since it has been a year since they were released, but do we have any objections/ concerns?

Yes we should. release a major version with breaking change note; should be fine

phamhieu (approved):

Vinzent03:
>  release a major version with breaking change note; should be fine

@dshukertjr If we do it that way, we could solve some issues from #287

dshukertjr:
@Vinzent03 Yeah, I was thinking roughly one major release per year at maximum. We released our v1 in October, so maybe we can do a developer preview release few months prior to that and can start working on those changes! Unlike the previous one, I feel like most of the change we want to ship are less breaking, so I think the confusions around it would be smaller!

Vinzent03:
Does that mean you don't plan to bump the major version for the Dart sdk bump or do you want to wait util october then?

dshukertjr:
> Does that mean you don't plan to bump the major version for the Dart sdk bump

@Vinzent03 We do have to bump the major version if any of the sub libraries have a major version bump. The schedule is yet to be determined. We have to talk internally and see what the js SDK is doing. The timeline shared above is just a rough one that I had in my mind, but it is subject to change.

@phamhieu We do support the latest versions by the way! I was just debating whether to bump the minimum supported versions or not.

PR #501, opened 2023-06-04
fix(gotrue): catch some errors in signOut

### LINKED ISSUE #500: Consider expose _removeSession function
Apple store requires us to provide Account Deletion to users, i tried to implement using Edge Functions, but i can't clear session in user device. If i call `auth.signOut`, i always got `Unhandled Exception: AuthException(message: User not found, statusCode: 404)`. So i suggest to change `_removeSession` to `removeSession` to allows developers call.

Vinzent03:
close #500
I'm catching specifc errors on sign out like in [gotrue-js](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L1059) to prevent errors being thrown in `signOut()`.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -613,9 +613,8 @@ class GoTrueClient {
       try {
         await admin.signOut(accessToken);
       } on AuthException catch (error) {
-        // ignore 401s since an invalid or expired JWT should sign out the         
+        // ignore 401s since an invalid or expired JWT should sign out the current session
         // ignore 404s since user might not exist anymore
-current session
         if (error.statusCode != '401' && error.statusCode != '404') {
           rethrow;
         }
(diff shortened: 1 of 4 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -610,7 +610,17 @@ class GoTrueClient {
...
+        return await admin.signOut(accessToken);
+      } on AuthException catch (error) {
+        // ignore 404s since user might not exist anymore
+        // ignore 401s since an invalid or expired JWT should sign out the current session
+        if (error.statusCode == '401' || error.statusCode == '404') {
+          return;
+        } else {
+          rethrow;
+        }
+      }
```
dshukertjr, line 623: Just a minor cleanup

```suggestion
      try {
        await admin.signOut(accessToken);
      } on AuthException catch (error) {
        // ignore 401s since an invalid or expired JWT should sign out the         
        // ignore 404s since user might not exist anymore
current session
        if (error.statusCode != '401' && error.statusCode != '404') {
          rethrow;
        }
      }
```

dshukertjr (commented):
Thanks for the quick fix!

packages/gotrue/lib/src/gotrue_client.dart
Vinzent03: Ah great catch!

dshukertjr (approved):
Looks awesome!

PR #507, opened 2023-06-09
refactor(realtime_client): Add docs on `.subscribe()` and give callback parameters names

dshukertjr:

- Add some basic docs on `.subscribe()` method
- Give names to the parameters of `callback` on `.subscribe()`

### DIFF
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -180,7 +180,7 @@ class RealtimeChannel {
     }
   }
 
-  /// Subscribe registers your client with the server
+  /// Subscribes to receive real-time changes
   ///
   /// Pass a [callback] to react to different status changes. Values of `status`
   /// can be `SUBSCRIBED`, `CHANNEL_ERROR`, `CLOSED`, or `TIMED_OUT`.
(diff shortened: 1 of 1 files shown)


packages/realtime_client/lib/src/realtime_channel.dart
```diff
@@ -180,8 +180,16 @@ class RealtimeChannel {
     }
   }
 
-  void subscribe(
-      [void Function(String, [Object?])? callback, Duration? timeout]) {
+  /// Subscribe registers your client with the server
```
Vinzent03, line 183: I'm not sure what this sentence means. Do you mean with "Subscribe" the method name?  Then I would make it lower case and wrap it in `[]` or remove it completely.

Vinzent03 (changes requested):

packages/realtime_client/lib/src/realtime_channel.dart
Vinzent03: That's better. Sometimes I see `realtime` and sometimes `real-time`. Is there any reason or are both ways fine?

dshukertjr: Yeah, "Realtime" is a name of a service that Supabase offers, and "real-time" is a term to describe multiple events happening simultaneously. "realtime" is the grammatically incorrect version of "real-time". I'm sure within Supabase products, there are tons of those typos though 😂

Vinzent03 (approved):

PR #509, opened 2023-06-09
fix(supabase): add stackTrace on stream exception

dshukertjr:

There is a handy method called `_addException` that adds the error on the streamController and passes a current stack trace if there are none provided. This PR uses the method instead of directly passing the error to the stream.


Vinzent03:
I now noticed, don't we have to use the package as scope in the pr title?

dshukertjr:
@Vinzent03 I think Melos auto detects which packages were updated in a commit, so we should be good as far as releasing packages go, but for clarity we should. I have updated the title.

Vinzent03:
Ah okay, thanks!

Vinzent03 (approved):

PR #512, opened 2023-06-12
fix(gotrue): Add missing members on `User` json serialization

dshukertjr:

Because some members are missing from the `Session` object and its sub members, session that is recovered from local storage is missing some informations such as MFA AAL. This PR fixes the issue.

- Adds missing members on `User` object json serialization
- Adds missing members on `==` operator and `hashcode` definitions for Session and its sub property classes
- Adds a test to try out json serialization & deserialization after signing in using MFA


packages/gotrue/lib/src/types/user.dart
```diff
@@ -54,54 +57,121 @@ class User {
     if (json['id'] == null) {
       return null;
     }
+
     return User(
-      id: json['id'] as String,
-      appMetadata: json['app_metadata'] as Map<String, dynamic>,
+      id: json['id'] ?? '',
+      appMetadata: json['app_metadata'] as Map<String, dynamic>? ?? {},
```
Vinzent03, line 63:
```suggestion
      appMetadata: json['app_metadata'] ?? {},
```
Im pretty sure the `as Map` in the line below is unnecessary as well.

bdlukaa: Not really. `json['app_metadata']` may be a `Map<dynamic, dynamic>` when parsing from a parent json

bdlukaa (approved):

packages/gotrue/lib/src/types/user.dart
dshukertjr: Let's just keep it! Either way it won't hurt!

Vinzent03: @bdlukaa That wouldn't work. You can't change the generics on `Map` via `as`, because these are separate types. [Here](https://zapp.run/edit/dart-z25w06ir25x0?entry=lib/main.dart&file=lib/main.dart) is a quick example.

PR #517, opened 2023-06-21
feat(gotrue): add reauthenticate and resend method

Vinzent03:
I've noticed gotrue-js contains the `reauthenticate` and `resend` method, but `gotrue-dart`does not. 
[gotrue-js pr](https://github.com/supabase/gotrue-js/pull/688) for reauthenticate.

Though I'm very unsure about the parameter for `resend`. It uses different classes and enums for phone and email


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -479,8 +479,52 @@ class GoTrueClient {
...
+      '$_url/reauthenticate',
+      RequestMethodType.get,
+      options: options,
+    );
+  }
+
+  ///Resends an existing signup confirmation email, email change email, SMS OTP or phone change OTP.
+  ///
+  ///Use [EmailResendParams] or [PhoneResendParams] to specify the type of resend.
+  Future<ResendResponse> resend(ResendParams params) async {
```
dshukertjr, line 504: I really like what you have going on here with the enum and two classes extending a base abstract class, but I personally think it would be easier for the users to use this method if the parameters were simple named parameters like this and add comment docs and assertion to make sure the user knows what value to pass or not pass. 

```suggestion
  Future<ResendResponse> resend({
    String? email,
    String? phone,
    OtpType type,
    String? captchaToken,
  }) async {
```

Vinzent03: Yeah I thought that as well and wasn't sure. Will change hat.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -479,8 +479,69 @@ class GoTrueClient {
...
+  Future<ResendResponse> resend({
+    String? email,
+    String? phone,
+    required OtpType type,
+    String? captchaToken,
+  }) async {
+    assert((email != null && phone == null) || (email == null && phone != null),
+        '`email` or `phone` needs to be specified.');
+
+    _removeSession();
```
Vinzent03, line 513: I've copied that from gotrue-js, but I think always removing the session is wrong, isn't it? If the use just wants to update their email and has to resend that, we don't want to remove the session?

Best way is probably like in `verifyOTP`,right?
```suggestion
    if (type != OtpType.emailChange && type != OtpType.phoneChange) {
      _removeSession();
    }
```

dshukertjr: Yeah, I think you are correct! Maybe you could also comment in the gotrue PR to let them know about this one?

packages/gotrue/lib/src/types/user.dart
```diff
@@ -268,3 +268,12 @@ class UserIdentity {
...
 }
+
+class ResendResponse {
+  /// Only set for phone resend
+  String? messageId;
+
+  ResendResponse({
+    this.messageId,
+  });
+}
```
dshukertjr, line 279: Just a small thing, but maybe this should be in `auth_response.dart` file?

dshukertjr (commented):
Thank you so much for the detailed check! The PR looks amazing! Just had one small thing that I missed earlier.

packages/gotrue/lib/src/types/user.dart
Vinzent03: Oh sure, fixed.

dshukertjr:
Let's wait to merge this until we confirm with the gotrue-js team to see if the remove session thing was intentional or not.

Vinzent03:
@dshukertjr This can be merged now, right?

dshukertjr (approved):
🎉🚀

haykodarb:
Is it possible this stopped working due to changes in the backend? 

I'm getting the following exception 

AuthException(message: Missing one of these types: signup, email_change, sms, phone_change, statusCode: 400)

when using 

`  static Future<BackendResponse> resendEmail({required String email}) async {
    try {
      final SupabaseClient client = Supabase.instance.client;

      final ResendResponse response = await client.auth.resend(
        type: OtpType.email,
        email: email,
      );

      return BackendResponse(success: true, payload: response.messageId);
    } catch (e) {
      return BackendResponse(success: false, payload: e);
    }
  }`

Vinzent03:
@haykodarb  So your project is hosted on supabase and it already worked for you and stopped working now or is your project new and doesn't work?

haykodarb:
The project is a couple of months old, hosted on Supabase and I'm just now implementing the resend email functionality. 

Testing the function threw the previously mentioned exception but now I'm getting the usual 

```AuthException(message: Email rate limit exceeded, statusCode: 429)```

Which is fine.

Vinzent03:
I guess your project is on the free plan? Please try to pause and unpause your project to hopefully apply the latest gotrue version.

haykodarb:
Yes, I'll try that, thank you.

haykodarb:
Paused and unpaused project, still getting this exception.

```AuthException(message: Missing one of these types: signup, email_change, sms, phone_change, statusCode: 400)```

After 3 times of getting that exception I get rate limited without receiving any emails. 

```AuthException(message: Email rate limit exceeded, statusCode: 429)```

dshukertjr:
@haykodarb 
I know it will give you an deprecation warning, but could you try calling `resend` method with type of `OtpType.signup` or `OtpType.emailChange`, whichever is appropriate in your case?

PR #515, opened 2023-06-15
feat(supabase_flutter): Add `signInWithGoogle()` method to perform native Google login

dshukertjr:
## What kind of change does this PR introduce?

- Adds native Google login support for iOS and Android via `signInWithGoogle()` method.
- Add instruction to setup native login on readme.md
- Some cleanups on the auth section of the readme
```dart
await supabase.auth.signInWithGoogle(
  iosClientId: 'IOS_CLIENT_ID',
  androidClientId: 'ANDROID_CLIENT_ID',
);
```

The `package_info_plus` dependency requires Android SDK version 19, and Flutter by default uses 16. This may be an issue, but since [Play store requires 19](https://developers.google.com/android/guides/setup#:~:text=A%20compatible%20Android%20device%20that,Google%20Play%20Store%20app%20installed.) anyway, I think it's fine.

PR #518, opened 2023-06-22
Added signInWithCredential method to perform native

trongtindev:

Added signInWithCredential method like [firebase_auth](https://pub.dev/packages/firebase_auth) (Current support Apple, Google and Facebook).

## Usage
```dart
// package: sign_in_with_apple
var credential = await SignInWithApple.getAppleIDCredential();
var appleAuthCredential = AppleAuthProvider.credential(
  credential.identityToken!,
  credential.authorizationCode,
);
await Supabase.instance.client.auth.signInWithCredential(appleAuthCredential);
```
```dart
// package: google_sign_in
var client = GoogleSignIn();
var result = await client.signIn();
var authentication = await result!.authentication;

var appleAuthCredential = AppleAuthProvider.credential(
  credential.identityToken!,
  credential.authorizationCode,
);
var googleAuthCredential = GoogleAuthProvider.credential(
  idToken: authentication.idToken,
  accessToken: authentication.accessToken,
);
await Supabase.instance.client.auth.signInWithCredential(googleAuthCredential);
```

```dart
// package: flutter_facebook_auth
final LoginResult result = await FacebookAuth.instance.login();
var facebookAuthCredential = FacebookAuthProvider.credential(result.accessToken!.token);

await Supabase.instance.client.auth.signInWithCredential(facebookAuthCredential );
```


dshukertjr (changes requested):
Thanks for this nice PR! I really like the direction of the `signInWithCredential` instead of providing individual auth providers like `signInWithApple()`.

There might be one problem with this, and that it's not using nonce any where. I believe using the [google_sign_in](https://pub.dev/packages/google_sign_in) package generates nonce, but we cannot specify the nonce on our end, so we are going another direction and using app auth in [this PR](https://github.com/supabase/supabase-flutter/pull/515), which I'm sorting out few things right now.

If you could provide screen recording of each providers on both iOS and Android, that would be great.

trongtindev:
@dshukertjr I think we should temporarily use `flutter_appauth` until the issue [https://github.com/google/GoogleSignIn-iOS/pull/244](https://github.com/google/GoogleSignIn-iOS/pull/244) is resolved. Another problem is that currently, `idToken` only supports `Google` and `Apple`, but I need to add more. Additionally, some providers like Facebook return an `accessToken` format, which requires an endpoint like `signInWithAccessToken` or `grant_type=access_token`. I have tried searching, but it doesn't seem to be open source.

dshukertjr:
@trongtindev 
Thanks for the additional work, but I want to stop you there. There are currently no endpoint on gotrue that signs a user in just with an access token provided by third party login provider. I will bring it up with our Auth team, but the auth team has a lot on their plates, so no guarantee when it might happen or it might not happen at all.

After some thoughts, I don't think we want to introduce a new `signInWithCredential` methods, as the current `signInWithIdToken` method is fairly simple, and introducing another layer on top of it might not make things necessary simpler I think.

With that, even though I really like this massive contribution, I would like to close this PR. Thanks for all the work that you have put in!

Issue #399, opened 2023-03-13
Full Native Auth support


DanMossa:
The criteria for having a successful native authentication process is the ability to do the following:

- [x] Sign in on Google working on Android, iOS, and Web
- [x] Sign in with Apple working on Android, iOS, and Web
  - [x] Sign in with Apple with multiple bundle IDs
- [x] When creating an account with social auth, you can then link your email with it.
- [x] When creating an account with email first, you can then link your social auth to it.

## Sign in on Google working on Android, iOS, and Web
When using sign in with Google, you're currently limited to having sign in with Google on either Android and Web, or iOS. The reason for this is because on GCP, you create credentials and you choose the type of device that the OAuth client is. The web application credentials work for both Android and Web where as for iOS, you're required to use the iOS credentials. Supabase only has a slot for a single client ID/Secret.

This PR is also required for us to pass in a Nonce when using iOS
https://github.com/google/GoogleSignIn-iOS/pull/244

## Sign in with Apple working on Android, iOS, and Web
I was able to get iOS working on iPhone, but I'm having trouble getting Sign in with Apple working on the web as well. These are the instructions I'm following but I'll update this when I / others have better luck setting this up
https://pub.dev/packages/sign_in_with_apple

## When creating an account with social auth, you can then link your email with it.
When creating an account with Google the user's email address is registered inside the `auth.users` table. A user can then link their accounts by following the forgot password route with the same email.

When creating an account with Apple, depending on the user's preferences, either the plain text email or the private email is registered inside `auth.users`. A user can then link their accounts by following the forgot password route with the same email. This is only possible if the user does not use Apples's private email, but Apple users are aware of this.

## When creating an account with email first, you can then link your social auth to it.
This is not possible at the moment. We can an error back stating there's a duplicate of the email already. I've read that toggling `isSSO` fixes this but I haven't tested it yet.

dshukertjr:
Mentioned it [here](https://github.com/supabase/supabase-flutter/issues/5#issuecomment-1467451495) as well, but auth team will be working on a solution to fix the Google login platform problem. 

We will probably add a `clientId` parameter on the OIDC endpoint so that when signing in with Android or iOS we can pass it as a parameter.

gregorym:
> 

Does it mean we will be able to support authentication on web and mobile with different Google OAuth IDs?

dshukertjr:
@gregorym Yup!

medclans:
@dshukertjr 
[...]

PR #520, opened 2023-06-23
feat(gotrue): add accessToken to `signInWithIdToken` method

dshukertjr:

Adds `accessToken` parameter to `signInWithIdToken` method. 

Related https://github.com/supabase/gotrue-js/pull/690
Related https://github.com/supabase/supabase-flutter/issues/399

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -309,6 +309,7 @@ class GoTrueClient {
     required String idToken,
     String? nonce,
     String? captchaToken,
+    String? accessToken,
   }) async {
     _removeSession();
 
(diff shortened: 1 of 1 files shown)


Vinzent03:
I don't really understand the documentation from gotrue-js for `access_token`, because I've never worked with native auth or similar, but I guess this should also be included in our doc.

> If the ID token contains an `at_hash` claim, then the hash of this value is compared to the value in the ID token.

dshukertjr:
@Vinzent03 Thanks! I've added it!

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -302,11 +302,21 @@ class GoTrueClient {
   /// The [idToken] is verified for validity and a new session is established.
   /// This method of signing in only supports [Provider.google] or [Provider.apple].
   ///
+  /// If the ID token contains an `at_hash` claim, then [accessToken] must be
+  /// provided to compare its hashthe hash with the value in the ID token.
```
Vinzent03, line 306:
```suggestion
  /// provided to compare its hash with the value in the ID token.
```

Vinzent03 (changes requested):

Vinzent03 (approved):

Issue #314, opened 2022-12-23
initialize client with multiple schemas


chirastefan:
Flutter client doesn't seem to support initializing Supabase with multiple schemas and then querying based on those schemas. One client = one schema.

the code below throws this error `This instance is already initialized`
```
var schemaOne = await Supabase.initialize(url: endpoint, anonKey: anonKey, schema: 'schema_one', debug: kDebugMode);
var schemaTwo = await Supabase.initialize(url: endpoint, anonKey: anonKey, schema: 'schema_two', debug: kDebugMode);
```
1. A workaround for this is to dispose of the client and re-init it every time a new query is made to a different schema. IMO this doesn't sound that good. I guess it raises some performance issues and other side effects, but maybe not and that's the intended behavior.

2. Another workaround would be to not use the `Supabase.initialize()` method and init the `SupabaseClient` directly:
```
await SupabaseAuth.initialize();
var schemaOne = SupabaseClient(url, anonKey, schema: 'schema_one');
var schemaTwo = SupabaseClient(url, anonKey, schema: 'schema_two');
```
This is a bit of a hack also, this way the `log()` method from `Supabase` class doesn't know about the `_debugEnable` flag.
Is there another way of doing this?

Edit: I tried solution 2 and it doesn't work. `Supabase.initialize()` must be called, so the way I did it is to call that method which initializes Supabase with the default scheme - `public`  and then separately initialize other schemas with `SupabaseClient()`

```
enum SupabaseSchemas { public, schema1, schema2 }

late final Map<SupabaseSchemas, SupabaseClient> schemas = {};

Future<void> _initSchemas() async {
    schemas[SupabaseSchemas.public] = (await Supabase.initialize(url: endpoint, anonKey: anonKey, debug: kDebugMode)).client;
    final List<SupabaseSchemas> customSchemas = SupabaseSchemas.values.filterNot((sch) => sch == SupabaseSchemas.public).toList();

    for (final schema in customSchemas) {
      schemas[schema] = schemas[schema] == null || schemas[schema] is! SupabaseClient
          ? (SupabaseClient(endpoint, anonKey, schema: schema.name))
          : schemas[schema]!;
    }
  }
```

Currently, the above solution seems to work, but I'm not sure if this is the proper way to do it or if there's another way of doing this. This works for insert/upsert/update queries, but for `select` I don't have a solution to query the from other schemas.
An ideal solution would be to init the client with the custom schemas and then for every query to be able to set the schema as an option.

[...]

PR #525, opened 2023-06-28
feat(postgrest, supabase): add `useSchema()` method for making rest API calls on custom schema.

dshukertjr:

Allows developers to easily query custom schema.
Fixes https://github.com/supabase/supabase-flutter/issues/314

There isn't a way to query other schemas once Supabase client is initialized. Some workarounds that I can think about right now would be
- Don't use custom schemas
- Create postgres views/ functions in the public schema for every custom schema table
- `dispose` and re-initialize the client every time when querying custom schema
All of them are kind of painful to implement.

## What is the new behavior?

Developers can query custom schema like this:
```dart
final data = await supabase.useSchema('custom_schema').from('table').select();
```

## Additional Context

Since `schema` name is already taken, we couldn't name the method `schema()` like how the js SDK does it.

### DIFF
--- a/packages/postgrest/lib/src/postgrest.dart
+++ b/packages/postgrest/lib/src/postgrest.dart
@@ -50,12 +50,25 @@ class PostgrestClient {
   }
 
   /// Perform a table operation.
-  PostgrestQueryBuilder<void> from(String table, {String? schema}) {
+  PostgrestQueryBuilder<void> from(String table) {
     final url = '${this.url}/$table';
     return PostgrestQueryBuilder<void>(
       url,
       headers: {...headers},
-      schema: schema ?? this.schema,
+      schema: schema,
+      httpClient: httpClient,
+      isolate: _isolate,
+    );
+  }
+
+  /// Select a schema to query or perform an function (rpc) call.
+  ///
+  /// The schema needs to be on the list of exposed schemas inside Supabase.
+  PostgrestClient setSchema(String schema) {
+    return PostgrestClient(
+      url,
+      headers: {...headers},
+      schema: schema,
       httpClient: httpClient,
       isolate: _isolate,
     );
--- a/packages/postgrest/lib/src/postgrest.dart
+++ b/packages/postgrest/lib/src/postgrest.dart
@@ -50,12 +50,12 @@ class PostgrestClient {
   }
 
   /// Perform a table operation.
-  PostgrestQueryBuilder<void> from(String table) {
+  PostgrestQueryBuilder<void> from(String table, {String? schema}) {
     final url = '${this.url}/$table';
     return PostgrestQueryBuilder<void>(
       url,
       headers: {...headers},
-      schema: schema,
+      schema: schema ?? this.schema,
       httpClient: httpClient,
       isolate: _isolate,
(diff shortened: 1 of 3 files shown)


Vinzent03:
Is there a reason you are adding that as a parameter to from instead of a new method as it seems proposed in https://github.com/supabase/postgrest-js/issues/280 ?

Additionally, I think `rpc` needs that param as well.

dshukertjr:
@Vinzent03 No, I had forgotten about that one. Thanks for bringing it up.

Vinzent03:
`.setSchema()` sounds like you would change the schema of the existing postgrest client. What about just `.schema()`. That's what they [use in js](https://github.com/supabase/postgrest-js/pull/455/files) now as well.

dshukertjr:
@Vinzent03 
I should have mentioned it in the PR somewhere, but the `schema` name is already taken by the `schema` property of `SupabaseClient` and the `PostgrestClient`, so had to go with something else. In the next major release though, we should be able to rename it to `schema()` after making the `schema` property private.

Vinzent03:
Ah okay. I would choose `useSchema()` or `fromSchema()`. I tend towards `fromSchema()`, because it's similar to `from()` and that's okay, because they both do pretty much the same.

bdlukaa:
I like `useSchema` better. setSchema can be misleading

dshukertjr:
Thanks for the comments. It was very close for me between `setSchema()` and `useSchema()` for me, so I'd say `useSchema()` is the winner. Also want to hear your opinion on another thing. What do you think about a syntax like this:

```dart
.from('table', schema: 'my_schema')
```

which is what I had originally on this PR. I know it's not at all similar to the js implementation, but I think it's cleaner, because we don't have extra words like `use` or `set` with the `schema` keyword. I also slightly don't like the fact that you can chain `useSchema` like this:

```dart
supabase.useSchema('my_schema').useSchema('another_schema')...
```

It means the IDE will suggest `useSchema()` after calling `useSchema` like this. 
Screenshot 2023-08-07 at 23 34 48

Vinzent03:
Using the current workflow allows using one postgrest client with a specific schema for multiple calls. Adding it to `from` means you have to repeat it always. So I would stay with the current approach.

dshukertjr:
Cool, in that case this PR should be ready for review.

Vinzent03 (approved):

garysm:
@dshukertjr Can you give an example of what the recommended permissions should be for a new schema for the Flutter client? For example, a table that would only be available to `postgres`, `service_role`, and authenticated users.  I have tried adding these myself but I still get a `PGRST106` error.

dshukertjr:
@garysm 
Have you added your new schema to [search path](https://supabase.com/dashboard/project/_/settings/api) in your Supabase dashboard under settings>API>Extra Search Path?

garysm:
> @garysm Have you added your new schema to [search path](https://supabase.com/dashboard/project/_/settings/api) in your Supabase dashboard under settings>API>Extra Search Path?

No, I've been developing locally, but I assume that can be configured in `config.toml`.  I'll give that a shot.  Thanks!

PR #542, opened 2023-07-10
fix: update the google auth setup instruction to use reversed client ID

dshukertjr:

When there are underscores used in the application name for an Android app, the redirect URL in the form of `'$applicationId:/google_auth'` becomes invalid. This PR updates the Google login instructions to a one that works even the application name contains underscores.

### DIFF
--- a/packages/supabase_flutter/README.md
+++ b/packages/supabase_flutter/README.md
@@ -91,7 +91,7 @@ First, you need to create a client ID in your Google Cloud console and add them 
 - [Steps to obtain Android client ID](https://developers.google.com/identity/sign-in/android/start-integrating#configure_a_project)
 - [Steps to obtain iOS client ID](https://developers.google.com/identity/sign-in/ios/start-integrating#get_an_oauth_client_id)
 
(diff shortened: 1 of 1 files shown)


packages/supabase_flutter/README.md
```diff
@@ -124,15 +124,8 @@ Future<AuthResponse> signInWithGoogle() {
...
-  ///
-  /// Application ID or Bundle ID for your app.
-  /// If the application ID and Bundle ID is different for Android and iOS,
-  /// make sure the value here matches the platform you are running on.
-  const applicationId = 'com.supabase.example';
-
-  /// Fixed value for google login
-  const redirectUrl = '$applicationId:/google_auth';
+  /// reverse DNS form of the client ID + `:/` is set as the redirect URL
+  final redirectUrl = '${clientId.split('.').reversed.join('.')}:/';
```
Vinzent03, line 128: Is there no `google_auth` required at the end?

dshukertjr: Yeah, I misread it, but `google_auth` wasn't a requirement even for the first method either. 
https://stackoverflow.com/questions/30194644/how-to-handle-redirect-uri-mismatch-error-when-android-app-obtains-offline-acces

Vinzent03 (approved):

PR #543, opened 2023-07-11
fix(postgrest): update docs to mention views

dshukertjr:

A user reported [on Twitter](https://twitter.com/millergodev/status/1678435300157292545) that our docs does not mention that the postgrest methods can be used on views. This PR adds the missing mention.


Vinzent03:
Aren't views read only? So no delete, update, upsert queries.

dshukertjr:
@Vinzent03 This one might be a niche use case, but you could actually create views where you can perform delete, update or inserts when satisfying certain conditions! But yeah, in typical use cases views are likely to be read only. 
https://www.postgresqltutorial.com/postgresql-views/postgresql-updatable-views/

I took these updated comments from the [JS doc](https://supabase.com/docs/reference/javascript/insert)

PR #547, opened 2023-07-13
Events per second comment

KyleParker-Gongyuan:
Not well documented.

Comment/comment documentation 

## Additional context

From what I can tell it's automatically 10 per second server-side

### DIFF
--- a/packages/supabase/lib/src/realtime_client_options.dart
+++ b/packages/supabase/lib/src/realtime_client_options.dart
@@ -1,5 +1,6 @@
 /// {@template realtime_client_options}
-/// Options to pass to the RealtimeClient
+/// Options to pass to the RealtimeClient.
+/// Default is 10 per second
 /// {@endtemplate}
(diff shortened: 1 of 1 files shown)


packages/supabase/lib/src/realtime_client_options.dart
```diff
@@ -1,5 +1,6 @@
 /// {@template realtime_client_options}
-/// Options to pass to the RealtimeClient
+/// Options to pass to the RealtimeClient.
+/// Default is 10 per second
```
dshukertjr, line 3: Thanks for this! Maybe we should add the note about the default value on `eventsPerSecond` property instead of on `RealtimeClientOptions` itself. 

Also now that you mention it, maybe we can clarify the wording on the comments of `eventsPerSecond`. Do you think something like "How many broadcast events the RealtimeClient can push in a second" would be better?

dshukertjr (approved):

PR #550, opened 2023-07-14
feat(postgrest): updates for postgREST 11

Vinzent03:
# Changes
- add `any` and `all` version to `like` and `ilike`
- set columns search parameter in bulk insert/upsert to allow missing column in one row
- add `defaultToNull` to use default value instead of null for missing value in bulk insert/upsert

# Ressources
- [postgREST doc](https://postgrest.org/en/stable/references/api/tables_views.html#bulk-insert)
- [postgrest-js pr](https://github.com/supabase/postgrest-js/pull/417) for postgrest 11 changes
- [postgrest-js pr](https://github.com/supabase/postgrest-js/pull/176)  for columns for bulk insert

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Monorepo containing all [Supabase](https://supabase.com/) libraries for Flutter.
 - [supabase_flutter](https://github.com/supabase/supabase-flutter/tree/main/packages/supabase_flutter)
 - [yet_another_json_isolate](https://github.com/supabase/supabase-flutter/tree/main/packages/yet_another_json_isolate)
 
-- Documentation: https://supabase.com/docs/reference/dart/introduction
+Documentation: https://supabase.com/docs/reference/dart/introduction
 
 ---
 
--- a/packages/postgrest/lib/src/postgrest_query_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_query_builder.dart
@@ -124,7 +124,7 @@ class PostgrestQueryBuilder<T> extends PostgrestBuilder<T, T> {
     _body = values;
 
     if (values is List) {
-      _setColumsSearchParam(values);
+      _setColumnsSearchParam(values);
     }
 
     return PostgrestFilterBuilder<T>(this);
@@ -182,7 +182,7 @@ class PostgrestQueryBuilder<T> extends PostgrestBuilder<T, T> {
     }
 
     if (values is List) {
-      _setColumsSearchParam(values);
+      _setColumnsSearchParam(values);
     }
 
     _body = values;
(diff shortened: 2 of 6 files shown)


Vinzent03:
With the update to postgREST v11 the test `Default http client row level security error` somehow fails now, because the call now returns null/ empty list when appending `select()` instead of throwing an error. I can't find this change in the changelog though. Should I just remove the test?

README.md
```diff
@@ -11,7 +11,7 @@ Monorepo containing all [Supabase](https://supabase.com/) libraries for Flutter.
 - [supabase_flutter](https://github.com/supabase/supabase-flutter/tree/main/packages/supabase_flutter)
 - [yet_another_json_isolate](https://github.com/supabase/supabase-flutter/tree/main/packages/yet_another_json_isolate)
 
-- Documentation: https://supabase.com/docs/reference/dart/introduction
+Documentation: https://supabase.com/docs/reference/dart/introduction
```
Vinzent03, line 14: Because of the empty line in that list, it renders a loose Markdown list, causing huge spacing between the items. By removing "Documentation"  from the list, it is properly spaced.

dshukertjr:
Thanks for this! I am a bit swamped today, but will take a look tomorrow!

dshukertjr (commented):
The PR is spotless as usual 🚀 Just the one thing about the default value for `defaultToNull` parameter gets me. I have left a comment here, so let's see what the js team says. 
https://github.com/supabase/postgrest-js/pull/417/files#r1268195130

packages/postgrest/lib/src/postgrest_query_builder.dart
```diff
@@ -94,6 +94,8 @@ class PostgrestQueryBuilder<T> extends PostgrestBuilder<T, T> {
   ///
   /// By default no data is returned. Use a trailing `select` to return data.
   ///
+  /// [defaultToNull] Make missing fields default to `null`. Otherwise, use the default value for the column.
```
dshukertjr, line 97: Could we update the comments to something like this to make it clear that it applies to only when inserting multiple rows in bulk?

```suggestion
  /// When inserting multiple rows in bulk, [defaultToNull] can be used to set the values of the missing fields to be either `null` or the default value for the column.
```

But even this is still kind of unclear. This new behavior is a bit tricky, isn't it 😂

dshukertjr (approved):
The code looks great! Feel free to edit the comments to whatever you think best communicates the behavior to the users!

dshukertjr (changes requested):
Oops, could you check why the tests might be failing? It seems like it has nothing to do with any of the changes in this PR though.

Vinzent03:
@dshukertjr See https://github.com/supabase/supabase-flutter/pull/550#issuecomment-1638224960 above for the test failing. With postgrest 9 or 10 it works, but 11 doesn't.

packages/postgrest/lib/src/postgrest_query_builder.dart
Vinzent03: Great idea! I've added another sentence for single row insertions in https://github.com/supabase/supabase-flutter/pull/550/commits/8686868f02f273fd105b66a323e53b9a7edf323a

dshukertjr:
@Vinzent03 
Sorry for missing your comment about the failing test. I think it's alright to just remove the test.

dshukertjr (approved):
LGTM! Thanks for the amazing work!

PR #558, opened 2023-07-17
fix(realtime): use access token from headers

### LINKED ISSUE #557: supabase_flutter stream not work with RLS custom auth jwt
**Describe the bug**
When we use the stream() method with a policy defined in Row Level Security that uses a field defined in the JWT token, no data is returned (stream() does not work)

**
Define any policy in RLS that makes use of a field contained in the JWT token:**
```
create policy "Allow only notifications to Owner's in JWT"
on notifications
for all
with check (
  (auth.jwt() ->> 'Id')::int = notifications."owner"
);
```

**At Supabase Client initialize, pass the JWT token in the header**
```
await Supabase.initialize(
    url: 'xxxxxxx',
    anonKey: 'xxxxxxxxxx',
    headers: {
      'Authorization':
          'Bearer xxxxxxxxxx',
    },
  );
```

**An example in flutter using stream**

```
late final Stream<List<Map<String, dynamic>>> _stream;

  @override
  void initState() {
    _stream = Supabase.instance.client.from('notifications').stream(['id']);
    super.initState();
  }

  @override
    return StreamBuilder(
      stream: _stream,
      builder: (context, snapshot) {
        return Container();
      },
    );
  }

```

_RLS is working normally. But the STREAM stops working!_
If you disable RLS or set a policy that does not use JWT token data, the stream method works normally

**Expected behavior**
That the stream method continues to issue changes to the table.

- Any platform
- supabase_flutter: ^1.10.9

Vinzent03:
Use authorization header to set initial access token. That's necessary, when passing custom jwt on Supabase initialization.

close #557

### DIFF
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -91,6 +91,8 @@ class RealtimeClient {
       eventsPerSecondLimitMs = (1000 / int.parse(eventsPerSecond)).floor();
     }
 
+    accessToken = this.headers['Authorization']?.split(' ').last;
+
     this.reconnectAfterMs =
         reconnectAfterMs ?? RetryTimer.createRetryFunction();
     this.encode = encode ??


dshukertjr (approved):
LGTM!

PR #564, opened 2023-07-21
fix: `maybeSingle` no longer logs error on Postgrest API

dshukertjr:

Before this update, using `maybeSingle` and getting 0 results threw errors on the server. This PR fixes it by not requesting for a single object, but requesting arrays, and on the client side see if the returned array contains 0 or 1 objects. 

Fixes https://github.com/supabase/supabase-flutter/issues/560

### DIFF
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -79,7 +79,7 @@ class PostgrestBuilder<T, S> implements Future<T> {
       isolate: _isolate,
       httpClient: _httpClient,
       options: _options,
-      // maybeSingle: _maybeSingle,
+      maybeSingle: _maybeSingle,
       converter: converter,
     );
   }
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -27,7 +27,8 @@ typedef _Nullable<T> = T?;
 class PostgrestBuilder<T, S> implements Future<T> {
   dynamic _body;
   late final Headers _headers;
-  bool _maybeEmpty = false;
+  // ignore: prefer_final_fields
+  bool _maybeSingle = false;
   String? _method;
   late final String? _schema;
   late Uri _url;
@@ -75,9 +76,7 @@ class PostgrestBuilder<T, S> implements Future<T> {
       isolate: _isolate,
       httpClient: _httpClient,
       options: _options,
-    )
-      .._maybeEmpty = _maybeEmpty
-      .._converter = converter;
+    ).._converter = converter;
   }
 
   void _assertCorrectGeneric(Type R) {
@@ -301,8 +300,8 @@ class PostgrestBuilder<T, S> implements Future<T> {
             details: response.reasonPhrase,
           );
 
-          if (_maybeEmpty) {
-            return _handleMaybeEmptyError(response, error);
+          if (_maybeSingle) {
+            return _handleMaybeSingleError(response, error);
           }
         } catch (_) {
           error = PostgrestException(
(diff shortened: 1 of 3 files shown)


packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -398,7 +415,25 @@ class PostgrestBuilder<T, S> implements Future<T> {
 
     try {
       final response = await _execute();
-      final data = response.data;
+      var data = response.data;
+
+      // Workaround for https://github.com/supabase/supabase-flutter/issues/560
```
Vinzent03, line 420: Isn't this already done in `_parseResponse()`?

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -75,9 +76,7 @@ class PostgrestBuilder<T, S> implements Future<T> {
       isolate: _isolate,
       httpClient: _httpClient,
       options: _options,
-    )
-      .._maybeEmpty = _maybeEmpty
-      .._converter = converter;
+    ).._converter = converter;
```
Vinzent03, line 79: What's the reason behind this removal?

packages/postgrest/lib/src/postgrest_transform_builder.dart
```diff
@@ -182,8 +182,14 @@ class PostgrestTransformBuilder<T> extends PostgrestBuilder<T, T> {
   ///
   /// By specifying this type via `.select<Map<String,dynamic>?>()` you get more type safety.
   PostgrestTransformBuilder<T> maybeSingle() {
-    _headers['Accept'] = 'application/vnd.pgrst.object+json';
-    _maybeEmpty = true;
+    // Temporary fix for https://github.com/supabase/supabase-flutter/issues/560
+    // Issue persists e.g. for `.insert([...]).select().maybeSingle()`
+    if (_method?.toUpperCase() == 'GET') {
+      _headers['Accept'] = 'application/json';
```
Vinzent03, line 188: What about adding a `limit=2` here as well? This would increase the performance, becauset the payload doesn't get that big.

Vinzent03 (changes requested):

packages/postgrest/lib/src/postgrest_builder.dart
dshukertjr: Yeah, I just saw a variable being assigned itself here, so I removed it. 

```dart
_maybeEmpty = _maybeEmpty
```

packages/postgrest/lib/src/postgrest_transform_builder.dart
dshukertjr: Hmm, I would rather not do anything that the user didn't ask for. I also fear that down the line there could be a case where we shoot ourselves on the foot.

packages/postgrest/lib/src/postgrest_builder.dart
dshukertjr: Oops, I meant to take this one out. Thanks for catching it!

packages/postgrest/lib/src/postgrest_builder.dart
Vinzent03: It's not assigned to itself. We construct a new `PostgrestBuilder` object here and it doesn't contain the `maybeEmpty` field in the constructor, so we are setting it afterward.

packages/postgrest/lib/src/postgrest_transform_builder.dart
dshukertjr: Using `maybeSingle` already requires developers to make sure that it only returns 0 or 1 results anyway, so it shouldn't matter. Anyway, I have edited the comment doc of the `maybeSingle` method slightly.

Vinzent03 (approved):

packages/postgrest/lib/src/postgrest_builder.dart
dshukertjr: You are right. Thanks for this one.

PR #567, opened 2023-07-24
fix(gotrue): remove OtpType deprecation and add assertion to check type for resend method

dshukertjr:

Some `OtpType` enum values were deprecated in this PR, but it seems like Gotrue server requires the deprecated values in some APIs. 

This PR removes the deprecated flag so that the users can use the resend method without deprecation warnings. 

We can add back the deprecation warnings when the backend properly starts accepting the new `email` type. 

Related https://github.com/supabase/gotrue-js/pull/642

Issue report from user https://github.com/supabase/supabase-flutter/pull/517#issuecomment-1646917472

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -528,7 +528,6 @@ class GoTrueClient {
 
   /// Resends an existing signup confirmation email, email change email, SMS OTP or phone change OTP.
   ///
-  ///
   /// For [type] of [OtpType.signup] or [OtpType.emailChange] [email] must be
   /// provided, and for [type] or [OtpType.sms] or [OtpType.phoneChange],
   /// [phone] must be provided
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -526,9 +526,13 @@ class GoTrueClient {
     );
   }
 
-  ///Resends an existing signup confirmation email, email change email, SMS OTP or phone change OTP.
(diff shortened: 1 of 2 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -526,9 +526,13 @@ class GoTrueClient {
     );
   }
 
-  ///Resends an existing signup confirmation email, email change email, SMS OTP or phone change OTP.
+  /// Resends an existing signup confirmation email, email change email, SMS OTP or phone change OTP.
   ///
-  ///Use [EmailResendParams] or [PhoneResendParams] to specify the type of resend.
+  /// Use [EmailResendParams] or [PhoneResendParams] to specify the type of resend.
```
Vinzent03, line 531:
```suggestion
```
This line can be removed as it's a wrong leftover of #517, because these classes don't exist

Vinzent03 (changes requested):

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -526,9 +526,12 @@ class GoTrueClient {
     );
   }
 
-  ///Resends an existing signup confirmation email, email change email, SMS OTP or phone change OTP.
+  /// Resends an existing signup confirmation email, email change email, SMS OTP or phone change OTP.
   ///
-  ///Use [EmailResendParams] or [PhoneResendParams] to specify the type of resend.
+  ///
```
dshukertjr, line 531:
```suggestion
```

Vinzent03 (approved):

PR #571, opened 2023-07-28
chore: Add notes about using flavors to support multiple SHA-1 fingerprint certificates

dshukertjr:

Current example oof the Google login setup assumes there is only one SHA1 fingerprint certificate, but in production there will be multiple of them, from debug certificate to the one provided by Play Store. This PR adds an instruction on how to handle such situation.

### DIFF
--- a/packages/supabase_flutter/README.md
+++ b/packages/supabase_flutter/README.md
@@ -182,6 +182,27 @@ Future<AuthResponse> signInWithGoogle() {
 }
 ```
 
+When going to production on Android, you may have multiple SHA-1 fingerprint certificates, debug and the production certificate provided from Play Store. You will need to create one Google OAuth 2.0 Client ID for each certificates. You can then utilize [flavors](https://docs.flutter.dev/deployment/flavors#using-flavors-in-android) to dynamically update the `appAuthRedirectScheme` like this:
+
+```groovy
+flavorDimensions "default"
+productFlavors {
+    dev {
+        dimension "default"
+        versionNameSuffix ".dev"
+        manifestPlaceholders += [
+            'appAuthRedirectScheme': 'com.googleusercontent.apps.*DEBUG_CLIENT_ID*'
+        ]
+    }
+    prod {
+        dimension "default"
+        manifestPlaceholders += [
+            'appAuthRedirectScheme': 'com.googleusercontent.apps.*PROD_CLIENT_ID*'
+        ]
+    }
+}
+```
+
 ### OAuth login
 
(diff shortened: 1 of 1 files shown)


romaingyh:
Hello, thanks a lot for your help (it's me who e-mailed support about it).

I've done a little research of my own, and here are a few things that might help you : 

- If I create only debug Client ID with debug fingerprint and follow the doc (before this PR), android native google auth works in debug mode but also in release mode. I can't explain why, and I can't test in production with play store fingerprint because my app is not ready to publish.

- I've found this in [AppAuth Android SDK](https://github.com/openid/AppAuth-Android) : 
>Alternatively, the redirect URI can be directly configured by adding an intent-filter for AppAuth's RedirectUriReceiverActivity to your AndroidManifest.xml:
>```gradle
><activity
>        android:name="net.openid.appauth.RedirectUriReceiverActivity"
>        tools:node="replace">
>    <intent-filter>
>        <action android:name="android.intent.action.VIEW"/>
>        <category android:name="android.intent.category.DEFAULT"/>
>        <category android:name="android.intent.category.BROWSABLE"/>
>        <data android:scheme="com.example.app"/>
>    </intent-filter>
></activity>
>```

Maybe create one `intent-filter` per ClientID without using `appAuthRedirectScheme` can work, I don't know very well Android.


- Using Flavors as this PR suggests, I think we should maybe adapt the `clientId` definition in Flutter part with something like this ?
```dart
      final isCupertino = Platform.isIOS || Platform.isMacOS;

      final clientId = switch(isCupertino) {
        true => 'IOS_CLIENT_ID',
        false when Flavor.isProduction => 'ANDROID_PROD_CLIENT_ID',
        false when Flavor.isRelease => 'ANDROID_RELEASE_CLIENT_ID',
        false when Flavor.isDevelopment => 'ANDROID_DEV_CLIENT_ID',
      };
```



I don't know if it's the right place to discuss about this, if not I've created a [discussion](https://github.com/orgs/supabase/discussions/16060) about this few days ago.

dshukertjr:
Yeah, I am having second thoughts about this PR. There are many ways a developer can add multiple Google OAuth client IDs, and I don't know if suggesting one way to the user is a good idea... At the same time though, there probably is going to be the same question asked from the users over and over.

bdlukaa:
>I< think that, if the developer is using flavors, they know what they are doing. I don't think it'll be useful to add flavors to our documentation

dshukertjr:
Yeah, I agree. It's probably not a good idea to add this to the readme. Closing this PR.

PR #580, opened 2023-08-04
feat(postgrest): adding null support for eq and neq filters

TimurKr:

Postgrest feature

When using the `eq` and `neq` filters with nullable values, exception is thrown if the value is null, because these filter by default do not work with null values.

## What is the new behavior?

If null value is used in the `eq` and `neq` filters, the `is` and `not.is` filter is used automatically, making it more convenient to filter based on nullable values without conditional filtering. 

## Additional context

-

### DIFF
--- a/packages/postgrest/lib/src/postgrest_filter_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_filter_builder.dart
@@ -64,6 +64,8 @@ class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
   PostgrestFilterBuilder<T> eq(String column, dynamic value) {
     if (value is List) {
       appendSearchParams(column, 'eq.{${_cleanFilterArray(value)}}');
+    } else if (value == null) {
+      appendSearchParams(column, 'is.null');
     } else {
       appendSearchParams(column, 'eq.$value');
     }
@@ -81,6 +83,8 @@ class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
   PostgrestFilterBuilder<T> neq(String column, dynamic value) {
     if (value is List) {
       appendSearchParams(column, 'neq.{${_cleanFilterArray(value)}}');
+    } else if (value == null) {
+      appendSearchParams(column, 'not.is.null');
     } else {
       appendSearchParams(column, 'neq.$value');
     }


Vinzent03:
I really like the idea. Though, I'm unsure if this is the expected behavior or hides issues of a variable unintentionally being null.

TimurKr:
I after this is up for a debate. If this library simply mimics the REST url (which only mimics the underlying SQL syntax of `not null` vs `= value`) that it generates under the hood, than `eq` and `neq` shouldn't work with null values either. I can't say I understand why SQL is the way it is, but for me this seems like a harmless syntactic sugar to make working with SQL a little easier. 

After all if someone is handling a possible exception, which could be thrown if the value they are passing in is null, then they might as well handle it before even calling the method, whereas if the value could be null and the database's column is billable as well, it seems like a redundant step to conditionally use the methods and thus loosing the beauty of chasing the filters in one line.

Krystian95:
> I really like the idea here as well, and I can see it benefit those that are not so familiar with SQL. But, this client library should be a simple wrapper around PostgREST, and I don't want to introduce behaviors that might confuse those that are already familiar with postgres or PostgREST. Thanks for the nice suggestion, but I will close this PR.

Hi, for example I've got a database with some columns NULL. How could I create a Supabase query with Flutter to filter the rows with a column IS NULL?

dshukertjr:
@Krystian95 
You can use the [is_()](https://supabase.com/docs/reference/dart/is) filter like this:

```dart
final data = await supabase
  .from('cities')
  .select('country_id')
  .is_('name', null);
```

Krystian95:
> @Krystian95 You can use the [is_()](https://supabase.com/docs/reference/dart/is) filter like this:
> 
> ```dart
> final data = await supabase
>   .from('cities')
>   .select('country_id')
>   .is_('name', null);
> ```

Sorry, I mean "column IS NOT NULL".

Krystian95:
> @Krystian95 You can use the [is_()](https://supabase.com/docs/reference/dart/is) filter like this:
> 
> ```dart
> final data = await supabase
>   .from('cities')
>   .select('country_id')
>   .is_('name', null);
> ```

Sorry, I mean "column IS NOT NULL".

dshukertjr:
@Krystian95 
Sorry, missed read your comment. I haven't tested, but I think this will do:

```dart
final data = await supabase
    .from('cities')
    .select()
    .not('name','is',null);
```

Krystian95:
> @Krystian95 Sorry, missed read your comment. I haven't tested, but I think this will do:
> 
> ```dart
> final data = await supabase
>     .from('cities')
>     .select()
>     .not('name','is',null);
> ```

No it was my fault, I've make a typo on my original question (now corrected).

Me too I was thinking that that's should works, but unfortunately it is not 😕

Screenshot 2023-08-18 alle 16 55 26 copia

Screenshot 2023-08-18 alle 17 02 24 (1)

dshukertjr:
@Krystian95 
Hmm, let me look into this.

Krystian95:
> @Krystian95 Sorry, missed read your comment. I haven't tested, but I think this will do:
> 
> ```dart
> final data = await supabase
>     .from('cities')
>     .select()
>     .not('name','is',null);
> ```

There was an error in the second branch of the if-else. I confirm that this you version works! Really many thanks 🙂

dshukertjr:
@Krystian95 Nice! Glad to hear it works now!

PR #591, opened 2023-08-14
Update app_links dependency to support AGP 8

### LINKED ISSUE #590: Update build.gradle to add support for Android Gradle Plugin 8
Unable to build app due to the current app_links version dependency not supporting AGP 8.
This is fixed in the latest app_links.
https://github.com/llfbandit/app_links/commit/181fdcc4ca08489f01328adb253752e50b2d2ec6

DanMossa:
Fixes #590 

Let me know if there's something I forgot to update version wise!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,28 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2023-07-13
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+- There are no breaking changes in this release.
+
+Packages with other changes:
+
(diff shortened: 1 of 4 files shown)


Vinzent03:
I wouldn't make a new release for a minor update of a package we don't even maintain. Existing users have to update `supabase_flutter` so they can just update  `app_links` instead via
`flutter pub upgrade app_links`. New users get the newest version by default. You shouldn't depend on pubspec.yaml to search for packages to update.

bdlukaa:
I agree. We can enforce the latest version, but publish a release just for that is unecessary

dshukertjr:
Thanks @DanMossa for this, but I would agree with the two here as well. I am going to close this one.

PR #592, opened 2023-08-15
feat: add `logLevel` parameter to `RealtimeClientOptions`

dshukertjr:

Adds a `logLevel` parameter, which changes the level of logs to be shown on the Supabase dashboard for realtime events. Setting `logLevel` add a query parameter to the websocket query parameter with key `log_level`. 

Currently the JS SDK doesn't have a complete implementation, but [here](https://github.com/supabase/realtime-js/blob/master/src/RealtimeClient.ts#L27) is where `log_level` parameter on js is defined.

### DIFF
--- a/packages/supabase/test/client_test.dart
+++ b/packages/supabase/test/client_test.dart
@@ -8,7 +8,7 @@ import 'utils.dart';
 
 void main() {
   group('Standard Header', () {
-    const supabaseUrl = '';
+    const supabaseUrl = 'https://nlbsnpoablmsiwndbmer.supabase.co';
     const supabaseKey = '';
     late SupabaseClient client;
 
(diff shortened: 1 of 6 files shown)


packages/supabase/test/client_test.dart
```diff
@@ -8,8 +8,9 @@ import 'utils.dart';
 
 void main() {
   group('Standard Header', () {
-    const supabaseUrl = '';
-    const supabaseKey = '';
+    const supabaseUrl = 'https://nlbsnpoablmsiwndbmer.supabase.co';
```
Vinzent03, line 11: I don't think this was intentional.

packages/supabase/test/client_test.dart
```diff
@@ -31,6 +32,29 @@ void main() {
...
+      var realtimeWebsocketURL = Uri.parse(client.realtime.endPointURL);
+      expect(
+        realtimeWebsocketURL.queryParameters,
+        containsPair('apikey', supabaseKey),
+      );
+      expect(realtimeWebsocketURL.queryParameters['log_level'], isNull);
+
+      client = SupabaseClient(supabaseUrl, supabaseKey,
+          realtimeClientOptions:
+              RealtimeClientOptions(logLevel: RealtimeLogLevel.info));
```
Vinzent03, line 46:
```suggestion
              RealtimeClientOptions(logLevel: RealtimeLogLevel.info),);
```
And reformat

Vinzent03 (changes requested):

packages/supabase/test/client_test.dart
dshukertjr: We can add [require_trailing_comma](https://dart.dev/tools/linter-rules/require_trailing_commas) lint rule if we want to enforce these. Otherwise, I don't think we need to be so sensitive about code format as long as it's passing the lints.

packages/supabase/test/client_test.dart
dshukertjr: They are actually intentional. I wanted to make sure the URL transformation is properly being done, and an empty string isn't really a realistic test value.

Vinzent03 (approved):

packages/supabase/test/client_test.dart
dshukertjr: But yeah, I am all in for it to add `require_trailing_comma` to our lint. We can probably introduce a few more rules to improve our code base.

PR #596, opened 2023-08-22
fix(realtime_client, supabase): pass apikey as the initial access token for realtime client

dshukertjr:

Realtime server expects the anon key to be sent with the heart beat for unauthenticated users, but currently `null` is being sent. This PR sets the anon key as the initial value of `accessToken` for `RealtimeClient` so that unauthenticated client will send the anon key to realtime server .

Related https://github.com/supabase/realtime-js/pull/245

### DIFF
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -104,7 +104,8 @@ class RealtimeClient {
       eventsPerSecondLimitMs = (1000 / int.parse(eventsPerSecond)).floor();
     }
 
-    accessToken = this.headers['Authorization']?.split(' ').last;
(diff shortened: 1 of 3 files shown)


packages/supabase_flutter/lib/src/supabase_auth.dart
```diff
@@ -159,6 +159,7 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
+    // ignore:missing_enum_constant_in_switch
```
dshukertjr, line 162: There was a new `AppLifecycleState.hidden` enum value added in the latest stable version of Flutter. This line was added to ignore the check to add all enum values to the switch.

Vinzent03 (approved):

PR #598, opened 2023-08-22
refactor!: use Dart 3

Vinzent03:
This is the first pr to begin working on v2 of our packages.

Where do we want to merge this into? Do we create a new v2 branch and collect all prs there and keep main for v1 or do we do it the other way around?
I would keep the possibility to do fixes to v1 while working on v2.

# Changes
- Make Dart 3 minimum version
- upgrade lints packages
  - fix new linter warnings
- upgrade mocktail in realtime_client
- upgrade dotenv in gotrue

### DIFF
--- a/.github/workflows/supabase_flutter_ci.yml
+++ b/.github/workflows/supabase_flutter_ci.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        flutter-version: ['3.0.0', '3.x']
+        flutter-version: ['3.10.x', '3.13.x']
 
     defaults:
       run:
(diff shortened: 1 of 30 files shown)


dshukertjr:
This is nice! Let's actually create a new `next` branch, and merge all the changes for v2 there.

.github/workflows/supabase_flutter_ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        flutter-version: ['3.0.0', '3.x']
+        flutter-version: ['3.10.x', '3.13.x']
```
dshukertjr, line 35: Should we keep the `3.x` to test it against the latest Flutter version?

Vinzent03: Hm yeah. I've replaced 3.13.x now with 3.x, because they are currently the same.

dshukertjr (approved):
LGTM. The CI failing is a bit annoying, but it's the dev branch so I'm not too worried.

Vinzent03:
Yeah that's really strange, I'm trying to reproduce it, but the tests pass with Dart dev and beta.

PR #603, opened 2023-08-23
Fix: double http package

lebe24:

Remove duplicate HTTP package

### DIFF
--- a/packages/storage_client/lib/src/fetch.dart
+++ b/packages/storage_client/lib/src/fetch.dart
@@ -3,7 +3,6 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'package:http/http.dart' as http;
-import 'package:http/http.dart';
 import 'package:http_parser/http_parser.dart' show MediaType;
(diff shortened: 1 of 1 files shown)


dshukertjr (changes requested):
Thanks for this PR! It seems like we want to remove the other import and adjust the code in `fetch.dart` file?

lebe24:
Hi @dshukertjr, Yes there is a duplicate HTTP package In the ``fetch.dart`` file the delete one is not been used

dshukertjr:
@lebe24 Failing tests suggest otherwise.

lebe24:
@dshukertjr, When I ran it locally on my machine without the duplicate package all functions worked as expected but will have to write a local test to see if the package is needed else will close the pull request.

Vinzent03:
It's not about writing a test, because this a compilation time error. As seen in the ci, the analyzer shows the error and not a run test.

lebe24:
Okay, everything is all good, I will close the request. Thank you

Issue #541, opened 2023-07-09
Allow for handling non-supabase related deep link /applink urls


RMcGhee:
**Is your feature request related to a problem? Please describe.**
Currently, the way that supabase handles app links, it automatically figures out if an incoming link is related to an auth changes, and updates the auth state accordingly, which is great. However, I have a need to also handle other kinds of links in my app, and doing so with uni_links or uni_links_desktop breaks the supabase url handling flow, where uni_links_desktop will catch the url instead of supabase, and auth state is no longer updated. I'd like to be able to handle the following:
supabase auth links, like signing in via a magic link email
supabase reset-password link: com.mycompany.myapp://reset-password < supabase currently does a reset password workflow, but when this link is clicked from a user's email, Supabase-flutter signs the user in, and doesn't expose the url to the app, as far as I can tell.
custom url scheme: com.mycompany.myapp://share/some-share-id

**Describe the solution you'd like**
If supabase could expose a stream to the uriLinkStream, or expose a stream that allows us to handle links that supabase doesn't do anything with, that would be ideal. AFAICT, in supabase_auth._handleDeepLink(), after checking _isAuthCallbackDeeplink(uri) if the link isn't a supabase auth link, it simply returns without doing anything. If instead, it could emit the url to another stream, or just emit all urls received to a stream (to handle the reset-password case mentioned above), that would be great.

**Describe alternatives you've considered**
I initially started trying to handle the deepLinks myself, and then using Supabase.instance.client.auth.getSessionFromUrl(uri), but linking on desktop vs mobile is different enough that I encountered unexpected problems. Primarily, on macOS, you can preempt the deep link handler that supabase uses with uni_links_desktop, but on android/iOS, I wasn't able to get preemption working. I'm sure I could get this working eventually, but I feel like the solution would be fairly fragile, since uni_links_desktop is beta, and I'm not sure if preemption would be stable on different versions of the same OS.

**Additional context**
Love y'alls product, it's been a blast working with supabase!

Vinzent03:
So you listened to the `linkStream`, but it didn't emit any events? Did the initial link method return `null` as well?

RMcGhee:
Correct, if I set up my own instance of AppLinks, and listen to the uriLinkStream, no events are received. As well, getInitialAppLink returns null, but that's expected since these are links that come in when the app has been running already.
Supabase still handles urls properly in this case.

Vinzent03:
[...]

PR #605, opened 2023-08-25
Expose a deeplinks stream controller to allow non-supabase related deeplinks handling

romaingyh:

This PR adds a `Stream<Uri>`  to `SupabaseAuth.dart`, it allows to listen for deeplinks not handled by supabase auth.

Currently we can't listen for deeplinks using plugins like `AppLinks` because `supabase_flutter` also uses it for auth (email validation, password reset...) and after experimentations, when there are two (or more) subscriptions to `AppLinks`'s uri stream, only one (the last or first registered I don't remember exactly) can receive events.

I think it's the cause of #541 and this PR should solve it.

## What is the new behavior?

The added Stream property make a bridge between `AppLinks` and user's app. We can use 
```dart
SupabaseAuth.instance.uriLinksStream.listen((event) {
    debugPrint('New uri link: $event');
});
```
in ours apps to listen for deeplinks _not handled by supabase_.

## Additional context

I'm not sure if `SupabaseAuth.dart` is the good place to add this code, I hesitated with SupabaseClient or GoTrueClient, let me know.
Also with this implementation, only the links not handled by supabase are added to the the Stream.

### DIFF
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -56,6 +56,12 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   final _appLinks = AppLinks();
 
+  /// A [StreamController] that emits the deep link when it's received and not handled by the library.
+  final _onUriLinkController = StreamController<Uri>.broadcast();
+
+  /// A [Stream] that emits the deep link when it's received and not handled by the library.
+  Stream get uriLinksStream => _onUriLinkController.stream;
+
   /// A [SupabaseAuth] instance.
   ///
   /// If not initialized, an [AssertionError] is thrown
@@ -232,6 +238,7 @@ class SupabaseAuth with WidgetsBindingObserver {
   void _stopDeeplinkObserver() {
     Supabase.instance.log('***** SupabaseDeepLinkingMixin stopAuthObserver');
     _deeplinkSubscription?.cancel();
+    _onUriLinkController.close();
   }
 
   /// Handle incoming links - the ones that the app will receive from the OS
@@ -281,7 +288,11 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   /// Callback when deeplink receiving succeeds
   Future<void> _handleDeeplink(Uri uri) async {
-    if (!_instance._isAuthCallbackDeeplink(uri)) return;
+    if (!_instance._isAuthCallbackDeeplink(uri)) {
+      // notify listener that a deeplink was received and not handled
+      _onUriLinkController.add(uri);
+      return;
+    }
 
     Supabase.instance.log('***** SupabaseAuthState handleDeeplink $uri');
 


Vinzent03:
On which platform did you have that issue? As said [here](https://github.com/supabase/supabase-flutter/issues/541), I couldn't reproduce that issue.

romaingyh:
> On which platform did you have that issue? As said [here](https://github.com/supabase/supabase-flutter/issues/541), I couldn't reproduce that issue. And make sure it's after `runApp` so best to use the api from App Links at the same place where you would listen to the supabase exposed api in your code.

Hello, sorry for late reply. Here is a code to reproduce the issue. Now I can confirm only the last added listener to `Applinks` will receive events.

```dart
import 'package:app_links/app_links.dart';

void main() async {

  await Supabase.initialize(
    url: const String.fromEnvironment('SUPABASE_URL'),
    anonKey: const String.fromEnvironment('SUPABASE_ANON_KEY'),
    authCallbackUrlHostname: 'auth',
  );

  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return MaterialApp(
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: const MyHomePage(title: 'Deep links PR'),
    );
  }
}

class MyHomePage extends StatelessWidget {
  const MyHomePage({super.key, required this.title});

  final String title;

  @override
    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: Text(title),
      ),
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            const Text('Last deeplink opened :'),
            StreamBuilder(
              stream: AppLinks().allUriLinkStream,
              builder: (context, uri) {
                return Text(
                  '${uri.data}',
                  style: Theme.of(context).textTheme.headlineMedium,
                );
              },
            ),
          ],
        ),
      ),
    );
  }
}
```
Running on ios simulator with `flutter run --dart-define-from-file=.env.json`. I created a fresh supabase project for this test, url and supabase key are in `.env.json` file. I set pr.deeplinks://auth as redirect url in dashboard and also configured the corresponding scheme/host in Info.plist and AndroidManifest.

If you run code like this and open :
`pr.deeplinks://auth`
`pr.deeplinks://other`

Both will works and will be displayed on screen, they've been handled by the stream listener but not Supabase.

You can comment lines 49 to 57 to remove listener, restart app and now re-open `pr.deeplinks://auth`. There is an error in console that confirms Supabase was not handling the deeplink before with the StreamBuilder.

I can provide my repo if you can't reproduce

Vinzent03:
I'm pretty sure I already tested this on android and/or linux device and couldn't reproduce. But I wouldn't wonder if that's an Apple only issue. I don't have a mac device, so can @dshukertjr verify this? 

Instead of fixing this in supabase-flutter, I would instead fix that in `app_links` so that they create only one stream and broadcast that.

gadfly361 (approved):

gadfly361:
My apologies, didn't mean to approve, accidental click.

dshukertjr:
Closing in favor of https://github.com/supabase/supabase-flutter/pull/848

proformance:
I am also struggling with this; I cannot use AppLink in my app for other uses due to supabase-flutter already reading the stream. I initially came up with a very similar solution before I noticed this PR and https://github.com/supabase/supabase-flutter/issues/541.

I don't think #848 solves the issue, reason being that disabling it only allows me to use one or the other.


I do however I fully agree with 

> ...
> Instead of fixing this in supabase-flutter, I would instead fix that in `app_links` so that they create only one stream and broadcast that.

I tried to solve it here, seems to work for me now, let's hope it gets merged
https://github.com/llfbandit/app_links/pull/145

proformance:
It's now available in [6.2.0](https://pub.dev/packages/app_links/changelog) 🎉

Issue #453, opened 2023-04-26
Flutter Web bugs and compatibility with native web (cookies vs indexed DB)


tomekit:
The https://pub.dev/packages/supabase_flutter package has Web support listed on pub.dev, I was wondering what's the state of this package and is it production ready. I am using latest version `1.7.0`, but was using `1.6.2` for quite a while and same issues appear.

1. First off, the session itself is kept in different storage comparing to native Web, meaning that if some parts of your sites are Flutter and some parts are native Web... they won't understand each other.
Next.js web, cookie: supabase-auth-token
Flutter web, indexed DB: supabase_authentication

Are there any plans to unify these, so it's possible to combine native Web with Flutter Web?

2. Sometimes when revisiting Flutter Web user often gets logged out with: "Invalid Refresh Token: Refresh Token Not Found".
Refreshing webpage again doesn't render this error, but user refresh token is lot.

3. Sometimes when revisiting Flutter Web user often gets logged out with:
```
- flutter: ***** Supabase init completed Instance of 'Supabase'
- flutter: Recovering corrupted box.
```

4. When revisiting webpage my app often gets stuck on (seem to happen only during release mode).
```
Got object store box in database supabase_authentication. main.dart.js:44443:78
Uncaught Error: Invalid argument(s): Invalid or corrupted pad block
```

I don't have enough time to exactly investigate these issue, they seem to happen at random making Flutter Web unusable to me. 
I've though it's still better to report these under somewhat incomplete form, so at least Supabase team is aware of these.

Thanks !

PR #608, opened 2023-08-26
feat(supabase_flutter): use SharedPreferences for access token

Vinzent03:
# Changes
Switches from [Hive](https://pub.dev/packages/hive) to [Shared Preferences](https://pub.dev/packages/shared_preferences) for storing the access token

This is done by still including the Hive package:
1. reading the access token
2. writing it to shared prefs
3. removing it from Hive
4. Removing the file and the parent directory

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -773,29 +773,14 @@ class GoTrueClient {
     );
   }
 
-  /// Recover session from persisted session json string.
-  /// Persisted session json has the format { currentSession, expiresAt }
-  ///
-  /// currentSession: session json object, expiresAt: timestamp in seconds
+  /// Recover session from stringified [Session].
   Future<AuthResponse> recoverSession(String jsonStr) async {
-    final persistedData = json.decode(jsonStr) as Map<String, dynamic>;
-    final currentSession =
-        persistedData['currentSession'] as Map<String, dynamic>?;
-    final expiresAt = persistedData['expiresAt'] as int?;
-    if (currentSession == null) {
-      throw _notifyException(AuthException('Missing currentSession.'));
-    }
-    if (expiresAt == null) {
-      throw _notifyException(AuthException('Missing expiresAt.'));
-    }
-
-    final session = Session.fromJson(currentSession);
+    final session = Session.fromJson(json.decode(jsonStr));
     if (session == null) {
       throw _notifyException(AuthException('Current session is missing data.'));
     }
 
-    final timeNow = (DateTime.now().millisecondsSinceEpoch / 1000).round();
-    if (expiresAt < (timeNow + Constants.expiryMargin.inSeconds)) {
+    if (session.isExpired) {
       if (_autoRefreshToken && session.refreshToken != null) {
         return await _callRefreshToken(
           refreshToken: session.refreshToken,
--- a/packages/supabase_flutter/lib/src/local_storage.dart
+++ b/packages/supabase_flutter/lib/src/local_storage.dart
@@ -1,6 +1,8 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:hive_flutter/hive_flutter.dart';
 import 'package:shared_preferences/shared_preferences.dart';
@@ -16,63 +18,54 @@ const supabasePersistSessionKey = 'SUPABASE_PERSIST_SESSION_KEY';
 ///   * [SupabaseAuth], the instance used to manage authentication
 ///   * [EmptyLocalStorage], used to disable session persistence
 ///   * [HiveLocalStorage], that implements Hive as storage method
+///   * [SharedPreferencesLocalStorage], that implements SharedPreferences as storage method
+///   * [MigrationLocalStorage], to migrate from Hive to SharedPreferences
 abstract class LocalStorage {
-  const LocalStorage({
-    required this.initialize,
-    required this.hasAccessToken,
-    required this.accessToken,
-    required this.persistSession,
-    required this.removePersistedSession,
-  });
+  const LocalStorage();
 
   /// Initialize the storage to persist session.
-  final Future<void> Function() initialize;
+  Future<void> initialize();
 
   /// Check if there is a persisted session.
-  final Future<bool> Function() hasAccessToken;
+  Future<bool> hasAccessToken();
 
   /// Get the access token from the current persisted session.
-  final Future<String?> Function() accessToken;
+  Future<String?> accessToken();
 
   /// Remove the current persisted session.
-  final Future<void> Function() removePersistedSession;
+  Future<void> removePersistedSession();
 
   /// Persist a session in the device.
-  final Future<void> Function(String) persistSession;
+  Future<void> persistSession(String persistSessionString);
 }
 
 /// A [LocalStorage] implementation that does nothing. Use this to
 /// disable persistence.
 class EmptyLocalStorage extends LocalStorage {
   /// Creates a [LocalStorage] instance that disables persistence
-  const EmptyLocalStorage()
-      : super(
-          initialize: _initialize,
-          hasAccessToken: _hasAccessToken,
-          accessToken: _accessToken,
-          removePersistedSession: _removePersistedSession,
-          persistSession: _persistSession,
-        );
+  const EmptyLocalStorage();
 
-  static Future<void> _initialize() async {}
-  static Future<bool> _hasAccessToken() => Future.value(false);
-  static Future<String?> _accessToken() => Future.value();
-  static Future<void> _removePersistedSession() async {}
-  static Future<void> _persistSession(_) async {}
+  @override
+  Future<void> initialize() async {}
+
+  @override
+  Future<bool> hasAccessToken() => Future.value(false);
+
+  @override
+  Future<String?> accessToken() => Future.value();
+
+  @override
+  Future<void> removePersistedSession() async {}
+
+  @override
+  Future<void> persistSession(persistSessionString) async {}
 }
 
 /// A [LocalStorage] implementation that implements Hive as the
 /// storage method.
 class HiveLocalStorage extends LocalStorage {
   /// Creates a LocalStorage instance that implements the Hive Database
-  const HiveLocalStorage()
-      : super(
-          initialize: _initialize,
-          hasAccessToken: _hasAccessToken,
-          accessToken: _accessToken,
-          removePersistedSession: _removePersistedSession,
-          persistSession: _persistSession,
-        );
+  const HiveLocalStorage();
 
   /// The encryption key used by Hive. If null, the box is not encrypted
   ///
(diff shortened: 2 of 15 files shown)


packages/supabase_flutter/lib/src/local_storage.dart
```diff
@@ -16,63 +18,54 @@ const supabasePersistSessionKey = 'SUPABASE_PERSIST_SESSION_KEY';
...
+  Future<bool> hasAccessToken() => Future.value(false);
+
+  @override
+  Future<String?> accessToken() => Future.value();
+
+  @override
+  Future<void> removePersistedSession() async {}
+
+  @override
+  Future<void> persistSession(persistSessionString) async {}
```
dshukertjr, line 61: Love this cleanup!

packages/supabase_flutter/lib/src/local_storage.dart
```diff
@@ -93,31 +87,129 @@ class HiveLocalStorage extends LocalStorage {
...
+
+  @override
+  Future<void> initialize() async {
+    WidgetsFlutterBinding.ensureInitialized();
+    _prefs = await SharedPreferences.getInstance();
+  }
+
+  @override
+  Future<bool> hasAccessToken() {
+    return Future.value(_prefs.containsKey(supabasePersistSessionKey));
```
dshukertjr, line 132: Could we maybe use the [same storage key as the js client library](https://github.com/supabase/gotrue-js/blob/master/src/lib/constants.ts#L3)?
Related https://github.com/supabase/supabase-flutter/issues/453

Vinzent03: I still don't think this enables both native web and flutter to work together. Shared Prefs stores them in localstorage and native web uses cookies (Though I'm not exactly sure how they work).

dshukertjr: @Vinzent03 The JS sdk stores them in a local storage, and there are [auth-helpers](https://supabase.com/docs/guides/auth/auth-helpers) that takes the session and puts them in cookies. No need to worry about the cookie part, but no sense in having a different storage keys compared to the JS library. Let's match it.

Vinzent03:
SharedPreferences [uses a prefix](https://pub.dev/packages/shared_preferences#migration-and-prefixes), so it's not possible to use the package to use the js localstorage entry as well. I've now accessed the `localStorage` directly to use the same entry. Is a bit more complicated, because you can't import `dart:html` on non web platform.

dshukertjr:
@Vinzent03 
>Setting the prefix to an empty string '' will allow access to all preferences created by any non-flutter versions of the app (for migrating from a native app to flutter).

Can we do this and use the same storage key as the JS SDK?

Vinzent03:
No, because that's not per instance, but for the global Shared preference and would therefore conflict with the other users use of this.

dshukertjr:
It seems like there are discussions around exposing the underlying storage for shared preferences. If this lands, we might be able to access the session object stored by supabase-js from this library. 
https://github.com/flutter/flutter/issues/123078

It's not going to make it compatible with supabase-js just yet because of the `flutter.` prefix and shared_preference's lack of object support, but let's modify our new storage key to match what the js client library has instead of our current `SUPABASE_PERSIST_SESSION_KEY` just of parity then.
https://github.com/supabase/supabase-js/blob/master/src/SupabaseClient.ts#L99

Vinzent03:
Ah supabase-js uses a different storage key, okay. Will fix that. I don't get your first point, though. I've [already implemented](https://github.com/supabase/supabase-flutter/pull/608/files#diff-11b3cc75e31d9e751704e1c81334d73db33bd797fbc6ad1c59930b009f14da10) the access to the `window.localStorage`, which works fine. But I see now that the [session object in js](https://github.com/supabase/gotrue-js/blob/22923e7ea4780287e9a6b4c596889c51a00b9039/src/lib/types.ts#L179) is differently structured than what [we store in Dart](https://github.com/supabase/supabase-flutter/blob/36c8a58bf8b9ca32c00dca8ce1efb62de4f18d94/packages/gotrue/lib/src/types/session.dart#L81)

So do we want both to work together or not?

dshukertjr:
@Vinzent03 
> I don't get your first point, though. I've [already implemented](https://github.com/supabase/supabase-flutter/pull/608/files#diff-11b3cc75e31d9e751704e1c81334d73db33bd797fbc6ad1c59930b009f14da10) the access to the window.localStorage

I didn't realize that for web, you are access the local storage directly, instead of accessing it though shared preferences. That's great! I didn't have to worry about all the prefix issues and stuff that I had mentioned previously 😂

You might be aware of it, but the JS SDK doesn't stringify the session object when storing it in the local storage, so that is something we have to take into account when storing and retrieving session on the web with local storage.

>But I see now that the [session object in js](https://github.com/supabase/gotrue-js/blob/22923e7ea4780287e9a6b4c596889c51a00b9039/src/lib/types.ts#L179) is differently structured than what [we store in Dart](https://github.com/supabase/supabase-flutter/blob/36c8a58bf8b9ca32c00dca8ce1efb62de4f18d94/packages/gotrue/lib/src/types/session.dart#L81)

Yeah, didn't think about this one, but good catch! Let's do it!

packages/supabase_flutter/lib/src/local_storage_web.dart
```diff
@@ -1,19 +1,17 @@
...
 
-Future<void> removePersistedSession() async =>
-    _localStorage.remove(SharedPreferencesLocalStorage.persistSessionKey);
+Future<void> removePersistedSession(String persistSessionKey) async =>
+    _localStorage.remove(persistSessionKey);
 
-Future<void> persistSession(persistSessionString) async =>
-    _localStorage[SharedPreferencesLocalStorage.persistSessionKey] =
-        persistSessionString;
+Future<void> persistSession(
```
dshukertjr, line 15: On JS client, the session is stored as an object rather than a String. Could we mimic the behavior here for Flutter web?

https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L1765

dshukertjr (commented):
This is great. Just one final check, and I think this PR is good to be merged!

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -773,29 +773,14 @@ class GoTrueClient {
...
-    final expiresAt = persistedData['expiresAt'] as int?;
-    if (currentSession == null) {
-      throw _notifyException(AuthException('Missing currentSession.'));
-    }
-    if (expiresAt == null) {
-      throw _notifyException(AuthException('Missing expiresAt.'));
-    }
-
-    final session = Session.fromJson(currentSession);
+    final session = Session.fromJson(json.decode(jsonStr));
```
bdlukaa, line 778: Is the `jsonStr` long? We may take advantage of the `compute` function if it's a complicated parsing

packages/supabase_flutter/lib/src/local_storage_web.dart
dshukertjr: Ah, didn't realize that. The chrome debug tool made it look like it was an object, so I just assumed it was stored as an object! Great catch!

dshukertjr: I tried it locally with an Next.js web app and Flutter web app, and the session is persisted 🎉 Thanks for the amazing work!

packages/gotrue/lib/src/gotrue_client.dart
Vinzent03: When auto formatted, it looks like the following. I don't think it's necessary to spin up an isolate for this.
```
{
    "access_token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2OTM5MDk4MzQsInN1YiI6IjE4YmM3YTRlLWMwOTUtNDU3My05M2RjLWUwYmUyOWJhZGE5NyIsImVtYWlsIjoiZmFrZTFAZW1haWwuY29tIiwicGhvbmUiOiIxNjY2MDAwMDAwMDAiLCJhcHBfbWV0YWRhdGEiOnsicHJvdmlkZXIiOiJlbWFpbCIsInByb3ZpZGVycyI6WyJlbWFpbCJdfSwidXNlcl9tZXRhZGF0YSI6e30sInJvbGUiOiIiLCJhYWwiOiJhYWwxIiwiYW1yIjpbeyJtZXRob2QiOiJwYXNzd29yZCIsInRpbWVzdGFtcCI6MTY5MzkwNjIzNH1dLCJzZXNzaW9uX2lkIjoiZDdhZDdhMzgtY2YzOS00ZGI3LTllNDYtMTlmNDc2ZjViNDBmIn0.ZtPq2EYKIvH8fyjJ7sbdtLPjvzS1DQFO4a1UHkojrkU",
    "expires_in": 3600,
    "refresh_token": "a-x22GFhn19riGhRm52kZg",
    "token_type": "bearer",
    "provider_token": null,
    "provider_refresh_token": null,
    "user": {
        "id": "18bc7a4e-c095-4573-93dc-e0be29bada97",
        "app_metadata": {
            "provider": "email",
            "providers": [
                "email"
            ]
        },
        "user_metadata": {},
        "aud": "",
        "confirmation_sent_at": null,
        "recovery_sent_at": null,
        "email_change_sent_at": null,
        "new_email": null,
        "invited_at": null,
        "action_link": null,
        "email": "fake1@email.com",
        "phone": "166600000000",
        "created_at": "2023-09-05T09:30:34.682647Z",
        "confirmed_at": "2023-09-05T09:30:34.682647Z",
        "email_confirmed_at": "2023-09-05T09:30:34.682647Z",
        "phone_confirmed_at": "2023-09-05T09:30:34.682647Z",
        "last_sign_in_at": "2023-09-05T09:30:34.860853363Z",
        "role": "",
        "updated_at": "2023-09-05T09:30:34.863812Z",
        "identities": [
            {
                "id": "18bc7a4e-c095-4573-93dc-e0be29bada97",
                "user_id": "18bc7a4e-c095-4573-93dc-e0be29bada97",
                "identity_data": {
                    "email": "fake1@email.com",
                    "sub": "18bc7a4e-c095-4573-93dc-e0be29bada97"
                },
                "provider": "email",
                "created_at": "2023-09-05T09:30:34.682647Z",
                "last_sign_in_at": "2023-09-05T09:30:34.682647Z",
                "updated_at": "2023-09-05T09:30:34.682647Z"
            }
        ],
        "factors": [
            {
                "id": "1d3aa138-da96-4aea-8217-af07daa6b82d",
                "friendly_name": "UnverifiedFactor",
                "factor_type": "totp",
                "status": "unverified",
                "created_at": "2023-09-05T09:30:34.682647Z",
                "updated_at": "2023-09-05T09:30:34.682647Z"
            }
        ]
    }
}
```

dshukertjr (approved):
Looks awesome! Making supabase_flutter compatible with the JS client is huge 💪

tomekit:
Just my two cents. Have you guys considered using: https://pub.dev/packages/flutter_secure_storage instead of SharedPreferences?

Vinzent03:
It's not really necessary, as the access token is accessible on the web via localstorage as well. And it's easier to maintain as it's supported on all platforms and means a dependency less. If you want to, you can still use it by providing your own `LocalStorage` implementation.

tomekit:
I've assumed that SharedPreferences replaces Hive on all platforms. For Web SharedPreferences is more than fine, browsers have built-in security, so no other website can steal the token (on desktop that's not the case).

tomekit:
To follow up on my previous comment, what I really meant is where the session data ends on the user's device and whether it is actually secure to use SharedPreferences (that is whether other apps could potentially access Supabase session). It's still way better than a previously used Hive.

Please find below comparison tables:
### flutter_secure_storage
```
Platform | Location
Android | EncryptedSharedPreferences
iOS/macOS | Keychain
Linux | libsecret
Web | ?
Windows | ?
```

### shared_preferences
```
Platform | Location
Android | SharedPreferences
iOS/macOS | NSUserDefaults
Linux | In the XDG_DATA_HOME directory
Web | LocalStorage
Windows | In the roaming AppData directory
```

PR #613, opened 2023-09-01
feat(gotrue): add WhatsApp support for OTP login

dshukertjr:

Adds WhatsApp support for OTP login

JS implementation are the following:
- https://github.com/supabase/gotrue-js/pull/616
- https://github.com/supabase/gotrue-js/pull/647

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -194,7 +194,7 @@ class GoTrueClient {
         'password': password,
         'data': data,
         'gotrue_meta_security': {'captcha_token': captchaToken},
-        'channel': (channel ?? OtpChannel.sms).name,
+        'channel': channel.name,
       };
       final fetchOptions = GotrueRequestOptions(headers: _headers, body: body);
       response = await _fetch.request('$_url/signup', RequestMethodType.post,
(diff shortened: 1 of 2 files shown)


Vinzent03:
Isn't the implementation for `signUp()` missing? And what about `resend()`? js doesn't have it implemented, but is it because the server doesn't support it, or they forgot it?

dshukertjr:
>Isn't the implementation for signUp() missing? 

Done.

>js doesn't have it implemented, but is it because the server doesn't support it, or they forgot it?

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -141,13 +141,16 @@ class GoTrueClient {
...
+  ///
+  /// [channel] Messaging channel to use (e.g. whatsapp or sms)
   Future<AuthResponse> signUp({
     String? email,
     String? phone,
     required String password,
     String? emailRedirectTo,
     Map<String, dynamic>? data,
     String? captchaToken,
+    OtpChannel? channel,
```
bdlukaa, line 153: I like to make things non-nullable as possible.

```suggestion
    OtpChannel channel = OtpChannel.sms,
```

This way it makes it clear for the dev the default value when hovered in the IDE, for example. This could be mentioned in the inline docs as well

Vinzent03 (approved):

PR #616, opened 2023-09-04
fix(functions_client): use header for response parsing

### LINKED ISSUE #614: FormatException: Unexpected character (at character 1)
**Describe the bug**
If the Edge Function completes its execution due to a timeout, we receive a JSON parsing error. Although, it would be more logical to receive a timeout error

**To Reproduce**
1) Write a function that will run for a very long time.
2) Call this function in a Flutter application.
3) Expect an error.

**Expected behavior**
It would be correct to receive a timeout error

**Screenshots**
Screenshot 2023-09-04 at 11 10 36
Screenshot 2023-09-04 at 11 10 50

├── supabase_flutter 1.10.14
│   ├── supabase 1.11.1
│   │   ├── functions_client 1.3.2
│   │   ├── gotrue 1.11.2
│   │   ├── postgrest 1.5.0
│   │   ├── realtime_client 1.2.1
│   │   ├── storage_client 1.5.1

Vinzent03:
# Changes
Switches to depend on the response headers on how to parse the response data

That's the way how it's done in `functions-js`  as well. I'm thinking about switching `FunctionResponse.data` to `Object`. That way the user has to verify the response is of their desired type, but I'm unsure this is appreciated.

Close #614

### DIFF
--- a/packages/functions_client/lib/src/types.dart
+++ b/packages/functions_client/lib/src/types.dart
@@ -1,28 +1,9 @@
-enum ResponseType {
-  json,
-  text,
-  arraybuffer,
-  blob,
-}
-
 enum HttpMethod {
   get,
   post,
   put,
   delete,
   patch,
-}
-
-class FunctionInvokeOptions {
-  final Map<String, String>? headers;
-  final Object? body;
-  final ResponseType? responseType;
-
-  FunctionInvokeOptions({
-    this.headers,
-    this.body,
-    this.responseType,
-  });
 }
 
 class FunctionResponse {
--- a/packages/functions_client/lib/src/types.dart
+++ b/packages/functions_client/lib/src/types.dart
@@ -1,3 +1,6 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
 enum HttpMethod {
   get,
   post,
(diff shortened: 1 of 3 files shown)


packages/functions_client/lib/src/types.dart
```diff
@@ -13,19 +9,11 @@ enum HttpMethod {
   patch,
 }
 
-class FunctionInvokeOptions {
```
Vinzent03, line 16: Somehow this isn't even used anymore.

dshukertjr:
Could I ask how this PR fixes https://github.com/supabase/supabase-flutter/issues/614 ?

dshukertjr:
> That way the user has to verify the response is of their desired type, but I'm unsure this is appreciated.

I prefer to keep it dynamic. Most of the case the returned type should be `Map`, and Object is annoying, because it doesn't let you just access values using `[]` directly.

Vinzent03:
The `FormatException` occurs, because you locally specify that the format needs to be a json type, but the server actually sends a text, which can't be parsed to json. So because it only depends on the server how the format actually looks like, I don't think you should be able to specify locally how the result has to look like. And the current functions api is similar to the postgrest api issue, where a method parameter changes the method's return type, which is not common in Dart and I think should be avoided.
But I see now that for the case of a timeout, we should maybe throw errors for non 200 responses. This would also solve this specific issue.

dshukertjr (approved):

PR #617, opened 2023-09-05
fix(gotrue): export everything in constants.dart and hide what we want to hide instead of using show

dshukertjr:

Currently we use the `show` keyword in exporting symbols from constants.dart file in gotrue. This requires us to add the new symbol to the export every time we create a new one. This PR changes to using `hide` to hide what we don't want to export so that new symbols are exported by default instead.


dshukertjr:
I'm going to do another release just because I want to get the whatsapp otp sign in out.

Issue #579, opened 2023-08-03
Stream errors after system sleep (Desktop)


maxfornacon:
**Describe the bug**
I'm experiencing an issue on Flutter desktop especially on MacOS (other platforms not properly tested yet). When the OS goes into sleep mode and awakens after a longer period of time streams seem to not be reconnecting properly.

I'm getting a couple of error messages:

First a couple of these:
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: 
#0      SupabaseStreamBuilder._addException (package:supabase/src/supabase_stream_builder.dart:442:67)
#1      SupabaseStreamBuilder._getStreamData.<anonymous closure> (package:supabase/src/supabase_stream_builder.dart:348:9)
#2      RealtimeChannel.subscribe.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:202:39)
#3      RealtimeChannel.onError.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:331:36)
#4      RealtimeChannel.trigger (package:realtime_client/src/realtime_channel.dart:569:22)
#5      RealtimeClient._triggerChanError (package:realtime_client/src/realtime_client.dart:408:15)
#6      RealtimeClient._onConnClose (package:realtime_client/src/realtime_client.dart:389:7)
#7      RealtimeClient.connect.<anonymous closure> (package:realtime_client/src/realtime_client.dart:147:11)
#8      _RootZone.runGuarded (dart:async/zone.dart:1581:10)
#9      _BufferingStreamSubscription._sendDone.sendDone (dart:async/stream_impl.dart:392:13)
#10     _BufferingStreamSubscription._sendDone (dart:async/stream_impl.dart:402:7)
#11     _BufferingStreamSubscription._close (dart:async/stream_impl.dart:291:7)
#12     _ForwardingStream._handleDone (dart:async/stream_pipe.dart:99:10)
#13     _ForwardingStreamSubscription._handleDone (dart:async/stream_pipe.dart:161:13)
#14     _RootZone.runGuarded (dart:async/zone.dart:1581:10)
#15     _BufferingStreamSubscription._sendDone.sendDone (dart:async/stream_impl.dart:392:13)
#16     _BufferingStreamSubscription._sendDone (dart:async/stream_impl.dart:402:7)
#17     _BufferingStreamSubscription._close (dart:async/stream_impl.dart:291:7)
[...]

PR #620, opened 2023-09-06
fix(realtime_client): No exception is thrown when connection is closed. 

dshukertjr:

Currently, the realtime_client will contain an exception if the user has a call back registered within `.subscribe` like this:

```dart
supabase
  .channel('channel')
  .on(...)
  .subscribe((status, [error]) {
    // `error` will be non-null when the connection is closed for example when the app goes to background mode etc.
  });
```

This PR fixes it. 

Fixes https://github.com/supabase/supabase-flutter/issues/579

### DIFF
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -26,6 +26,26 @@ typedef RealtimeDecode = void Function(
   void Function(dynamic result) callback,
 );
 
+/// Event details for when the connection closed.
+class RealtimeCloseEvent {
+  /// Web socket protocol status codes for when a connection is closed.
+  ///
+  /// The full list can be found at the following:
+  ///
+  /// https://datatracker.ietf.org/doc/html/rfc6455#section-7.4
+  final int code;
+
+  /// Connection closed reason sent from the server
+  ///
+  /// https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.6
+  final String? reason;
+
+  const RealtimeCloseEvent({
+    required this.code,
+    required this.reason,
+  });
+}
+
 class RealtimeClient {
   String? accessToken;
   List<RealtimeChannel> channels = [];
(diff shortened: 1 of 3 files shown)


packages/realtime_client/lib/src/realtime_client.dart
```diff
@@ -387,13 +407,17 @@ class RealtimeClient {
 
   /// communication has been closed
   void _onConnClose() {
-    final event = conn?.closeReason ?? '';
+    final statusCode = conn?.closeCode;
+    RealtimeCloseEvent? event;
+    if (statusCode != null) {
+      event = RealtimeCloseEvent(code: statusCode, reason: conn?.closeReason);
+    }
```
dshukertjr, line 414: JS client passes [ClosedEvent](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent) when the connection is closed, so I added something similar here.

packages/realtime_client/lib/src/realtime_client.dart
```diff
@@ -387,13 +407,17 @@ class RealtimeClient {
...
     log('transport', 'close', event);
 
     /// SocketStates.disconnected: by user with socket.disconnect()
     /// SocketStates.closed: NOT by user, should try to reconnect
     if (connState == SocketStates.closed) {
-      _triggerChanError(event);
+      _triggerChanError();
```
dshukertjr, line 420: Removing the `event` here to align the behavior with the JS client
https://github.com/supabase/realtime-js/blob/master/src/RealtimeClient.ts#L388

bdlukaa (approved):

packages/realtime_client/lib/src/realtime_client.dart
dshukertjr: Also wanted to share for reference that `conn.closeReason` was often an empty string where as `closeCode` had values whenever the connection is closed, so it should be more helpful.

packages/realtime_client/lib/src/realtime_client.dart
Vinzent03: I'm unsure about this change. This breaks the current heartbeat timeout test. This causes a channel error to be emitted, but without any reason.

dshukertjr: >This causes a channel error to be emitted, but without any reason.

This is the behavior of the JS SDK, so we are aligning the behavior there.

PR #622, opened 2023-09-10

dshukertjr:
- realtime_client@1.2.2
 - supabase@1.11.4
 - supabase_flutter@1.10.17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,36 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2023-09-10
+
+### Changes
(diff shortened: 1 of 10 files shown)


Vinzent03:
To be honest, I'm still very unsure about the changes in #620. On heartbeat timeout, it just emits a `CHANNEL_ERROR` and `CLOSED` event, with the data always being null. So you have no idea why it closed.

dshukertjr:
@Vinzent03 At least that is how the JS client library works as well, and emitting the error there was causing unwanted errors to be thrown. Maybe we can look at the particular issue you are pointing out in another PR, but for now, this version ships more bug fixes than creating bugs.

Vinzent03:
Hmm okay ...

PR #628, opened 2023-09-16
Add a Callback So We Can listen To Incoming Websocket Payloads From Supabase

MelbourneDeveloper:

This gives consumers the capacity to listen to all incoming web socket payloads

Consumers only see certain kinds of incoming payloads based on the PostgREST client

## What is the new behavior?

It is now possible to include a callback to log incoming payloads

```dart
supabase.client.from(table).stream(
  primaryKey: ['id'],
  onTrigger: (
    type, [
    payload,
    ref,
  ]) {
    //This converts the incoming payload back to JSON
    //and logs it in the console
    final json = jsonEncode(payload);
    print(json);
  },
).eq(column, value);
```

This is the output

image

## Additional context

This is only a rough draft, and I only added this to explain an issue. The naming is probably poor. It is called `onTrigger` because the Realtime client has a method called `trigger` where payloads arrive. The name can be changed to anything, and I can add API docs.

### DIFF
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -9,6 +9,8 @@ import 'package:realtime_client/src/retry_timer.dart';
 import 'package:realtime_client/src/transformers.dart';
 
 typedef BindingCallback = void Function(dynamic payload, [dynamic ref]);
+typedef TriggerCallback = void Function(String type,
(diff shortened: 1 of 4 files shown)


dshukertjr:
Thanks for this PR. I personally would probably lean towards not exposing low-level APIs as much as possible to the user, but could I ask what the expected use case for this new method be? Is it primary for debugging purposes?

Issue #627, opened 2023-09-16
PostgREST Dart Client - new payload in _getStreamData on 'UPDATE' missing column


MelbourneDeveloper:
* PostgreSQL version: PostgreSQL 15.1 (Ubuntu 15.1-1.pgdg20.04+1) on aarch64-unknown-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0, 64-bit
* PostgREST version: 1.5
* Operating system: macOS

## Issue One - Missing Column (resume)
Supabase seems to limit one of the columns in the payload of incoming websocket calls. This is probably just a quote issue, however, I'm not seeing any error logs on the back-end or through the websocket client.

If I send an UPDATE (PATCH) to the server for the `resume` column, the record appears in the payload via the websocket, and the record column (quite large) is in the payload. If I send an UPDATE for a different column, a similar payload appears, but the `resume` column is missing, perhaps because of a [quota issue](https://supabase.com/docs/guides/realtime/quotas).  

The problem is that I don't see anything in the back-end logs, and I don't see anything in the websocket payload

Server logs
<img width="1109" alt="image" src="https://github.com/supabase/supabase-flutter/assets/16697547/bbb85e0e-b36e-4d29-9bec-bda4721aae15">

Incoming Websocket Payloads
You can see here that it has removed the `resume` column.
```JSON
{
    "data": {
        "columns": [
            {
                "name": "id",
                "type": "varchar"
            },
            {
                "name": "resume",
                "type": "text"
            },
            {
                "name": "embedding",
                "type": "vector"
            },
            {
                "name": "minsalaryperday",
                "type": "int4"
            },
            {
                "name": "minsalaryperyear",
                "type": "int4"
            },
            {
                "name": "locationtext",
                "type": "varchar"
            },
            {
                "name": "location",
                "type": "geography"
            },
            {
                "name": "preferredworktypes",
                "type": "_text"
            },
            {
                "name": "preferredcommutetypes",
                "type": "_text"
            }
        ],
        "commit_timestamp": "2023-09-16T20:58:03.554Z",
        "errors": null,
        "old_record": {
            "id": "c315090a-0cca-4f07-adfa-78c6fe433f2a"
        },
        "record": {
            "embedding": null,
            "id": "c315090a-0cca-4f07-adfa-78c6fe433f2a",
            "location": "0101000020E61000004703780B24E842C0789CA223B9844140",
[...]

PR #631, opened 2023-09-18
fix: don't try to decode an empty body

Vinzent03:

Bug fix

A `FormatException` is thrown on when trying to parse an empty body

## What is the new behavior?

The empty body is caught and no decoding is happening.

## Additional context

#627

### DIFF
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -212,7 +212,9 @@ class PostgrestBuilder<T, S> implements Future<T> {
       int? count;
 
       if (response.request!.method != METHOD_HEAD) {
-        if (response.request!.headers['Accept'] == 'text/csv') {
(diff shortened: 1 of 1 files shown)


bdlukaa (approved):

dshukertjr (approved):

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -212,7 +212,9 @@ class PostgrestBuilder<T, S> implements Future<T> {
       int? count;
 
       if (response.request!.method != METHOD_HEAD) {
-        if (response.request!.headers['Accept'] == 'text/csv') {
+        if (response.bodyBytes.isEmpty) {
```
MelbourneDeveloper, line 215: @Vinzent03, this looks like it will address the exception. Exceptions do interrupt the debugging flow when there is actually nothing exceptional happening.

Thanks

PR #636, opened 2023-09-19
refactor!: many auth breaking changes

Vinzent03:
# Breaking changes
- use inAppWebView for `signInWithOAuth`
- remove `authCallbackUrlHostname`
- remove `intialSession` from `SubabaseAuth`
	- I noticed that if there is no old session to be recovered, `initialSession` would return null. So I added the `initialSession` event, which is fired with a null session when no session gets recovered. In all other cases, either an exception or a `signedIn` event is emitted.
	- gotrue-js has that `intialSession` event as well, but I haven't taken a look at how exactly it's used. Maybe they use that instead of `signedIn` for the first event.
- rename `Provider` to `OAuthProvider`
- use pkce instead of implicit auth flow by default

I've bundled these changes in one pr, because I think that's less tedious. So I think it's easier to review each commit by its own.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -652,7 +652,7 @@ class GoTrueClient {
 
     final errorDescription = url.queryParameters['error_description'];
     if (errorDescription != null) {
-      throw _notifyException(AuthException(errorDescription));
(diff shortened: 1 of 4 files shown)


dshukertjr (approved):
Thanks for this PR. This is great!

PR #637, opened 2023-09-20
fix: Remove error parameter on `_triggerChanError`

dshukertjr:

Bug fix.

When the device is offline, `SocketExceptions` and `WebSocketExceptions` gets thrown by the websocket connection stream [here](https://github.com/supabase/supabase-flutter/blob/main/packages/realtime_client/lib/src/realtime_client.dart#L168). This was causing rapid errors to be passed to the `.subscribe()` callback, and therefore causing rapid errors to be thrown when listening to data changes using `.stream()` method.

## What is the new behavior?

The errors from the web socket connection is not passed to `_triggerChanError`, which is aligned with the js SDK https://github.com/supabase/realtime-js/blob/master/src/RealtimeClient.ts#L402

## Additional context

This PR reverts some of the work done in https://github.com/supabase/realtime-dart/pull/62/files, but we can default to the same behavior as the js SDK for now and fix things if necessary in the future.

### DIFF
--- a/packages/realtime_client/lib/src/constants.dart
+++ b/packages/realtime_client/lib/src/constants.dart
@@ -9,7 +9,17 @@ class Constants {
   };
 }
 
-enum SocketStates { connecting, open, closing, closed, disconnected }
+enum SocketStates {
+  connecting,
+  open,
+  closing,
+
+  /// Socket being close not by the user. Realtime should attempt to reconnect.
+  closed,
+
+  /// Socket being closed by the user
+  disconnected,
+}
 
 enum ChannelStates { closed, errored, joined, joining, leaving }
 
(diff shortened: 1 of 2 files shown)


Vinzent03 (approved):

coolusaHD:
@dshukertjr Hey
I'm came accross theis PR while looking for a bugfix in my project.
I'm currently using the realtime function in my flutter app to retrieve live data. Currently when the devices loses connection I want to show th user that he is not connected to live data anymore.

I was using the `onError` function on the `stream.listen()` function.
With this update this no longer works.
As I am new to flutter I was wondering if you guys see any other way to get this behavior and help me a bit out ?

Thanks in advance. ✌️

dshukertjr:
@coolusaHD 
Thanks for the commend. That is a good point. We would have to come up with a way to detect the connection state for when using `.stream()`. Would you be able to open a issue for this in this repo?

coolusaHD:
@dshukertjr 
See #674

Thanks

PR #640, opened 2023-09-21
refactor!: create package specific configs

Vinzent03:
# Breaking Changes
- create package specific config classes in `supabase` and `supabase_flutter`
- make many fields private
- mark `SupabaseAuth.initialize` as internal as I see no reason to manually initialize. It depends on `Supabase.instance` so it can't be initialized alone either.

This aligns a bit with how supabase-js does this, but I've done the naming a bit different.

Actually, I noticed there is not really a reason for a global `SupabaseAuth.instance`, because it only provides access to the localstorage used. So I would be open to remove it and instead make it a simple object of the `Supabase` class in `supabase_flutter`.

### DIFF
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -1,14 +1,8 @@
 import 'dart:async';
 
-import 'package:functions_client/functions_client.dart';
-import 'package:gotrue/gotrue.dart';
 import 'package:http/http.dart';
-import 'package:postgrest/postgrest.dart';
-import 'package:realtime_client/realtime_client.dart';
-import 'package:storage_client/storage_client.dart';
 import 'package:supabase/src/constants.dart';
-import 'package:supabase/src/realtime_client_options.dart';
-import 'package:supabase/src/supabase_query_builder.dart';
+import 'package:supabase/supabase.dart';
 import 'package:yet_another_json_isolate/yet_another_json_isolate.dart';
 
 import 'auth_http_client.dart';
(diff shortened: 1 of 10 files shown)


packages/supabase/lib/src/supabase_client.dart
```diff
@@ -125,49 +120,47 @@ class SupabaseClient {
   SupabaseClient(
     this.supabaseUrl,
     this.supabaseKey, {
-    String? schema,
-    bool autoRefreshToken = true,
+    PostgrestClientOptions postgrestOptions = const PostgrestClientOptions(),
+    GoTrueClientOptions goTrueOptions = const GoTrueClientOptions(),
+    StorageClientOptions storageOptions = const StorageClientOptions(),
+    RealtimeClientOptions realtimeClientOptions = const RealtimeClientOptions(),
```
dshukertjr, line 126: For the naming conventions, could we rename `postgrest` to `db` and `gotrue` to `auth` so it's a bit more aligned with the js client?
https://github.com/supabase/supabase-js/blob/master/src/lib/types.ts

dshukertjr (commented):
Thanks for the amazing work! I had one little thing about naming conventions, but other than that it looks amazing. On your point about removing `SupabaseAuth` class, I am with you, and I think we can remove it and migrate all the functinalities to `Supabase` class!

packages/supabase/lib/src/supabase_client.dart
Vinzent03: So it's `DBClientOptions`? I feel like that's not really aligned to any other naming, because `db` is currently nowhere used. Or do you only mean the argument name and keep the class name?

dshukertjr: Hmm, yeah js naming convention might be a bit misleading or ambiguous. Let's stick to what you have here for Postgrest then. I think auth can be updated though.

Vinzent03: Agree, I've updated it now to `AuthClientOptions`

dshukertjr (approved):
Looks beautiful!

PR #644, opened 2023-09-26
refactor!: disallow multiple stream filter

Vinzent03:

Improves developer experience around filtering .stream()

Users could chain multiple filters on .stream() and it would not throw static error. Users could only find out that it's invalid after running the code.

```dart
supabase.from('table')
  .stream([])
  .eq('column', 'val')
  .neq('another_column', 'val')
  .lt('third_column', 'val')
  .listen((_) {});
  ```

## What is the new behavior?

The above example fails at the .new statically. Additionally, I exported the `SupabaseStreamBuilder` class

## Additional context

There was already a [pr](https://github.com/supabase/supabase-dart/pull/179), but I think the effort of creating classes for order and limit are worth it, because I see no sense in wanting to call the same method multiple times, so I don't disallow it.

### DIFF
--- a/packages/supabase/lib/src/supabase_stream_filter_builder.dart
+++ b/packages/supabase/lib/src/supabase_stream_filter_builder.dart
@@ -1,0 +1,152 @@
+part of './supabase_stream_builder.dart';
+
+class SupabaseStreamFilterBuilder extends SupabaseStreamBuilder {
+  SupabaseStreamFilterBuilder({
+    required super.queryBuilder,
+    required super.realtimeTopic,
+    required super.realtimeClient,
+    required super.schema,
+    required super.table,
+    required super.primaryKey,
+  });
+
+  /// Filters the results where [column] equals [value].
+  ///
+  /// Only one filter can be applied to `.stream()`.
+  ///
+  /// ```dart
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/supabase/lib/src/supabase_stream_filter_builder.dart [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+  /// Only one filter can be applied to `.stream()`.
+  ///
+  /// ```dart
+  /// supabase.from('users').stream(primaryKey: ['id']).eq('name', 'Supabase');
+  /// ```
+  SupabaseStreamBuilder eq(String column, Object value) {
+    assert(
+      _streamFilter == null,
+      'Only one filter can be applied to `.stream()`',
+    );
```
dshukertjr, line 24: We should be able to get rid of these assertions, right?

dshukertjr (approved):
Thanks for this. Just had one thing about assertion, but otherwise looks amazing!

packages/supabase/lib/src/supabase_stream_filter_builder.dart
Vinzent03: Good catch!

PR #647, opened 2023-09-28
docs: update readme to v2

Vinzent03:
# Changes

Updates the READMEs to the api changes of v2.

Note that I already pushed a merge of `main` into `next` branch without a pull request, because that was easier with the few merge conflicts. But these were only in tests, so everything should be good.

I think the first dev release should be ready after this pr. :tada:


dshukertjr (approved):
Let me go through the changes again next week, and let's shoot for releasing v2 dev preview on the 9th of October. This coming up release should be the last v1 release, and I think we can just merge `next` branch into `main` starting the 9th.

PR #649, opened 2023-10-01
fix(supabase)!: make Supabase credentials private in `SupabaseClient`

dshukertjr:

Make the following private, because users shouldn't need to access them using the client library.
- Supabase Key
- Supabase URL
- `realtime` (RealtimeClient instance)
- `rest` (PostgrestClient instance)

## Additional context

I feel like there might have been a use case where the `rest` needed be public, but I cannot find the issue that had context about it. I feel like it had something to do with headers, which I think should now be solved since we expose a setter, but if anyone has a use cases for any of these to be public, please do let me know!

## Edited:
Not making `rest` and `realtime` private https://github.com/supabase/supabase-flutter/pull/649#issuecomment-1743710220

### DIFF
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -49,8 +49,8 @@ class SupabaseClient {
 
   /// Supabase Storage allows you to manage user-generated content, such as photos or videos.
   late final SupabaseStorageClient storage;
-  late final RealtimeClient realtime;
-  late final PostgrestClient rest;
+  late final RealtimeClient _realtime;
+  late final PostgrestClient _rest;
   String? _changedAccessToken;
   late StreamSubscription<AuthState> _authStateSubscription;
   late final YAJsonIsolate _isolate;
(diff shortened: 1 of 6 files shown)


packages/supabase/lib/src/supabase_client.dart
```diff
@@ -94,40 +93,24 @@ class SupabaseClient {
...
-  ///
-  /// Custom http client can be used by passing [httpClient] parameter.
-  ///
-  /// [storageRetryAttempts] specifies how many retry attempts there should be to
-  ///  upload a file to Supabase storage when failed due to network interruption.
-  ///
-  /// [realtimeClientOptions] specifies different options you can pass to `RealtimeClient`.
-  ///
-  /// Pass an instance of `YAJsonIsolate` to [isolate] to use your own persisted
-  /// isolate instance. A new instance will be created if [isolate] is omitted.
```
dshukertjr, line 118: This chunk of comment is included in the macro below.

dshukertjr:
@Vinzent03 Thanks for the reminder. Made `rest` and `realtime` public again.

Vinzent03 (approved):

PR #650, opened 2023-10-02
refactor!:  remove `signInWithApple` method and make `generateRawNonce` public

dshukertjr:

I know this PR could be controversial, but I want to hear what you all think. This PR includes the following
- Removes the `signInWithApple()` method
- Make `generateRawNonce method available for developers to easily generate a secure nonce. 

I think it's okay to remove sign_in_with_apple dependencies, as I see a lot of developers implementing Google login despite it being far more complicated than signing in with Apple. I think by looking at this, we learned that developers will be happy as long as there is a way to implement Apple sign in, so I personally think we should remove sign_in_with_apple dependencies, but am open to comments. 

Related: https://github.com/supabase/supabase-flutter/issues/399#issuecomment-1610598390


Vinzent03:
I'm fine with this change. Wasn't there an issue of a user not being able to build the app because of this dependency, without even using it? Should make things cleaner!

Vinzent03 (approved):

DanMossa (approved):

PR #651, opened 2023-10-02
refactor!(supabase_flutter): Remove webview from dependencies

dshukertjr:

This could be another controversial PR. This PR basically removes the webview dependency added in this PR https://github.com/supabase/supabase-flutter/pull/355.

The original reason for adding the webview was to fix issues of browser window not automatically going back tot he app when performing `signInWithOAuth()`, but with Supabase having native OAuth support for Apple and Google login, I think the use cases of using the webview are limited, and I think we are rather safe just dropping it from the dependencies.

### DIFF
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -12,8 +12,6 @@ import 'package:meta/meta.dart';
 import 'package:sign_in_with_apple/sign_in_with_apple.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
 import 'package:url_launcher/url_launcher.dart';
-import 'package:url_launcher/url_launcher_string.dart';
-import 'package:webview_flutter/webview_flutter.dart';
 
 /// SupabaseAuth
 class SupabaseAuth with WidgetsBindingObserver {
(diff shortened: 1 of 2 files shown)


Vinzent03:
Isn't Apple forcing you to use webview for social login? Correct me if I'm wrong.

dshukertjr:
@Vinzent03 I think using an in app webview was a requirement, yes. So any time someone wants to implement OAuth login with any provider other than Apple or Google, they might have to implement it using webview by themselves. I personally think this is okay, because I feel it's a relatively rare case now that we have native login.

I think carrying a webview plugin as a dependency is a big risk when it goes through breaking changes disrupting supabase_flutter. I would rather have nice documentations around how to implement OAuth using webiview than having easy method to implement OAuth, which most people probably will not use.

Vinzent03:
On the one hand. I feel like these many OAuth providers and how easy they are to set up are a big selling point of Supabase, but on the other hand, it's a heavy dependency and I think at some point you want to implement the webview yourself any way to better adapt it to your design.
I just feel like we're losing the quick setup and go ability. I'm unsure and fine either way.

dshukertjr:
Yeah, there definitely pros and cons, but I personally didn't think it was right for us to take care of the UI too.

PR #653, opened 2023-10-02
fix: compile with webdev

### LINKED ISSUE #641: Cannot build web app with pure Dart
**Describe the bug**

Building a web app not using Flutter fails due to the universal_io package that is currently lacking support for web.

**To Reproduce**

1. Create a project with `dart create -t web`.
2. Run `dart pub add supabase` and `dart pub get` to start using the supabase package.
4. Prepare to use Supabase by calling the SupabaseClient constructor.
5. Run the app by `webdev serve`.
6. See the warnings printed in the console.
7. Open `localhost:8080` in a browser.

main.dart should look like this:

```dart
import 'dart:html';

import 'package:supabase/supabase.dart';

void main() {
  SupabaseClient(
    'Your Supabase URL',
    'Your Supabase Key',
  );

  querySelector('#output')?.text = 'Your Dart app is running.';
}
```

**Expected behavior**

There is no warning during a build, and "Your Dart app is running." is shown in the browser.

**Screenshots**

N/A

```
└── supabase 1.11.5
    ├── functions_client 1.3.2
    ├── gotrue 1.12.2
    ├── postgrest 1.5.1
    ├── realtime_client 1.2.2
    ├── storage_client 1.5.2
```

**Additional context**

```
[WARNING] build_web_compilers:entrypoint on web/main.dart: Skipping compiling bug|web/main.dart with ddc because some of its
transitive libraries have sdk dependencies that not supported on this platform:

universal_io|lib/src/_exports_in_vm.dart
universal_io|lib/src/internet_address.dart
universal_io|lib/src/http_client.dart
```

Vinzent03:
Bug fix

- We only catch `SocketException`, which in only thrown in native Dart, but not for example in dart2js
- We use `File` from `dart:io`

## What is the new behavior?

- We instead use `ClientException`, which is also thrown in js runtime as documented [here](https://pub.dev/documentation/http/latest/http/Client-class.html)
- Created a stub for `File` to be able to compile the package with `webdev`

## Additional context
~~This was meant to fix #641, but this would require the removal of `File`, which im very unsure about. I think it makes things MUCH easier to be able to work with `File` on native platform, but somehow not being able to compile this to dart web is a bummer. I'm wondering why this package compiles for Flutter web though.~~
close #641

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -29,7 +29,7 @@ part 'gotrue_mfa_api.dart';
 /// [asyncStorage] local storage to store pkce code verifiers. Required when using the pkce flow.
 ///
 /// Set [flowType] to `AuthFlowType.pkce` to perform pkce auth flow.
-/// /// {@endtemplate}
+/// {@endtemplate}
 class GoTrueClient {
   /// Namespace for the GoTrue API methods.
   /// These can be used for example to get a user from a JWT in a server environment or reset a user's password.
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 import 'dart:math';
 
 import 'package:collection/collection.dart';
@@ -13,7 +14,6 @@ import 'package:http/http.dart';
 import 'package:jwt_decode/jwt_decode.dart';
 import 'package:meta/meta.dart';
 import 'package:rxdart/subjects.dart';
-import 'package:universal_io/io.dart';
 
 part 'gotrue_mfa_api.dart';
 
@@ -966,7 +966,7 @@ class GoTrueClient {
       _notifyAllSubscribers(AuthChangeEvent.tokenRefreshed);
       _refreshTokenCompleter!.complete(authResponse);
       return authResponse;
-    } on SocketException {
+    } on ClientException {
       _setTokenRefreshTimer(
         Constants.retryInterval * pow(2, _refreshTokenRetryCount),
         refreshToken: token,
--- a/packages/gotrue/pubspec.yaml
+++ b/packages/gotrue/pubspec.yaml
@@ -13,7 +13,6 @@ dependencies:
   crypto: ^3.0.2
   http: '>=0.13.0 <2.0.0'
   jwt_decode: ^0.3.1
-  universal_io: ^2.0.4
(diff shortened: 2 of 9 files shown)


dshukertjr:
It sounds like dart2js support will not be possible without removing the use of `File`, correct? 

How about we changing the `.storage.upload()` method to accept a Uint8List and get rid of the `.storage.uploadBinary()` method. It's a bit disruptive, but I always thought having two methods is a bit confusing. It might provide better DX if we had a single method for those storage operations.

dshukertjr:
It might be also pretty disruptive to drop support for uploading `File` object, so might not be a good idea. I personally wouldn't worry too much about supporting dart2js though. The use cases are very limited, and the SDK works just fine on Flutter Web. I'm more than happy to just close https://github.com/supabase/supabase-flutter/issues/641 as won't do.

Vinzent03:
I think I was able to fix the issue :tada: I created a stub for `File` which is used on js platform. The `webdev` build went well with only stubbing the used methods. This is actually no breaking change, so we could also merge this in main. :thinking:

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -951,7 +951,7 @@ class GoTrueClient {
       notifyAllSubscribers(AuthChangeEvent.tokenRefreshed);
       _refreshTokenCompleter!.complete(authResponse);
       return authResponse;
-    } on SocketException {
```
dshukertjr, line 954: I'm a bit confused about what is going on here. How can the thrown error type change without changing anything within the try block? Wouldn't `.request()` call on line 942 still throw SocketException when there is an issue with the network?

Vinzent03: From the [doc](https://pub.dev/documentation/http/latest/io_client/IOClient-class.html)
> If there is a socket-level failure when communicating with the server (for example, if the server could not be reached), [IOClient](https://pub.dev/documentation/http/latest/io_client/IOClient-class.html) will emit a [ClientException](https://pub.dev/documentation/http/latest/http/ClientException-class.html) that also implements [SocketException](https://api.dart.dev/stable/3.1.2/dart-io/SocketException-class.html). This allows callers to get more detailed exception information for socket-level failures, if desired.

So any SocketException from the HttpClient gets caught and put into a ClientException that also implements SocketException, so you can use both to catch the exception. See their [send implementation](https://pub.dev/documentation/http/latest/io_client/IOClient/send.html).

dshukertjr (approved):

Vinzent03:
I moved the commits now to the main branch, so we can ship this as well to v1.

PR #655, opened 2023-10-05

dshukertjr:
- realtime_client@1.3.0
 - supabase@1.11.7
 - supabase_flutter@1.10.20
 
(Hopefully) the last release for v1 🚀

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,36 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2023-10-05
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - There are no breaking changes in this release.
+
+Packages with other changes:
+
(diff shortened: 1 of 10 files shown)


Vinzent03:
I've merged now #653 as well into main, so this should be integrated into the changelog.

dshukertjr:
I think the easiest would be to just do two releases right now. Would that be fine? I will create another release right after merging this one.

Vinzent03:
Because of the tags right? The branch management should be easy, but you would have to remove the tags. I'm not really familiar with how that's handled, so if that's easier for you, do two releases.

dshukertjr:
@Vinzent03 Yeah, let's just do two releases. Do you want to approve this PR?

Vinzent03:
But creating another release with the #653  being in history before this is merged might also create issues. I actually think it's easier to try to remove this release.

Vinzent03 (approved):

dshukertjr:
@Vinzent03 It's alright, I would just publish it from `chore/publish` branch.

PR #669, opened 2023-10-14
Optionally ignore deep links

gadfly361:

This PR adds the ability to prevent supabase_flutter from handing deep links (which affords the opportunity for a user to fully handle deep linking on their own). This is done by adding a parameter `ignoreDeepLinks` to  the`Supabase.initialize` method.

This PR is another attempt to address #541.

Currently, supabase_flutter handles deep linking and there is no way for a user of supabase_flutter to opt out. A good summary of the issue can be found here: #541.

A different PR (#605) was made to tackle that issue, but it doesn't seem like it will be accepted. #605 was attempting to expose any deep links that supabase_flutter didn't handle to the user. However, the following was said in response to #605:

> Instead of fixing this in supabase-flutter, I would instead fix that in app_links so that they create only one stream and broadcast that.

This PR in contrast, is simply allowing for an escape hatch to stop supabase_flutter from handing any deep links, so a user can fully handle them on their own.

## What is the new behavior?

When `ignoreDeepLinks` is `true`, then supabase_flutter doesn't handle deep links at all.

### DIFF
--- a/packages/supabase_flutter/lib/src/supabase.dart
+++ b/packages/supabase_flutter/lib/src/supabase.dart
@@ -79,6 +79,7 @@ class Supabase {
     AuthFlowType authFlowType = AuthFlowType.implicit,
     GotrueAsyncStorage? pkceAsyncStorage,
     bool? debug,
+    bool ignoreDeepLinks = false,
   }) async {
     assert(
       !_instance._initialized,
@@ -103,6 +104,7 @@ class Supabase {
       localStorage: localStorage ?? const HiveLocalStorage(),
       authCallbackUrlHostname: authCallbackUrlHostname,
       authFlowType: authFlowType,
+      ignoreDeepLinks: ignoreDeepLinks,
     );
 
     return _instance;
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -56,6 +56,8 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   final _appLinks = AppLinks();
 
+  bool _ignoreDeepLinks = false;
+
   /// A [SupabaseAuth] instance.
   ///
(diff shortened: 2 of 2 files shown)


point-source:
I like this direction since it allows the supabase_flutter dependency to be less intrusive by default. However, I also like that supabase_flutter has the ability to handle these links. Perhaps in addition to the ignore arg, we could also add a method to the library which can act as a callback for handling a specific deep link.

For example, I use the auto_route package to catch my deep links. This is useful because it puts the links in the same place as routing is happening and I can benefit from context, guards, and other features that come for free that way. Right now, I have to use the pure-dart version of supabase in order to allow auto_route to function. This means I need to handle session persistence manually. What if I could just turn off supabase's deep link handling and then add to auto_route's handler like so:

```dart
            routerConfig: _appRouter.config(
              deepLinkBuilder: (deepLink) {
                if (deepLink.path.contains('special_route')) {
                  // Go somewhere special
                } else {
                  final result = await Supabase.instance.client.handleDeepLink(deepLink);
                  if (result.handled) return const DeepLink([SupaRoute()]);
                }

                return const DeepLink([HomeRoute()]);
              },
            ),
```

This should hopefully allow for the best of both worlds without requiring [a change to app_links](https://github.com/llfbandit/app_links/issues/80) which [may not even be possible](https://github.com/llfbandit/app_links/issues/68).

dshukertjr:
@point-source 
You can call the `getSessionFromUrl` method to get the session from deep links.
```dart
routerConfig: _appRouter.config(
  deepLinkBuilder: (deepLink) {
    if (deepLink.path.contains('special_route')) {
      // Go somewhere special
    } else {
      await Supabase.instance.client.auth.getSessionFromUrl(deepLink);
    }

    return const DeepLink([HomeRoute()]);
  },
),
```

gadfly361:
Closing in favor of #848

PR #670, opened 2023-10-16
chore: merges main into next

dshukertjr:

Merges `main` into `next` Once this is good. We can do a release for v2 developer preview. Let's actually keep the `main` branch and `next` branch separated until v2 comes out of developer preview. The duration of the developer preview is a bit arbitrary, but let's keep it developer preview for 4 weeks just in case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,101 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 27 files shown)


Vinzent03:
I would prefer if you would merge this locally with a merge commit and push the changes to `next` as I did last time, because you have now combined all commits into one, losing the correct association to the individual commit. This might be picky, but I feel like it's strange to commit this in a squash.

dshukertjr:
Cool, closing in favor of #672

PR #673, opened 2023-10-18
chore: publish v2.0.0-dev.0

dshukertjr:
Also includes a small change to `melos.yaml` to copy the proper version number long with `dev` suffix to each `version.dart` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,93 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2023-10-18
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - [`functions_client` - `v2.0.0-dev.0`](#functions_client---v200-dev0)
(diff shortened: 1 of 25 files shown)


Vinzent03:
I think the postgrest builder and supabase package config changes should be documented. So how to migrate to v2.

dshukertjr:
@Vinzent03 Yeah, there are a lot of documentations to be done. I will have them by the time this comes out of developer preview.

Vinzent03:
The reason behind a dev version is that some people can already try and test the new version. But without documentation on how to migrate, less people will take that effort.
I think at least the new count and config API should be easy visible in the changelog.
I also think it's standard to do that migration guide for dev version as well.

dshukertjr:
@Vinzent03 This v2 is a relatively minor upgrade from v1. Most of the APIs are the same, and those that have changed have changed in an intuitive way with enough comment documentation. Early adaptors that want to try out v2 developer preview should be fine with it. This developer preview still has possibilities of changing, so I will document the final result.

Vinzent03 (approved):

Issue #666, opened 2023-10-11
Cannot use `SupabaseClient` without dart:html or dart:io in Edge functions


maksym-tielnyi-solid:
**Describe the bug**
The `SupabaseClient` constructor throws `"Cannot create a client without dart:html or dart:io."` error when used in Dart Edge functions.
An error message looks like the following:
```
[Error] ub {
  a: "Cannot create a client without dart:html or dart:io.",
  "$thrownJsError": Unsupported operation: Cannot create a client without dart:html or dart:io.
    at Object.J (file:///home/deno/functions/dart_edge/main.dart.js:246:3)
    at Object.v (file:///home/deno/functions/dart_edge/main.dart.js:253:14)
    at Object.wV (file:///home/deno/functions/dart_edge/main.dart.js:2861:15)
    at Object.e2 (file:///home/deno/functions/dart_edge/main.dart.js:2859:14)
    at Object.Xa (file:///home/deno/functions/dart_edge/main.dart.js:2685:3)
    at e.$1 (file:///home/deno/functions/dart_edge/main.dart.js:5492:8)
    at file:///home/deno/functions/dart_edge/main.dart.js:5516:19
    at Gs.a (file:///home/deno/functions/dart_edge/main.dart.js:1401:63)
    at Gs.$2 (file:///home/deno/functions/dart_edge/main.dart.js:3545:14)
    at Object.DI (file:///home/deno/functions/dart_edge/main.dart.js:1381:11)
}
```

**To Reproduce**
`main.dart`:
```dart
import 'package:edge_http_client/edge_http_client.dart';
import 'package:supabase/supabase.dart';
import 'package:supabase_functions/supabase_functions.dart';

void main() {
  SupabaseFunctions(fetch: (request) {
    final supabaseUrl = Deno.env.get('SUPABASE_URL')!;
    final supabaseKey = Deno.env.get('SUPABASE_SERVICE_ROLE_KEY')!;

    // This SupabaseClient constructor throws an error:
    SupabaseClient(
      supabaseUrl,
      supabaseKey,
      httpClient: EdgeHttpClient(),
    );

    return Response('OK');
  });
}
```

`pubspec.yaml`:
```dart
name: supabase_flutter_test
description: A sample command-line application.
version: 1.0.0

environment:
  sdk: ^3.0.6

dependencies:
  edge_http_client: ^0.0.1+3
  supabase: 1.11.9
  supabase_functions: 0.0.2+4

dev_dependencies:
[...]

PR #675, opened 2023-10-20
fix: make Supabase client work in Dart Edge again

dshukertjr:

Beats me why this fixes it, but with this update SupabaseClient can be initialized within Dart Edge environment again.

fixes https://github.com/supabase/supabase-flutter/issues/666


Vinzent03:
The actual root issue is that the no http client was passed to realtime. Fixed in now.

Vinzent03 (approved):

PR #680, opened 2023-10-25
fix(gotrue): allow empty session response for verifyOtp method

dshukertjr:

Calling `auth.verifyOtp()` method with the 6-digit OTP returns an `AuthResponse` with `null` session. The supabase_flutter SDK currently however does not allow `null` session returned from the `verifyOtp()` method, but this PR enables empty session to be returned.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -495,14 +495,10 @@ class GoTrueClient {
 
     final authResponse = AuthResponse.fromJson(response);
 
-    if (authResponse.session == null) {
-      throw AuthException(
-        'An error occurred on token verification.',
-      );
+    if (authResponse.session != null) {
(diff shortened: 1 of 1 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -495,15 +495,11 @@ class GoTrueClient {
 
     final authResponse = AuthResponse.fromJson(response);
 
-    if (authResponse.session == null) {
-      throw AuthException(
-        'An error occurred on token verification.',
-      );
```
dshukertjr, line 501: This piece of code is mirroring [this](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L673) js SDK code, but we actually need to allow empty session.

We shouldn't need the `!data` check equivalent here, because `_fetch.request` will throw if it did not have any response data.

DanMossa (approved):

PR #702, opened 2023-11-10
fix(supabase_flutter): The session is not restored when the application is started.

rickypid:
Bug fix the session is not restored when the application is started #698

### DIFF
--- a/packages/supabase_flutter/CHANGELOG.md
+++ b/packages/supabase_flutter/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 2.0.0-dev.2
-
-- **FIX**(supabase_flutter): The session is not restored when the application is started. ([#698](https://github.com/supabase/supabase-flutter/issues/698)).
-
 ## 2.0.0-dev.1
 
  - Update a dependency to the latest release.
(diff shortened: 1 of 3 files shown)


packages/supabase_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.0-dev.2
+
+- **FIX**(supabase_flutter): The session is not restored when the application is started. ([#698](https://github.com/supabase/supabase-flutter/issues/698)).
```
dshukertjr, line 3: This will actually be automatically populated during package release, so no need to add it now.

packages/supabase_flutter/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: supabase_flutter
 description: Flutter integration for Supabase. This package makes it simple for developers to build secure and scalable products.
-version: 2.0.0-dev.1
+version: 2.0.0-dev.2
```
dshukertjr, line 3: This one as well will automatically be updated during package release, so no need to change it now.

dshukertjr (commented):
Thanks for this. I left a few comments, but if you could remove the changes on the changelog and pubspec.yaml, I think we can get this one merged in!

dshukertjr (approved):
Thanks for this PR! LGTM! A new version will be released on Monday!

PR #712, opened 2023-11-17
Fix supabase_flutter not supporting web due to Hive

DanMossa:

## What is the new behavior?

## Additional context


nokia6290:
Hive is just a local database. You can use any kind of database that caches locally data.

Vinzent03:
I'll close the pr as we should be pretty good with the current migration way to not cause any issues on web.

PR #721, opened 2023-11-21
Feat: adds geojson support for working with the PostGIS extension

encima:

Adds `geojson` support for returning PostGIS datatypes

Currently, geojson is not supported.

## What is the new behavior?

Adds a `geojson` method to pass the postgis content type and tell PostgREST to respond appropriately https://postgrest.org/en/stable/how-tos/working-with-postgresql-data-types.html#ww-postgis

## Additional context

Enabled the `PostGIS` extension for the test cases also (mentioning here for search-ability in case it introduces issues down the line)


packages/postgrest/lib/src/postgrest_transform_builder.dart
```diff
@@ -226,6 +226,18 @@ class PostgrestTransformBuilder<T> extends RawPostgrestBuilder<T, T, T> {
     return _copyWithType(method: METHOD_HEAD);
   }
 
+  /// Enables support for GeoJSON for use with PostGIS data types
+  /// Used when you need the complete response to be in GeoJSON format.
+  /// You will need to enable the PostGIS extension for this to work.
+  ///
+  /// https://supabase.com/docs/guides/database/extensions/postgis
+  ///
+  ResponsePostgrestBuilder geojson() {
```
dshukertjr, line 235: I think we can type cast it here to specify the response type. I know having to write the same type three times seems redundant, but it is a way that one of the maintainers brilliantly came up with to support the complex types that Supabase provides. 

```suggestion
  ResponsePostgrestBuilder<Map<String, dynamic>, Map<String, dynamic>,
      Map<String, dynamic>> geojson() {
```

dshukertjr (approved):
LGTM!

Issue #718, opened 2023-11-20
range() No effect, only the first n items are returned.


disonwei:
**Describe the bug**

I am using flutter as well as supabase database. I'm going to query the data by paging. The range() method mentioned in the documentation is used.
But I now range() does not work.
 For example,

```dart
 final data = await supabase
.from('record')
.select()
.range(start, end); 
```

When start = 0 and end = 2, the first 3 results are returned, when start = 3 end = 5, the first 3 results are returned, and when start = 6 end = 8, the first 3 results are returned.. range(start, end); It did not take effect, and all the first n results were returned.

**Expected behavior**
range(start, end); should be the correct offset query data, not always return the first few items

**Screenshots**
The following is loading more logic, as well as the output result：

<img width="536" alt="image" src="https://github.com/supabase/supabase-flutter/assets/55089487/ee01ea54-1f84-46e6-8ae7-f5550096876d">

<img width="705" alt="image" src="https://github.com/supabase/supabase-flutter/assets/55089487/34384dac-2e3f-4baa-9bc1-cfeb705beafb">

<img width="756" alt="image" src="https://github.com/supabase/supabase-flutter/assets/55089487/f5925912-3887-4c85-9ca7-73354a249170">

**Version (please complete the following information):**
macOS

➜ flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.16.0, on macOS 14.1.1 23B81 darwin-arm64, locale zh-Hans-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 15.0.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2022.3)
[✓] VS Code (version 1.84.2)
[✓] Connected device (4 available)
[✓] Network resources

  supabase_flutter: ^2.0.0-dev.2

dshukertjr:
Closing as this one is fixed.

PR #722, opened 2023-11-21
fix: issue where `.range()` not respecting the offset given.

dshukertjr:

fixes https://github.com/supabase/supabase-flutter/issues/718

### DIFF
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -317,7 +317,8 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
   ///
   /// [url] may be used to update based on a different url than the current one
   Uri appendSearchParams(String key, String value, [Uri? url]) {
-    final searchParams = Map<String, dynamic>.from(_url.queryParametersAll);
+    final searchParams =
+        Map<String, dynamic>.from((url ?? _url).queryParametersAll);
     searchParams[key] = [...searchParams[key] ?? [], value];
     return (url ?? _url).replace(queryParameters: searchParams);
   }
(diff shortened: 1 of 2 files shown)


packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -317,18 +317,19 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
...
+        Map<String, dynamic>.from((url ?? _url).queryParametersAll);
     searchParams[key] = [...searchParams[key] ?? [], value];
     return (url ?? _url).replace(queryParameters: searchParams);
   }
 
   /// Get new Uri with overridden queryParams
   ///
   /// [url] may be used to update based on a different url than the current one
-  Uri overrideSearchParams(String key, String value, [Uri? url]) {
+  Uri overrideSearchParams(String key, String value) {
```
dshukertjr, line 329: Removed the `url` parameter since it was not used anywhere.

Vinzent03: I think we should keep it to allow these similar methods to have the same parameters. But I see that's not really important.

Vinzent03 (commented):
Where does this exactly fix the issue? Fixing that in `appendSearchParams` the correct url is used shouldn't change anything to `range()`, because it doesn't pass a custom url.

packages/postgrest/lib/src/postgrest_builder.dart
dshukertjr: I personally think we can just add it back when we actually need it. I think it creates extra confusion for having a piece of code that is not being used.

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -317,18 +317,19 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
   ///
   /// [url] may be used to update based on a different url than the current one
   Uri appendSearchParams(String key, String value, [Uri? url]) {
-    final searchParams = Map<String, dynamic>.from(_url.queryParametersAll);
+    final searchParams =
+        Map<String, dynamic>.from((url ?? _url).queryParametersAll);
```
dshukertjr, line 321: @Vinzent03 
On line 124 [here](https://github.com/supabase/supabase-flutter/blob/next/packages/postgrest/lib/src/postgrest_transform_builder.dart#L124), we generate a new URL that contains the `offset`. 

In the next line on 125, we are adding another parameter with the URL created on line 124, but because we were not using the `url` parameter to create the `searchParams` variable, what we did on line 124 is being ignored, therefore the offset is never set.

Vinzent03: Oh yeah, makes sense. I checked that code, but I think I viewed the code on the master branch by accident.

Issue #716, opened 2023-11-20
Session is removed when app is resumed without network


giorgiogross:
**Describe the bug**
The locally stored auth credentials of the current session are removed when app is resumed without network after user already successfully logged in before.

**To Reproduce**
1. Launch an app with Supabase Auth while you have access to network.
2. Log in
3. Turn on airplane mode
4. Leave app, but don't close it! Eg go to your home screen
5. Wait for the access token to expire. Eg set the session duration to 30sec in Supabase dashboard.
6. Then return to the app. I set a breakpoint at my currently installed supabase version in `supabase_auth.dart -> _recoverSupabaseSession()` I can see this: <img width="764" alt="Screenshot 2023-11-20 at 11 25 56" src="https://github.com/supabase/supabase-flutter/assets/18177722/da7cf7bf-58f5-46ca-b03a-5ff505e0a844">
7. You can see that the local session has been removed
8. Fully close the app, and restart it. Supabase attempts to resume the session, but the session got deleted. Therefore, your user is logged out.

On physical iOS devices, this seems to be a bit tricker: The above behaviour happens already when launching the app while device is in airplane mode. As It's hard for me to debug this, I can only make an assumption atm. From experience, connection errors that come from active airplace mode resolve almost immediately, so that it could be possible that the `WidgetsBindingObserver` which `SupabaseAuth`  is implementing attaches much quicker than in simulator debug builds (where I simply turn off Wifi) so that the `didChangeAppLifecycleState` is already triggered when the app is launched. So when my state management logic checks supabase's auth store, it already sees that the session is delted on app-launch, and therefore navigates to the sign in screen.

**Expected behavior**
I can launch and resume the app while offline, and app can go offline and online while using it, and session never gets deleted unless user explicitly signs out.

**Version (please complete the following information):**
On macOS, tested on iOS Simulator & iOS physical device
```
── supabase_flutter 2.0.0-dev.2
│   ├── supabase 2.0.0-dev.2
│   │   ├── functions_client 2.0.0-dev.0
│   │   ├── gotrue 2.0.0-dev.1
│   │   ├── postgrest 2.0.0-dev.0
│   │   ├── realtime_client 2.0.0-dev.1
│   │   ├── storage_client 2.0.0-dev.0

```

**Additional context**
This also happens on any other kind of network error, eg time outs. In my app, I'm passing my own instance of `SharedPreferencesLocalStorage` for the local storage so that my app can check cached session data upon launching.

[...]

PR #723, opened 2023-11-22
fix(supabase_flutter): session recovery while offline does not remove the existing session.

dshukertjr:

Currently if the Flutter SDK calls `_recoverSupabaseSession()` and `auth.recoverSession()` throws, the stored session is deleted within supabase_flutter. This is unnecessary as gotrue will catch the errors and handle them [here](https://github.com/supabase/supabase-flutter/blob/next/packages/gotrue/lib/src/gotrue_client.dart#L960). 

This PR simply removes the supabase_flutter code that is deleting the session stored on local storage. 

fixes https://github.com/supabase/supabase-flutter/issues/716

### DIFF
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -100,23 +100,23 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   /// Recover/refresh session if it's available
   /// e.g. called on a splash screen when the app starts.
-  Future<bool> _recoverSupabaseSession() async {
+  Future<void> _recoverSupabaseSession() async {
     final bool exist = await _localStorage.hasAccessToken();
     if (!exist) {
-      return false;
+      return;
     }
 
     final String? jsonStr = await _localStorage.accessToken();
     if (jsonStr == null) {
-      return false;
+      return;
     }
 
     try {
       await Supabase.instance.client.auth.recoverSession(jsonStr);
-      return true;
     } catch (error) {
-      _localStorage.removePersistedSession();
-      return false;
+      // When there is an exception thrown while recovering the session,
+      // the appropriate action (retry, revoking session) will be taken by
+      // the gotrue library, so need to do anything here.
     }
   }
 


packages/supabase_flutter/lib/src/supabase_auth.dart
```diff
@@ -100,23 +100,23 @@ class SupabaseAuth with WidgetsBindingObserver {
 
   /// Recover/refresh session if it's available
   /// e.g. called on a splash screen when the app starts.
-  Future<bool> _recoverSupabaseSession() async {
+  Future<void> _recoverSupabaseSession() async {
```
dshukertjr, line 103: Changing the return type is not related to fixing the issue for this PR, but since the returned value was not used anywhere, I thought it would reduce any confusions in the future.

packages/supabase_flutter/lib/src/supabase_auth.dart
```diff
@@ -100,23 +100,23 @@ class SupabaseAuth with WidgetsBindingObserver {
...
 
     try {
       await Supabase.instance.client.auth.recoverSession(jsonStr);
-      return true;
     } catch (error) {
-      _localStorage.removePersistedSession();
-      return false;
+      // When there is an exception thrown while recovering the session,
+      // the appropriate action (retry, revoking session) will be taken by
+      // the gotrue library, so need to do anything here.
```
dshukertjr, line 119: This is the main change for this PR.

Vinzent03:
I think by removing that code, we may create a case where at every app start the user gets an exception that the session couldn't be recovered. One case I see is where in `recoverSession` `Session expired.` is thrown.

I think a better solution is to check whether the error is an `AuthException`. The `ClientException` from `http` should therefore not trigger the removal.

dshukertjr:
@Vinzent03 Good call. With only the changes in this PR there might be issues as you pointed out. 

I have merged this PR to https://github.com/supabase/supabase-flutter/pull/726. It's a bit of a big PR, but I think it addresses the issue you have pointed out. I'm going to close this PR, so if you could take a look at https://github.com/supabase/supabase-flutter/pull/726, that would be awesome!

PR #725, opened 2023-11-22
feat!(realtime_client): Introduce type safe realtime methods

dshukertjr:

* This is just a proposal, and would love to hear what you think about this idea. 

Currently, the realtime methods are not type safe. This is due to how `channel.on()` method has to be able to support postgres changes, broadcast, and presence. 

The proposal is to introduce three different methods `onPostgresChanges`, `onBroadcast`, and `onPresence` that will each handle their respective realtime features. That way, we would be able to restrict what parameters to accept or the type of the payload.

The public API would look something like this:

```dart
final channel = supabase.channel('name');

channel
  .onPostgresChanges(
      event: PostgresChangeEvent.all,
      schema: 'public',
      table: 'my_table',
      filter: PostgresChangeFilter(
        type: PostgresChangeFilterType.eq,
        column: 'id',
        value: 1,
      ),
      callback: (PostgresChangePayload payload) {
        // `PostgresChangePayload` is a new class dedicated for this payload
        final Map<String, dynamic> newRecord = payload
            .newRecord; // Wanted to use `new`, but `new` is a reserved keyword in Dart, so couldn't use it.
        final Map<String, dynamic> oldRecord = payload.oldRecord;
      })
  .onBroadcast(
      event: 'broadcast_event',
      callback: (Map<String, dynamic> payload) {
        print(payload);
      })
  .onPresence(
      event: PresenceEvent.join,
      callback: (Map<String, dynamic> payload) {
        print(payload);
      })
  .subscribe();

await myChannel.sendBroadcastMessage(
  event: 'cursor-pos',
  payload: {'x': 30, 'y': 50},
);
```

Open to hearing any opinions about it. I'm sure the proposed public API isn't the best design either, so would love to hear anything regarding it as well.


Vinzent03:
I think I really like the new design! I haven't worked much with presence or broadcast, but it was always strange to have all arguments for all realtime functionalities in one method. It's easier to read as well.

dshukertjr:
Sounds great! I will finish up this PR!

dshukertjr:
I have updated the `filter` to take `PostgresChangeFilter` object like this:

```dart
final channel = supabase.channel('name');

channel
  .onPostgresChanges(
      event: PostgresChangeEvent.all,
      schema: 'public',
      table: 'my_table',
      filter: PostgresChangeFilter(
        type: PostgresChangeFilterType.eq,
        column: 'id',
        value: 1,
      ),
      callback: (PostgresChangePayload payload) {
        // `PostgresChangePayload` is a new class dedicated for this payload
        final Map<String, dynamic> newRecord = payload
            .newRecord; // Wanted to use `new`, but `new` is a reserved keyword in Dart, so couldn't use it.
        final Map<String, dynamic> oldRecord = payload.oldRecord;
      })
  .onBroadcast(
      event: 'broadcast_event',
      callback: (Map<String, dynamic> payload) {
        print(payload);
      })
  .onPresence(
      event: PresenceEvent.join,
      callback: (Map<String, dynamic> payload) {
        print(payload);
      })
  .subscribe();
```

dshukertjr:
Also, currently we use the `send()` method to send broadcast messages, but since `send()` is used internally it has som extra arguments. I have created a `sendBroadcastMessage()` method to simplify the experience of using it. 

```dart
await myChannel.sendBroadcastMessage(
  event: 'cursor-pos',
  payload: {'x': 30, 'y': 50},
);
```

dshukertjr:
Once this PR, https://github.com/supabase/supabase-flutter/pull/736, and https://github.com/supabase/supabase-flutter/pull/737 are in, I think we can finally release a stable v2. Will update the docs on supabase.com website too!

packages/supabase/lib/src/supabase_stream_builder.dart
```diff
@@ -141,71 +139,71 @@ class SupabaseStreamBuilder extends Stream<SupabaseStreamEvent> {
   Future<void> _getStreamData() async {
     final currentStreamFilter = _streamFilter;
     _streamData = [];
-    String? realtimeFilter;
+    PostgresChangeFilter? realtimeFilter;
     if (currentStreamFilter != null) {
-      if (currentStreamFilter.type == _FilterType.inFilter) {
+      if (currentStreamFilter.type == PostgresChangeFilterType.inFilter) {
```
Vinzent03, line 144: I think this if can be removed as both cases are now the same.

packages/realtime_client/test/mock_test.dart
```diff
@@ -38,41 +38,6 @@ void main() {
                   ['postgres_changes']
               .first['filter'];
 
-          final replyString = jsonEncode({
```
Vinzent03, line 41: Whats the reason for this code block to be removed?

packages/realtime_client/lib/src/realtime_channel.dart
```diff
@@ -248,12 +248,47 @@ class RealtimeChannel {
...
+  }) {
+    return onEvents(
+      'postgres_changes',
+      ChannelFilter(
+        event: event.toRealtimeEvent(),
+        schema: schema,
+        table: table,
+        filter: filter?.toString(),
+      ),
+      (payload, [ref]) => callback(PostgresChangePayload.fromPayload(payload)),
```
Vinzent03, line 266: Previously, the user was able to obtain `ref`. This is now gone. I'm wondering if there is any real use case for it.

Vinzent03 (changes requested):

packages/realtime_client/lib/src/realtime_channel.dart
dshukertjr: Good point. Honestly not sure how the `ref` can be used as a developer, but the js SDK doesn't have it anymore, so we should be good to remove it. 
https://github.com/supabase/realtime-js/blob/master/src/RealtimeChannel.ts#L350

packages/supabase/lib/src/supabase_stream_builder.dart
dshukertjr: I might not be following you, but wouldn't we still need to check if the filter being applied is `inFilter` here?

packages/realtime_client/test/mock_test.dart
dshukertjr: Yeah, on line 72 here we have `'topic': 'realtime:public:todos'`, and on each of the test cases, we have `client.channel('public:todoos')`.

On this reply, we were replying to `todos`, but the channel each of the test client was subscribed to was `todoos` (extra "o"), which means this reply was ignored in the tests, and never was doing anything. 

I think we copied it over from the tests on supabase package [here](https://github.com/supabase/supabase-flutter/blob/main/packages/supabase/test/mock_test.dart#L132) and didn't notice the typo on the channels until now, therefore just had some dead code.

packages/supabase/lib/src/supabase_stream_builder.dart
Vinzent03: I don't think so, because the same `PostgresChangeFilter` is created in both cases. That class now handles the special formatting for the `inFilter`.

dshukertjr: Oh yeah, now I see it. Great catch! 

Will edit this one.

Vinzent03 (approved):

Issue #668, opened 2023-10-14
Supabase is Stuck in Supabase.initialize() when app is Offline.


SwiftSpotter:
I have used other services like Supabase and they work well even if the user's phone is offline. but with the Supabase app is stuck in the main when my device is offline.

**To Reproduce**
Put your device in offline mode and run the app.

Issue #630, opened 2023-09-18
`Supabase.initialize()` blocks when access token has expired


rkistner:
**Describe the bug**

There are a couple of issues with session management that effectively prevents users from using a Supabase/Flutter-based app while offline, even if all the required data is cached in the app.

1. When launching the app with an expired access token, `Supabase.initialize()` blocks until the session is refreshed. In the best case, this adds a delay to the app startup. If the user is offline, the app hangs completely while launching.
2. The above assuming the default of calling `await Supabase.initialize()` from `main()`. We can't just remove the await - there is no way to check whether or not the user is already logged in until that completes.
3. ~If the user is offline for a while in the app, the access token refresh will fail and is retried every 5 minutes or so. If the user comes online again, the user has to wait for that automatic retry, and all other calls will fail with a 401 error until it completes.~ There is no way to explicitly trigger a retry.

**To Reproduce**
1. Set JWT expiry limit to 90 seconds (unrealistically short, but makes this issue easier to reproduce).
4. Place the device in offline mode.
5. Wait 2 minutes.
6. Kill and restart the app.

**Expected behavior**

1. `Supabase.initialize()` should not block on any network calls. At most, `Supabase.initialize()` should block until the session could be loaded from local storage, even if the access token has expired.
2. ~Calls to Postgrest should automatically refresh the access token if required, instead of waiting for the automatic retry (also see https://github.com/supabase/supabase-flutter/issues/452#issuecomment-1584661212).~
3. An API to explicitly refresh the access token if expired should be added.

**Version (please complete the following information):**

```
└── supabase_flutter 1.10.16
    ├── supabase 1.11.3
    │   ├── functions_client 1.3.2
    │   ├── gotrue 1.12.1
    │   ├── postgrest 1.5.0
    │   ├── realtime_client 1.2.1
    │   ├── storage_client 1.5.2
```

**Workaroud**

As a workaround, setting the JWT expiry limit to the max (1 week) reduces the impact of this issue (limits the occurrence to once a week). But that is quite bad for security - there is then effectively no way to sign users out remotely.

Ideally I'd like to use 5 minute expiry limits, but even with the default of 1 hour this issue comes up quite often.

Vinzent03:
> Calls to Postgrest should automatically refresh the access token if required, instead of waiting for the automatic retry

that should already be the case since [supabase v1.9.8](https://github.com/supabase/supabase-flutter/blob/main/CHANGELOG.md#supabase---v198).

[...]

Issue #662, opened 2023-10-08
[Gotrue] client cannot login due to auth.currentSession being NULL


delfme:
If token is expired, Gotrue doesn't authenticate user (`auth.currentSession` is always returned null )
and issue doesn't resolve with subsequent requests.

I also filed a ticket coz initially I was thinking that the issue was affecting all clients due to maybe Gotrue server being down. Then I tried to run app from my second phone (with a different account) and discovered that issue was affecting only the previous account probably due to not refreshing session after token expiration. 

What is weird is that issue didnt resolve all day long. Cannot understand whether issue is caused by flutter plugin or Gotrue server itself.

Plugin version: ^1.10.21

To fix issue, had to delete account and signup.

dshukertjr:
Are you using email + password sin in? Perhaps is this account that you are experiencing this issue with is a brand new account that you just signed up with?

delfme:
Hello Tyler, thx for the kind attention.

We are using phone otp signup, so no pwds on signins.

Is refresh token managed automatically or must the client listen on auth to refreh token on expirations?
Coz as far as I can see, we are not refreshing token in our code. We just do:

```
static Session? get session => auth.currentSession;
```

As for your guess, indeed lately we also saw this very issue when doing a signup after account deletion, if user signups with the same phone number. No issue if user signups with a different phone number. But we are still investigating this to maybe open a separate issue, if needed. Do you think they might be intertwined?

dshukertjr:
@delfme 
>Is refresh token managed automatically or must the client listen on auth to refreh token on expirations?

Tokens are managed automatically. You do not need to manually refresh it or anything. 

Okay, let's step back a little. Since you skipped the issue template we provided, we don't have the necessary information to look into this issue. Could you provide detailed step-by-step instructions on how to reproduce the issue you are experiencing?

delfme:
Ops, my bad, filed this real quick from phone while travelling.
Will add missing info soon.

delfme:
Unfortunately we don't have a sample code for this and our app code is close.
But here is what we do:
```
...
static GoTrueClient get auth => Supabase.instance.client.auth;
static final authController = StreamController<AuthChangeEvent>.broadcast();
static Session? session;

static Future<void> initialize() async {
    await Supabase.initialize(
      url: _url,
      anonKey: _publicAnonKey,
      debug: kDebugMode, // optional
    );

  //Auth.currentSession is null when issue occurs
[...]

PR #726, opened 2023-11-22
fix!(supabase_flutter, gotrue): Load session from local storage first, and lazily refresh session later.

dshukertjr:

This PR updates the SDK to not hang at `Supabase.initialize()` for token refresh. Instead, it will fetch the session from local storage, and use it as the initial session. 

## New Supabase initialization flow:
1. `Supabase.initialize()` is called
2. session from local storage is obtained
3. gotrue sets the session obtained from local storage as `_currentSession` (even if it had expired) and emits `initialSession` event
4. `auth.recoverSession()` is called, but not awaited, so `Supabase.initialize()` is completed without waiting for network requests

Also in step 4 if `recoverSession()` fails and throws an `AuthException`, such as when the session string was malformed, or the user didn't enable auto refresh token and the access token had expired, gotrue calls `signOut()`, which will trigger the local storage to be deleted by `SupabaseAuth` class. Calling `signOut()` has benefits of triggering `signedOut` event on `onAuthStateChanged`, which is nice, because before this update, the user's session was silently deleted from local storage, and there were no feedback to the developer about the user getting kicked out. 

## Things to note:
- Before this update, `_currentSession` was guaranteed to be fresh right after `Supabase.initialize()` was completed, but that is no longer the case. 
  - There could be a case where the local storage session seemed valid, but the auth server rejected, hence the user gets signed out right away, which feels like the desired behavior, but it is a change from the current one.

## Other

fixes https://github.com/supabase/supabase-flutter/issues/668
fixes https://github.com/supabase/supabase-flutter/issues/630
fixes https://github.com/supabase/supabase-flutter/issues/716
related https://github.com/supabase/supabase-flutter/issues/662


Vinzent03 (approved):

tom-at-pixel:
Does supabase-js have the same behavior as this PR introduces? In other words, does `supabase.auth.initialize()` instantly consume the cached token and then in the background refreshes the token via network? This would be particularly useful on the client (especially for mobile apps). 😊

dshukertjr:
@tom-at-pixel 
JS SDK works in a slightly different way, but creating a Supabase client does not block initializing the app either, so it's good to be used in mobile apps as well!

tom-at-pixel:
@dshukertjr The main problem in the JS SDK is that if there's no internet on app boot, the cached token is not only ignored but also removed, leaving the user in a singed out state. I recognize this is another issue entirely, so I can look for or create a new issue as needed. Thanks for the info!

dshukertjr:
@tom-at-pixel 
Oh, I see. If that is the case, that might be unintended behavior, so if you could open an issue on the gotrue-js repo, that would be awesome!
https://github.com/supabase/gotrue-js

PR #730, opened 2023-11-24
fix!(gotrue): change the default signOut scope to local from global

dshukertjr:

Currently when the developer calls `auth.signOut()`, it signs out the user from every device. This PR changes the default value to `local`, which is probably more aligned with what a developer would expect to happen when calling `auth.signOut()`.

### DIFF
--- a/packages/gotrue/lib/src/constants.dart
+++ b/packages/gotrue/lib/src/constants.dart
@@ -65,11 +65,14 @@ enum OtpChannel {
   whatsapp,
 }
 
-///Determines which sessions should be logged out.
-///
-///[global] means all sessions by this account will be signed out.
-///
-///[local] means only this session will be signed out.
-///
-///[others] means all other sessions except the current one will be signed out. When using others, there is no [AuthChangeEvent.signedOut] event fired on the current session!
-enum SignOutScope { global, local, others }
+/// Determines which sessions should be logged out.
+enum SignOutScope {
+  /// All sessions by this account will be signed out.
+  global,
+
+  /// Only this session will be signed out.
+  local,
+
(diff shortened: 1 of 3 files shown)


DanMossa:
What exactly does global mean and how does it work? Like multiple accounts authed on the same device vs just the current active one?

packages/gotrue/lib/src/constants.dart
```diff
@@ -65,11 +65,14 @@ enum OtpChannel {
...
+enum SignOutScope {
+  /// All sessions by this account will be signed out.
+  global,
+
+  /// Only this session will be signed out.
+  local,
+
+  /// All other sessions except the current one will be signed out. When using others, there is no [AuthChangeEvent.signedOut] event fired on the current session!
+  others,
+}
```
dshukertjr, line 78: @DanMossa 
Good point. Currently it's hard for the developers to understand what the `scope` parameter does. I have updated the comment docs like this to improve the situation. 

If I am signed in on my app from a desktop and my smart phone and I call `signOut(SignOutScope.global)` on my phone, I will also be signed out from the desktop. 

With `local`, I would only be signed out from the phone, and the desktop session will remain active. I think `local` is the behavior that typically developers would expect.

DanMossa: Oh! Yeah local definitely makes more sense then.

But I thought we use JWT tokens, so how would the Supabase goTrue server sign us out on other devices?

DanMossa (approved):

packages/gotrue/lib/src/constants.dart
dshukertjr: @DanMossa 
I think it works by revoking the refresh token, so technically the sessions on other devices will be active until the access token expires.

PR #731, opened 2023-11-24
fix(supabase_flutter): session migration from hive to sharedPreferences now works properly

dshukertjr:

We have a `MigrationLocalStorage` class for anyone migrating from supabase_flutter v1 to v2 so that the auth session would keep working while we have switched from using hive to using sharedPreferences for storing session data. 

There  is currently a bug where the migration fails. This PR fixes the failing migration.

### DIFF
--- a/packages/supabase_flutter/lib/src/local_storage.dart
+++ b/packages/supabase_flutter/lib/src/local_storage.dart
@@ -196,7 +196,9 @@ class MigrationLocalStorage extends LocalStorage {
       final hasHive = await hiveLocalStorage.hasAccessToken();
       if (hasHive) {
         final accessToken = await hiveLocalStorage.accessToken();
-        await sharedPreferencesLocalStorage.persistSession(accessToken!);
+        final session =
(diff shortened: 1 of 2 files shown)


packages/supabase_flutter/lib/src/local_storage.dart
```diff
@@ -196,7 +196,13 @@ class MigrationLocalStorage extends LocalStorage {
       final hasHive = await hiveLocalStorage.hasAccessToken();
       if (hasHive) {
         final accessToken = await hiveLocalStorage.accessToken();
-        await sharedPreferencesLocalStorage.persistSession(accessToken!);
+        final session =
+            Session.fromJson(jsonDecode(accessToken!)['currentSession']);
```
dshukertjr, line 200: Here is the format of the session string stored on local storage for v1. Note that the session object is wrapped in a `currentSession` key. 
https://github.com/supabase/supabase-flutter/blob/main/packages/gotrue/lib/src/types/session.dart#L81

Vinzent03 (approved):

PR #734, opened 2023-11-25
fix(gotrue): signing in with pkce flow fires two `signedIn` auth event

dshukertjr:

Currently when a user signs in with the pkce flow, two `signedIn` events are fired. This PR makes it one.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -641,11 +641,6 @@ class GoTrueClient {
             'No session found for the auth code.');
       }
 
-      if (storeSession == true) {
-        _saveSession(session);
-        notifyAllSubscribers(AuthChangeEvent.signedIn);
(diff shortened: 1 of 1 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -641,11 +641,6 @@ class GoTrueClient {
             'No session found for the auth code.');
       }
 
-      if (storeSession == true) {
-        _saveSession(session);
-        notifyAllSubscribers(AuthChangeEvent.signedIn);
-      }
-
```
dshukertjr, line 648: The same set of code can be found in the `exchangeCodeForSession()` function [here](https://github.com/supabase/supabase-flutter/blob/main/packages/gotrue/lib/src/gotrue_client.dart#L318), so this codeblock is unnecessary.

Vinzent03 (approved):

PR #736, opened 2023-11-27
chore: rename `client` to `supabase`

dshukertjr:

No feature changes. 

This PR renames `client` variable to `supabase` in tests and readme files to align with the docs.


Vinzent03:
I find `Supabase.instance.supabase` a little strange. I would stay with `Supabase.instance.client´. That change is definitely not chore.

Vinzent03 (approved):
I was thinking about whether we could remove the needed `.client` entirely, but it's not really working with `.dispose()`.

PR #739, opened 2023-11-28
fix: emit passwordRecovery event when using pkce

### LINKED ISSUE #664: Missing passwordRecovery AuthChange event when using AuthFlowType.pkce
I can confirm that upon opening the password recovery link, I get `AuthChangeEvent.signedIn` followed by `AuthChangeEvent.passwordRecovery`. This is working as expected, as upon clicking email confirmation, you are signed into the app, hence the initial `signedIn` event, and then followed by the `passwordRecovery` event because the deeplink was for password recovery using ```AuthFlowType authFlowType = AuthFlowType.implicit,```. 

```dart
supabase.auth.onAuthStateChange.listen((data) {
     final event = data.event;
     if(event== AuthChangeEvent.passwordRecovery) {
       // redirect user to update password page
     }
   });
 }
 ```

This recovery event is not fired, when we chose AuthFlowType.pkce. Instead I'm getting two signedIn events. I don't know if this is expected. But it would be nice if we had the recovery password flow implemented in all authFlowTypes.
More precisely the query parameter  `type` is lost after redirect. It is present in the magic link send to the user email. 
```https://myprojectdomain.supabase.co/auth/v1/verify?redirect_to=com.mydomain.myapp://login-callback&token=pkce_37cf6a568461c7ae32dfee4615f30ed1a5a851a9aaa198940c00f1ba&type=recovery```
and this is how it looks in _handleDeepLink function
```com.mydomain.myapp://login-callback?code=acabff8e-a9bd-4b4f-9597-4b92ef112393```

then ```getSessionFromUrl``` is called and it finishes the execution with return ```return AuthSessionUrlResponse(session: session, redirectType: null);```

The parsing of the get params type is happening outside of this `if` and is never executed for pkce
```dart
if (_flowType == AuthFlowType.pkce) {
  /// some code 
  if (storeSession == true) {
    _saveSession(session);
    _notifyAllSubscribers(AuthChangeEvent.signedIn); 
  }
  return AuthSessionUrlResponse(session: session, redirectType: null);
}
final redirectType = url.queryParameters['type'];

if (storeSession == true) {
  _saveSession(session);
  if (redirectType == 'recovery') {
    _notifyAllSubscribers(AuthChangeEvent.passwordRecovery);
  } else {
    _notifyAllSubscribers(AuthChangeEvent.signedIn);
  }
}
```

Another issue is that handleDeepLink function swallows the AuthException. I would expect to be able to capture some error and show to the user a message that his magic link has expired 

```dart
try {
      await Supabase.instance.client.auth.getSessionFromUrl(uri);
    } on AuthException catch (error, stackTrace) {
      Supabase.instance.log(error.toString(), stackTrace);
    } catch (error, stackTrace) {
      Supabase.instance.log(error.toString(), stackTrace);
    }
```

**To Reproduce**
1. Initialize Supabase

```dart 
await Supabase.initialize(
    url: dotenv.get('SUPABASE_URL'),
    anonKey: dotenv.get('ANON_KEY'),
    authCallbackUrlHostname: 'login-callback',
    debug: true,
  );
  ```
  
2. Invoke Supabase auth method

```dart 
_auth.resetPasswordForEmail(
        email,
        redirectTo: redirectTo,
        captchaToken: captchaToken,
      )
```
      

3. Listen to authStateChanges

```dart
supabaseClient.auth.onAuthStateChange.listen(
      (event) {
        log('${event.event}');
        log('${event.session}');
     
        if (event.event == AuthChangeEvent.passwordRecovery) {
          log('THIS IS RECOVERY');
        }
      },
      onError: (dynamic error) {
        log('app log: $error');
        _supabaseClient.auth.signOut();
   
      },
    );
```
   
 
4. Observe logs and place breakpoints in [getSessionFromUrl](https://github.com/supabase/supabase-flutter/blob/4e5df1fefdd3888fe7055038817fa21dd880db62/packages/gotrue/lib/src/gotrue_client.dart#L629C34-L629C51)

**Expected behavior**
AuthChangeEvent.passwordRecovery to be fired when using AuthFlowType.pkce

On Windows
supabase_flutter 1.9.0
supabase 1.8.0
functions_client 1.2.0
gotrue 1.7.0
postgrest 1.2.3
realtime_client 1.0.3
storage_client 1.3.0

[√] Flutter (Channel stable, 3.7.12, on Microsoft Windows [Version 10.0.19045.3448], locale bg-BG)
[√] Windows Version (Installed version of Windows is version 10 or higher)
[√] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
[√] Chrome - develop for the web
[√] Visual Studio - develop for Windows (Visual Studio Professional 2022 17.5.4)
[√] Android Studio (version 2022.3)
[√] VS Code (version 1.83.0)
[√] Connected device (4 available)
[√] HTTP Host Availability

The same issue exists in the latest version of go_true client, just checked. The pkce flow ends the logic prematurely without checking the type query parameter.

Vinzent03:

Bug fix

When using pkce no password recovery event is send. Instead just a sign in event is send.

## What is the new behavior?

Send a password recovery event when needed

## Additional context

Note that I haven't had the time to actual test this. I don't even know if the url from the redirect contains the `type` query parameter, but I guess so.

Will try to add tests.

close #664

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -312,12 +312,6 @@ class GoTrueClient {
 
     final authResponse = AuthResponse.fromJson(response);
 
-    final session = authResponse.session;
-    if (session != null) {
-      _saveSession(session);
-      notifyAllSubscribers(AuthChangeEvent.signedIn);
-    }
-
     return authResponse;
   }
 
@@ -630,6 +624,8 @@ class GoTrueClient {
   }) async {
     if (_flowType == AuthFlowType.pkce) {
       final authCode = originUrl.queryParameters['code'];
+      final redirectType = originUrl.queryParameters['type'];
+
       if (authCode == null) {
         throw AuthPKCEGrantCodeExchangeError(
             'No code detected in query parameters.');
@@ -639,6 +635,15 @@ class GoTrueClient {
       if (session == null) {
         throw AuthPKCEGrantCodeExchangeError(
             'No session found for the auth code.');
+      }
+
+      if (storeSession == true) {
+        _saveSession(session);
+        if (redirectType == 'recovery') {
+          notifyAllSubscribers(AuthChangeEvent.passwordRecovery);
+        } else {
+          notifyAllSubscribers(AuthChangeEvent.signedIn);
+        }
       }
 
       return AuthSessionUrlResponse(session: session, redirectType: null);


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -630,6 +624,8 @@ class GoTrueClient {
   }) async {
     if (_flowType == AuthFlowType.pkce) {
       final authCode = originUrl.queryParameters['code'];
+      final redirectType = originUrl.queryParameters['type'];
```
dshukertjr, line 627: Actually, the `type` query parameter does not exist, and the only parameter we have is `code`. 

I think fixing this issue would require saving additional data in the local storage. Let's wait to see how js lib fixes it and match what they do. 
https://github.com/supabase/gotrue-js/pull/813

Vinzent03:
Done in #744

PR #744, opened 2023-11-30
fix: PKCE flow not emitting password recovery event

dshukertjr:

Currently with PKCE flow, `passwordRecovery` auth event is not emitted for email password reset flow. This PR fixes it.

`signIn` event is emitted after clicking on the link sent to the email.

## What is the new behavior?

`passwordRecovery` event is emitted after opening the link sent to the email.

## Additional context

Fixes https://github.com/supabase/supabase-flutter/issues/664

### DIFF
--- a/packages/gotrue/lib/src/constants.dart
+++ b/packages/gotrue/lib/src/constants.dart
@@ -28,13 +28,13 @@ enum AuthChangeEvent {
 }
 
 extension AuthChangeEventExtended on AuthChangeEvent {
-  static GenerateLinkType fromString(String? val) {
-    for (final event in GenerateLinkType.values) {
+  static AuthChangeEvent fromString(String? val) {
+    for (final event in AuthChangeEvent.values) {
       if (event.name == val) {
         return event;
       }
     }
-    return GenerateLinkType.unknown;
+    return AuthChangeEvent.signedIn;
   }
 }
 
--- a/packages/gotrue/lib/src/constants.dart
+++ b/packages/gotrue/lib/src/constants.dart
@@ -27,6 +27,17 @@ enum AuthChangeEvent {
   mfaChallengeVerified,
 }
 
+extension AuthChangeEventExtended on AuthChangeEvent {
+  static GenerateLinkType fromString(String? val) {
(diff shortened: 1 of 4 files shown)


packages/gotrue/lib/src/constants.dart
```diff
@@ -27,6 +27,17 @@ enum AuthChangeEvent {
...
 }
 
+extension AuthChangeEventExtended on AuthChangeEvent {
+  static AuthChangeEvent fromString(String? val) {
+    for (final event in AuthChangeEvent.values) {
+      if (event.name == val) {
+        return event;
+      }
+    }
+    return AuthChangeEvent.signedIn;
```
Vinzent03, line 37: I think it's strange to return `signedIn` as default case. Instead, I would simply return `null`. I know this doesn't really matter at the moment, because that method is only used once. But for future uses I think it's easier to understand to return `null` and let the method caller decide what's the default case should be.

dshukertjr: Sounds good. Updated!

Vinzent03:
`getSessionFromUrl` still returns `redirectType: null` when using pkce despite that we now have the information. I think we should use `AuthSessionUrlResponse` as return type in `exchangeCodeForSession`, so that we can return the redirectType in `getSessionFromUrl`.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -310,15 +316,18 @@ class GoTrueClient {
     await _asyncStorage!
         .removeItem(key: '${Constants.defaultStorageKey}-code-verifier');
 
-    final authResponse = AuthResponse.fromJson(response);
+    final authSessionUrlResponse = AuthSessionUrlResponse(
+        session: Session.fromJson(response)!, redirectType: redirectType?.name);
 
-    final session = authResponse.session;
-    if (session != null) {
```
dshukertjr, line 316: Got rid of this check for session. 

The signature of the return type on the js SDK guarantees that there is a session, so not sure why the js SDK has the check for session there. 
https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L534

PR #760, opened 2023-12-13
feat(gotrue, supabase_flutter): Add identity linking and unlinking methods.

dshukertjr:

This PR brings one of the core auth launch announced at LWX.

Add a method to retrieve URL for manual linking OAuth providers, and adds a method to launch the URL obtained from Gotrue to supabase_flutter.

Manual identity linking is not possible using supabase_flutter.

## What is the new behavior?

Manual identity linking is now poissible.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -800,6 +800,7 @@ class GoTrueClient {
       redirectTo: redirectTo,
       scopes: scopes,
       queryParams: queryParams,
+      skipBrowserRedirect: true,
     );
     final res = await _fetch.request(urlResponse.url!, RequestMethodType.get,
         options: GotrueRequestOptions(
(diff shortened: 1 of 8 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -792,6 +793,49 @@ class GoTrueClient {
     );
   }
 
+  /// Gets all the identities linked to a user.
+  Future<List<UserIdentity>> getUserIdentities() async {
+    final res = await refreshSession();
```
Vinzent03, line 798: I think the new `getUser()` method is more appropriate, because it doesn't create a new session, but gets the latest user updates.

dshukertjr: Great call!

dshukertjr: Updated!

Vinzent03 (changes requested):
This hasn't to do with this pr specifically, but I think the `url` field from `OAuthResponse` can be `String` and not `String?`

dshukertjr:
@Vinzent03 Done!

Vinzent03:
The tests are failing now.l, because of unnecessary `!`.

dshukertjr:
Thanks! Should've caught that one 😂

Vinzent03 (approved):

PR #761, opened 2023-12-13
feat: add getUser() method

### LINKED ISSUE #752: Supabase auth admin missing proper getUser function
The NodeJS version has an admin function called getUser, this function accepts a custom token and practically allows verifying the user identity when handling a request in a custom server environment. This function seems to be missing in the current Dart package.

**Describe the solution you'd like**
A function with the same functionality being added to the package

**Describe alternatives you've considered**
At least describe the API request needed to get the same information (I tried reading the source code and trying to replicate the desired API request myself, but I couldn't find all the needed info)

**Additional context**
Example use case: There's a custom-made Dart TCP server that needs to be integrated into the project. The server first needs to verify the identity of the connecting socket and drop the connection if it's not authentic. It gets a special auth request package, parses it, gets the user access token sent by them, passes this token to a supabase.auth.admin.getUser(token) function and in return either gets a valid user object or null, indicating that the verification has failed.

Edit: added example use case

Vinzent03:

feature

There's no `getUser` method to get a user object from a jwt.

## What is the new behavior?

Add `getUser`

## Additional context

close #752

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -596,6 +596,23 @@ class GoTrueClient {
     }
   }
 
+  /// Gets the current user details from current session or custom [jwt]
+  Future<UserResponse> getUser(String? jwt) async {
+    if (jwt == null && currentSession?.accessToken == null) {
+      throw AuthException('Cannot get user: no current session.');
+    }
+    final options = GotrueRequestOptions(
+      headers: _headers,
+      jwt: jwt ?? currentSession?.accessToken,
+    );
+    final response = await _fetch.request(
+      '$_url/user',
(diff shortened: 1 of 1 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -596,6 +596,23 @@ class GoTrueClient {
     }
   }
 
+  /// Gets the current user details from current session or custom [jwt]
+  Future<UserResponse> getUser(String? jwt) async {
```
dshukertjr, line 600: We probably want to make the parameter optional I think.
```suggestion
  Future<UserResponse> getUser([String? jwt]) async {
```

dshukertjr (approved):
Thanks for this! This is great!

PR #768, opened 2023-12-17
fix: enable filtering and tranformation on count with head

dshukertjr:

Currently when performing count with head: true, we cannot chain filters (eq, match, etc.). 

This PR will allow filters to be chained on `.count()` like this:

```dart
final int count = await supabase.from('table')
  .count() // count with head: true
  .eq('name', 'Supabase');
```

## Cons

Might not be anything to be too concerned about, but wanted to leave a note here. The change in this PR will actually allow a whole bunch of other things to be chained to `.count()` that doesn't make a whole lot of sense. 

```dart
final res = await supabase.from('table')
  .count()
  .eq('name', 'Supabase')
  .limit(1) // It's a head request, and limits and ranges are ignored for count request, so this is not doing anything
  .withConverter<List<int>>((data) => [data]); // `data` here is the count value of type int. Again, doesn't really make a whole lot of sense being able to chain this here.
```

### DIFF
--- a/packages/postgrest/lib/src/postgrest_query_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_query_builder.dart
@@ -250,11 +250,10 @@ class PostgrestQueryBuilder<T> extends RawPostgrestBuilder<T, T, T> {
   /// ```dart
   /// int count = await supabase.from('users').count();
   /// ```
-  RawPostgrestBuilder<int, int, int> count(
-      [CountOption option = CountOption.exact]) {
-    return _copyWithType(
+  PostgrestFilterBuilder<int> count([CountOption option = CountOption.exact]) {
+    return PostgrestFilterBuilder<int>(_copyWithType(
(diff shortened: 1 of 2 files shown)


Vinzent03:
I'm pretty sure I made this by design. The filters should be added before the `count()` method.

dshukertjr:
@Vinzent03 
Do you mean something like this?

```dart
final count = await supabase.from('table').eq('name', 'supabase').count();
```

dshukertjr:
Because currently it is not possible to chain filters and transformers directly to `.from()`. I personally think 
```dart
supabase.from('table').count().eq('name', 'supabase'); 
```
 is a better API than 
```dart
supabase.from('table').eq('name', 'supabase').count();
```

for count with head:true.

Vinzent03:
Hmm okay. I agree with you that count should come first.

Vinzent03 (approved):

PR #774, opened 2023-12-20
fix(gotrue): Fix the issue where `verfiyOTP` emits `signIn` instead of `passwordRecovery` auth event.

### LINKED ISSUE #773: verifyOTP should emit AuthChangeEvent.passwordRecovery instead of AuthChangeEvent.signedIn when OtpType.recovery is passed
**Describe the bug**
When a user requests a password reset, they receive an email containing a magiclink and an OTP code. As far as I can tell, if they choose to use the code, the app must call the `verifyOTP` method with `OtpType.recovery`. Verification currently succeeds but then an `AuthChangeEvent.signedIn` event is emitted to all event listeners and then the user gets signed in rather than taken to the password reset screen. Under normal operation, an `AuthChangeEvent.passwordRecovery` ought to be emitted instead (which is the behavior that the magic link currently has).

The offending line of code appears to be here: https://github.com/supabase/supabase-flutter/blob/main/packages/gotrue/lib/src/gotrue_client.dart#L512

I propose we add a check that changes the event depending on the OtpType

**To Reproduce**
1. Subscribe to auth events via `onAuthStateChange`
2. call `resetPasswordForEmail()`
3. retrieve email OTP
4. call `verifyOTP()` with email and `OtpType.recovery`
5. Observe that `AuthChangeEvent.signedIn` is emitted to listeners instead of `AuthChangeEvent.passwordRecovery`

**Expected behavior**
When using the magiclink from the email, the user gets signed in but the emitted event is `AuthChangeEvent.passwordRecovery`. Both magiclinks and OTP should behave the same way once a user has been verified / authenticated.

macOS 14.2 (23C64)
```
├── supabase_flutter 2.0.1
│   ├── supabase 2.0.1
│   │   ├── functions_client 2.0.0
│   │   ├── gotrue 2.1.0
│   │   ├── postgrest 2.0.0
│   │   ├── realtime_client 2.0.0
│   │   ├── storage_client 2.0.0
```
```
Flutter 3.16.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 2e9cb0aa71 (9 days ago) • 2023-12-11 14:35:13 -0700
Engine • revision 54a7145303
Tools • Dart 3.2.3 • DevTools 2.28.4
```

point-source:

This fixes issue #773

`verifyOTP` emits `AuthChangeEvent.signedIn` when `OtpType.recovery` is passed

## What is the new behavior?

`verifyOTP` emits `AuthChangeEvent.passwordRecovery` when `OtpType.recovery` is passed

This allows listeners to properly navigate the user to the password reset screen rather than the default home screen after verification.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -509,7 +509,9 @@ class GoTrueClient {
     }
 
     _saveSession(authResponse.session!);
-    notifyAllSubscribers(AuthChangeEvent.signedIn);
+    notifyAllSubscribers(type == OtpType.recovery
+        ? AuthChangeEvent.passwordRecovery
+        : AuthChangeEvent.signedIn);
 
     return authResponse;
   }


dshukertjr (approved):
Legend 🚀

lordvins226:
Thanks for this🥳

PR #789, opened 2024-01-05
forward id from https://github.com/supabase/storage/pull/332

HannesGitH:

patch

when uploading a file a key as String is returned, the id of the new object isn't

## What is the new behavior?

the new return type includes both the key and the id

## Additional context

Simply fowards the changes introduced in https://github.com/supabase/storage/pull/332 to the dart/flutter client

### DIFF
--- a/packages/storage_client/CHANGELOG.md
+++ b/packages/storage_client/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.0.1
+
+> Note: This release has breaking changes.
+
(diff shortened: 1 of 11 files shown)


dshukertjr:
Thanks for this PR!

Unfortunately, we cannot do a breaking change until the next major release, v3, so we are going to have to wait until then.

HannesGitH:
Understood, reasonable, thanks for replying!
Is there anywhere I can propose this change for the next major tho?

dshukertjr:
I've left a note [here](https://github.com/supabase/supabase-flutter/issues/790)!

PR #791, opened 2024-01-07
feat: Add linkedin_oidc OAuthProvider

### LINKED ISSUE #779: linkedin_oidc is not a supported provider for the signInWithOAuth method
When using the signInWithOAuth with linkedin_oidc as the provider as this not a valid enum (only linkedin is defined)
 
the following error is returned:

Error: Member not found: 'linkedin_oidc'.
    supabase_flutter.Provider.linkedin_oidc, 

Per this documentation https://supabase.com/docs/guides/auth/social-login/auth-linkedin?language=js

When your user signs in, call [signInWithOAuth()](https://supabase.com/docs/reference/javascript/auth-signinwithoauth) with linkedin_oidc as the provider:

Vinzent03:

feature

## What is the new behavior?

Adds the `linkedin_oidc` OAuthProvider.

## Additional context
- [js pr](https://github.com/supabase/gotrue-js/pull/796)
- close #779

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -366,7 +366,7 @@ class GoTrueClient {
       options: GotrueRequestOptions(
         headers: _headers,
         body: {
-          'provider': provider.name,
+          'provider': provider.snakeCase,
           'id_token': idToken,
           'nonce': nonce,
           'gotrue_meta_security': {'captcha_token': captchaToken},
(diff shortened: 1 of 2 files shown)


dshukertjr (commented):
Until now, when js SDK has a snake case enum value, we were converting them to camel case. I think we should keep the same pattern for consistency here as well. Would you be able to update the code so that this would work with camel case?

dshukertjr (approved):

Issue #759, opened 2023-12-13
Add support for `signInWithSSO` for Connection to Identity Providers (IdPs) connected via SAML


encima:
**Is your feature request related to a problem? Please describe.**
Sign in with SSO is currently supported in the JS client and is a key function for allowing users to sign in through external Identity providers.

**Describe the solution you'd like**
Implementation of `signinWithSSO` to allow for login through Identity providers and to bring the flutter package in line with other client libs (as well as the hosted platform).

**Describe alternatives you've considered**
Not relevant

**Additional context**

PR #798, opened 2024-01-17
feat(gotrue, supabase_flutter): Add `signInWithSSO` method

dshukertjr:

Add `signInWithSSO()` method that can be called like the following:

```dart
await supabase.auth.signInWithSSO(
  domain: `company.com`
);
```

`signInWithSSO()` works in a similar way as `signInWithOAuth()` where `gotrue-dart` has a `getSSOSignInUrl()` that provides the sign in URL, and supabase_flutter handles the user redirect.

supabase-js implementation here: https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L758

Closes https://github.com/supabase/supabase-flutter/issues/759

### DIFF
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -294,6 +294,42 @@ extension GoTrueClientSignInProvider on GoTrueClient {
     return result;
   }
 
+  /// Attempts a single-sign on using an enterprise Identity Provider. A
+  /// successful SSO attempt will redirect the current page to the identity
+  /// provider authorization page. The redirect URL is implementation and SSO
+  /// protocol specific.
+  ///
+  /// You can use it by providing a SSO domain. Typically you can extract this
+  /// domain by asking users for their email address. If this domain is
+  /// registered on the Auth instance the redirect will use that organization's
+  /// currently active SSO Identity Provider for the login.
+  ///
+  /// If you have built an organization-specific login page, you can use the
+  /// organization's SSO Identity Provider UUID directly instead.
+  ///
+  /// Returns true if the URL was launched successful, otherwise either returns
+  /// false or throws a [PlatformException] depending on the launchUrl failure.
+  ///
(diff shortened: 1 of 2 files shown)


packages/supabase_flutter/lib/src/supabase_auth.dart
```diff
@@ -294,6 +294,42 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
+  ///
+  /// You can use it by providing a SSO domain. Typically you can extract this
+  /// domain by asking users for their email address. If this domain is
+  /// registered on the Auth instance the redirect will use that organization's
+  /// currently active SSO Identity Provider for the login.
+  ///
+  /// If you have built an organization-specific login page, you can use the
+  /// organization's SSO Identity Provider UUID directly instead.
+  ///
+  /// Returns true if the URL was launched successful, otherwise either returns
```
Vinzent03, line 310:
```suggestion
  /// Returns true if the URL was launched successfully, otherwise either returns
```

packages/supabase_flutter/lib/src/supabase_auth.dart
```diff
@@ -294,6 +294,42 @@ extension GoTrueClientSignInProvider on GoTrueClient {
...
+  /// );
+  /// ```
+  Future<bool> signInWithSSO({
+    String? providerId,
+    String? domain,
+    String? redirectTo,
+    String? captchaToken,
+    LaunchMode launchMode = LaunchMode.platformDefault,
+  }) async {
+    final ssoUrl = await getSSOSignInUrl();
```
Vinzent03, line 325: The parameters aren't passed on to the actual call.

Vinzent03 (changes requested):

packages/supabase_flutter/lib/src/supabase_auth.dart
dshukertjr: 😂

Vinzent03 (approved):

Issue #794, opened 2024-01-10
Supabase stops initialization on safari after page refresh


kaliosek:
**Describe the bug**
On flutter web on safari flutter app doesn't finish  await Supabase.initialize if page is refreshed.
In log I have: 
"***** Supabase init completed Instance of 'Supabase'"
"Creating objectStore box in database supabase_authentication..."
and stops here.

This works well on Chrome or osx app.

If I delete app data from browser app run with no problem. but one or many refreshes stops app.

**To Reproduce**
1. Create new flutter project
2. Add supabase_flutter dependency
3. add initialisation code:
`Future<void> main() async {
  WidgetsBinding widgetsBinding = WidgetsFlutterBinding.ensureInitialized();
  await Supabase.initialize(
    url: 'correct url',
    anonKey:' correct anon key',
  );
  runApp(const MyApp());
}`
4. Build and run default app in safari.
5. If page is loaded correctly, refresh page.

**Expected behavior**
Supabase should be initialised even if data in local storage exist.

**Screenshots**
app has white screen

**Version (please complete the following information):**
On Linux/macOS
Please run `dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"` in your project directory and paste the output here.
supabase_safari_localstorage_error 1.0.0+1
└── supabase_flutter 2.2.0
    ├── supabase 2.0.4
    │   ├── functions_client 2.0.0
    │   ├── gotrue 2.3.0
    │   ├── postgrest 2.1.0
    │   ├── realtime_client 2.0.0
    │   ├── storage_client 2.0.0

On Windows
Please run `dart pub deps | findstr "supabase gotrue postgrest storage_client realtime_client functions_client"` in your project directory and paste the output here.

**Additional context**
I think this is related with use SharedPreferences on web. Before 2.0.0 it works well.
This behaviour is always the same if I push new version of app to server.

dshukertjr:
I am unable to reproduce the issue with the steps you provided, and works just fine for me. Are there any extra steps that you haven't mentioned like authenticating or something?

kaliosek:
Hi I can reproduce this in every case :(
I attached video that shows this problem:

https://github.com/supabase/supabase-flutter/assets/67379323/f5c386ca-7f52-41c8-9bb0-e2fe3942306c

dshukertjr:
@kaliosek 
[...]

PR #799, opened 2024-01-17
fix(supabase_flutter): Add timeout to Hive.openBox to fix hanging issue

dshukertjr:

Currently when initializing Supabase within Safari while dev tool is opened, `Supabase.initialize()` hangs leaving the user with just a white screen.

This is caused by `await Hive.openBox(_hiveBoxName, encryptionCipher: encryptionCipher);` hanging. Although the reason for why the function does not complete is unknown, we could at least add a timeout so that when it does hang, we can at least catch the `TimeoutException` and proceed. 

Fixes https://github.com/supabase/supabase-flutter/issues/794

### DIFF
--- a/packages/supabase_flutter/lib/src/local_storage.dart
+++ b/packages/supabase_flutter/lib/src/local_storage.dart
@@ -87,7 +87,8 @@ class HiveLocalStorage extends LocalStorage {
       encryptionCipher = HiveAesCipher(base64Url.decode(encryptionKey!));
     }
     await Hive.initFlutter('auth');
-    await Hive.openBox(_hiveBoxName, encryptionCipher: encryptionCipher);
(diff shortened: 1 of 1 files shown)


packages/supabase_flutter/lib/src/local_storage.dart
```diff
@@ -185,7 +186,13 @@ class MigrationLocalStorage extends LocalStorage {
     await Hive.initFlutter('auth');
     hiveLocalStorage = const HiveLocalStorage();
     await sharedPreferencesLocalStorage.initialize();
-    await migrate();
+    try {
+      await migrate();
+    } on TimeoutException {
+      print('TimeoutException');
```
dshukertjr, line 192:
```suggestion
```

Issue #805, opened 2024-01-22
Update README and docs to explain JWT refresh flow


ellmaki:
Right now it is very confusing how to handle the JWT refresh flow for supabase with flutter. It is my understanding that the supabase package is supposed to generate a new access token automatically, but this is not a clear flow.

For example: are currentUser and currentSession reliably not-null anytime a user is authenticated? I have run into situations where they appear temporarily null inspite of the user being logged in. This has broken key parts of my app UI.

Is there any code needed on the client side to handle JWT refresh?

Are we supposed to listen to onAuthStateChanged to handle the JWT refresh?

Really a simple guide for Supabase and Flutter using something like Provider around persisting auth state and handling the JWT refresh would be immensely appreciated. As it stands this flow is very hard to figure out.

ellmaki:
For example I see this code 
```
supabase.auth.onAuthStateChanged
.listen((data) {})
.onError((error) {
  supabase.auth.refreshSession();  
});```
from another issue: https://github.com/supabase/supabase-flutter/issues/452#issuecomment-1588906208

Is this the recommended way to handle refresh?

ellmaki:
I am also not relying on the session state for any of my UI since I have found it to be unreliable. I am relying only on my user state Provider for that. Should we be able to depend on session state for UI? If it is true that we cannot this is something else which I think should be really explicitly documented.

dshukertjr:
@ellmaki 
supabase_flutter will automatically refresh your access token, and there is nothing you need to do. The SDK will automatically update the token if the session has expired when making requests. You shouldn't ever have to call `refreshSession()` either. 

Would you explain why you thought you had to implement the complicated workarounds you have shared?

dshukertjr:
>
I am also not relying on the session state for any of my UI since I have found it to be unreliable. I am relying only on my user state Provider for that. Should we be able to depend on session state for UI? If it is true that we cannot this is something else which I think should be really explicitly documented.

Yes, you can rely on the Supabase auth state in your UI. If `supabase.auth.currentSession` is not null, the user is signed in.

ellmaki:
I think this was possibly the source of all of my problems https://github.com/supabase/supabase-flutter/pull/806

dshukertjr:
That sounds good, and sorry for the confusion.

ellmaki:
Thanks for addressing it quickly.

ellmaki:
[...]

PR #806, opened 2024-01-22
fix(gotrue): Set _currentUser when setting initial session

dshukertjr:

Currently when the supabase_flutter SDK obtains the initial session from local storage and calls `setInitialSession()` method of gotrue, the `_currentUser` is not set causing a state where `auth.currentSession` is not null, but `auth.currentUser` being null. This PR fixes it. 

Related https://github.com/supabase/supabase-flutter/issues/805

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -901,6 +901,7 @@ class GoTrueClient {
     }
 
     _currentSession = session;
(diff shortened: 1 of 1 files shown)


ellmaki:
Realizing the issue probably is that I was relying on `currentUser` and not `currentSession`. This is honestly probably the root of all my issues.

Vinzent03 (approved):

PR #813, opened 2024-01-27
feat: Add token hash to verifyOtp

### LINKED ISSUE #803: Verify OTP support to hashed token
I'm working on a invite users flow and I'm missing the OTP verification using just the hashed token sent on the email.
The JS library has this implementation:

`const { data, error } = await supabase.auth.verifyOtp({ token_hash: tokenHash, type: 'email'})`

There is a reason for the dart implementation to drop this support? Also the docs for this method only mention SMS and mobile, which may be misleading.

Vinzent03:

feature

You can't verify a token hash

## Additional context

close #803


dshukertjr (approved):
🚀

PR #815, opened 2024-01-30
fix: Remove platform check to start deep link observer on every platform including Linux

LucasXu0:

- Support deeplink on Linux platform

## Additional context

The app_link package has supported the deep link feature since version 3.5.0.

https://github.com/llfbandit/app_links/pull/76


dshukertjr (approved):
Thanks for this. I had missed these points with this PR. https://github.com/supabase/supabase-flutter/pull/792

PR #824, opened 2024-02-04
docs(postgrest): Expand documentation for `contains` and `containedBy` methods

jocubeit:

Previous documentation provides code examples that are syntactically incorrect and do not explain how to provide inclusive and exclusive ranges as a value.

This PR does not introduce any code changes, only code documentation changes.

## Additional context

Documentation in VS Code for `contains` and `containedBy` methods was confusing, and looked like an obvious typo. While performing a deeper dive into the methods themselves I discovered I could pass a string with parenthesis or brackets, or an array object as a value. This was not exposed in the code comments for the documentation. I felt compelled to update it.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_filter_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_filter_builder.dart
@@ -248,11 +248,25 @@ class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
 
   /// Finds all rows whose json, array, or range value on the stated [column] contains the values specified in [value].
   ///
+  /// Pass an array or use brackets in a string for an inclusive range:
   /// ```dart
   /// await supabase
   ///     .from('users')
   ///     .select()
-  ///     .contains('age_range', '[1,2)');
+  ///     .contains('age_range', [1,2]);
+  /// 
+  /// await supabase
+  ///     .from('users')
+  ///     .select()
+  ///     .contains('age_range', '[1,2]');  
+  /// ```
+  ///
+  /// Use parenthesis in a string for an exclusive range:
+  /// ```dart
+  /// await supabase
+  ///     .from('users')
+  ///     .select()
+  ///     .contains('age_range', '(1,2)');
   /// ```
   PostgrestFilterBuilder<T> contains(String column, Object value) {
     final Uri url;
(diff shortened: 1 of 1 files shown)


packages/postgrest/lib/src/postgrest_filter_builder.dart
```diff
@@ -248,11 +248,25 @@ class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
...
   /// await supabase
   ///     .from('users')
   ///     .select()
-  ///     .contains('age_range', '[1,2)');
+  ///     .contains('age_range', [1,2]);
+  /// 
+  /// await supabase
+  ///     .from('users')
+  ///     .select()
+  ///     .contains('age_range', '[1,2]');
```
dshukertjr, line 261: I'm guessing this chunk is a typo since we have the same thing just above.

packages/postgrest/lib/src/postgrest_filter_builder.dart
```diff
@@ -272,11 +286,25 @@ class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
...
+  /// ```dart
+  /// await supabase
+  ///     .from('users')
+  ///     .select()
+  ///     .containedBy('age_range', [1,2]);
+  /// 
+  /// await supabase
+  ///     .from('users')
+  ///     .select()
+  ///     .containedBy('age_range', '[1,2]');
```
dshukertjr, line 299: And this one seems like a duplicate as well.

dshukertjr (commented):
Thanks for this. This PR is great. Just to take this PR a step further, maybe we can address the different data types that `contains` and `containedBy` can be used against. Both of these filters can be used against an array and range types (and json types with array). Maybe it's better to explain this and 

The JS docs explains it well, and yes, will bring this over to the Flutter docs as well.
https://supabase.com/docs/reference/javascript/containedby?example=on-jsonb-columns

I think we can add an example in each method to show how to use them with array columns. 

And then for the range example, I actually like the `[1,2)` sample that we currently have, because it shows that you can use inclusive and exclusive range in a single call. Maybe we can just say something like use "use brackets for inclusive range and parenthesis for exclusive range", and just have the current example that we have.

dshukertjr:
I have updated the docs on `contains` and `containedBy` with more intuitive examples. It would be great if we could bring these examples over to the comment docs on supabase_flutter if you're interested!

https://supabase.com/docs/reference/dart/contains

dshukertjr (approved):
I made some edits based on the examples provided in the [docs](https://supabase.com/docs/reference/dart/contains). Thanks for the PR!

PR #825, opened 2024-02-06
fix: Correct the class name used for custom storage

dshukertjr:


Vinzent03 (changes requested):
Now that I see it, I wouldn't call it `MockLocalStorage`, because it doesn't mock, but uses secure local storage.

dshukertjr:
Sure, updated!

bdlukaa (approved):

Vinzent03 (approved):

PR #832, opened 2024-02-07
fix: Add await to setInitialSession during initialization

### LINKED ISSUE #831: setInitialSession not awaited
Shouldn't this call to setInitialSession be awaited?

https://github.com/supabase/supabase-flutter/blob/a72166268facd97f07cf9588415a2f3d31f68c22/packages/supabase_flutter/lib/src/supabase_auth.dart#L57

whymatter:
Fixes #831

In case there are no reasons why it shouldn't be awaited, I think we should await it.


Vinzent03:
Yeah when the device is offline the `signOut()` call in `setInitialSession()` could take a long time, which delays the overall initialization time. Though I see that the `signOut` may produce an uncaught exception. I'm going to fix this in another pr.

Issue #837, opened 2024-02-14
Remove eventsPerSecond Client Throttling


w3b6x9:
[@supabase/realtime-js](https://github.com/supabase/realtime-js) has removed `eventsPerSecond` client throttling so this client lib should follow suit.

See: https://github.com/supabase/realtime-js/pull/260

w3b6x9:
@dshukertjr just making sure you're aware.

PR #838, opened 2024-02-15
refactor: Deprecate `eventsPerSecond` on Realtime

dshukertjr:

Client-side Realtime rate-limiting was causing confusion among developers when they saw that some of their Realtime events were not being reflected. After some discussions by the Realtime team, they concluded that the client side rate limiting behavior is not ideal, and decided to drop it. There is a server side rate limiting feature in place.

This PR deprecates the `eventsPerSecond` and removes client-side rate limiting.

JS implementation: https://github.com/supabase/realtime-js/pull/260
Closes https://github.com/supabase/supabase-flutter/issues/837

### DIFF
--- a/packages/realtime_client/lib/src/types.dart
+++ b/packages/realtime_client/lib/src/types.dart
@@ -101,7 +101,14 @@ class ChannelFilter {
   }
 }
 
-enum ChannelResponse { ok, timedOut, error }
+enum ChannelResponse {
+  ok,
+  timedOut,
+  @Deprecated(
+      'Client side rate limiting has been removed, and this enum value will never be returned.')
+  rateLimited,
+  error
+}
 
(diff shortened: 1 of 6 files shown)


packages/realtime_client/lib/src/types.dart
```diff
@@ -101,7 +101,7 @@ class ChannelFilter {
   }
 }
 
-enum ChannelResponse { ok, timedOut, rateLimited, error }
```
Vinzent03, line 104: I wouldn't remove the enum value as this an exposed type, which may cause compile time error now. Better to mark it as deprecated, like the config option.

Vinzent03 (changes requested):

packages/realtime_client/lib/src/types.dart
dshukertjr: Good call. Fixed it!

Issue #819, opened 2024-01-30
Issue with redirect back to app after loggining using signInWithOAuth with provider Spotify on IOS


benjaminmountain:
I am running into an issue with redirecting back to my app after logging in with Spotify using the signInWithOAuth method on IOS. I am able to get to the spotify login screen and login but then nothing happens. If I close the in app browser and trigger the login again the in app browser is opened and stops on a blank white screen like below.

![Simulator Screenshot - iPhone 15 - 2024-01-30 at 17 27 59](https://github.com/supabase/supabase-flutter/assets/49491200/62133b54-db6e-4979-8231-c03252ffe996)

I have added my Supabase Callback URL to my Spotify app.

I have added my Spotify credentials to my Supabase project.

Here is my redirect URL

<img width="888" alt="Screenshot 2024-01-30 at 4 26 17 PM" src="https://github.com/supabase/supabase-flutter/assets/49491200/a80ab36b-7bdc-488b-9052-cbfa66f6fba6">

I am using the supabase UI widget to trigger `signInWithOAuth` but I have gotten the same result by calling the method directly with the same params.

```dart
SupaSocialsAuth(
              socialProviders: const [OAuthProvider.spotify],
              redirectUrl: kIsWeb ? null : 'record://auth',
              onSuccess: (Session session) {
                log(session.toString());
              },
              onError: (e) {
                log(e.toString());
              },
            )
```

Here is the error I get from the above method.
```
PlatformException(Error, Error while launching https://ayjhezlfczlzymyzwngk.supabase.co/auth/v1/authorize?provider=spotify&redirect_to=record%3A%2F%2Fauth&flow_type=pkce&code_challenge=z45CR8YiGxFR5gi_AKB5TdcGGHbOHViCxi8Yivo9d8E&code_challenge_method=s256, null, null)
```

Here is what I added to my `info.plist`

```plist
<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleTypeRole</key>
			<string>Editor</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>record</string>
			</array>
		</dict>
	</array>
```

dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"

here is the result of the above command

├── supabase_auth_ui 0.4.1
│   └── supabase_flutter...
├── supabase_flutter 2.1.0
│   ├── supabase 2.0.3
│   │   ├── functions_client 2.0.0
[...]

PR #839, opened 2024-02-15
fix(gotrue, supabase_flutter): Throw error when parsing auth URL that contains an error description.

dshukertjr:

The SDK should throw an AuthException when parsing an auth URL that contains an error description.

This PR updates the `_isAuthCallbackDeeplink` in supabase_flutter to return `true` if the URL contains `error_description`, and also makes sure proper exception is thrown when parsing auth URL containing an error description.

Related https://github.com/supabase/supabase-flutter/issues/819#issuecomment-1944364896

### DIFF
--- a/packages/gotrue/test/client_test.dart
+++ b/packages/gotrue/test/client_test.dart
@@ -43,6 +43,7 @@ void main() {
           'apikey': anonToken,
         },
         asyncStorage: asyncStorage,
+        flowType: AuthFlowType.implicit,
       );
 
       adminClient = client = GoTrueClient(
@@ -62,6 +63,7 @@ void main() {
           'apikey': anonToken,
         },
         asyncStorage: asyncStorage,
+        flowType: AuthFlowType.implicit,
       );
     });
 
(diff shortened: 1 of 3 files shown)


packages/gotrue/test/client_test.dart
```diff
@@ -43,9 +43,10 @@ void main() {
           'apikey': anonToken,
         },
         asyncStorage: asyncStorage,
+        flowType: AuthFlowType.implicit,
```
Vinzent03, line 46: I'm unsure about this change as it changes the whole file to now use implicit again, doesn't it? We should properly do all tests for pkce and implicit.

dshukertjr: >I'm unsure about this change as it changes the whole file to now use implicit again, doesn't it?

Yes, it does. The issue here is that implicit flow was not tested anywhere and the `PKCE enabled client` group doesn't add anything. Now we have the same set of tests that we always had before v2.

>We should properly do all tests for pkce and implicit.

At least now we have the same set of test cases as the js SDK. We can always add more tests if we want to in the future.

Vinzent03 (approved):

PR #841, opened 2024-02-16
fix: Correct the id value passed to `.unlinkIdentity()` method

Alexays:
Just implemented manual linking, and found that the unlink method just doesnt work cause the wrong id is used.
```
AuthException(message: identity_id must be an UUID, statusCode: 400)
```
Here's the fix :)


dshukertjr (approved):

dshukertjr:
Thanks for this fix!

PR #853, opened 2024-03-09
docs: Fixed imprecision in Custom Local Storage README.md

alessandrobelottidev:
The README.md is not completely precise when providing an example of a Custom Local Storage. 

To check if it hasAccessToken it should return a bool (and so should return storage.containsKey(key: supabasePersistSessionKey) ).

And when retrieving the accessToken it should return a string (so the correct code would be: storage.read(key: supabasePersistSessionKey) )

Docs update

The two functions accessToken() and hasAccessToken() return the wrong type. It's probably simply an oversight when copy-pasting this code.

## What is the new behavior?
I swapped the two return lines so that it behaves correctly.

## Additional context
When trying to understand how to create a custom local storage I spent some time figuring out why my code was not behaving correctly, but then saw the wrong return type.

Im new to the open source community so I don't know if Im contributing correctly and if I explained myself correctly in this PR.

### DIFF
--- a/packages/supabase_flutter/README.md
+++ b/packages/supabase_flutter/README.md
@@ -388,12 +388,12 @@ class MySecureStorage extends LocalStorage {
 
   @override
   Future<String?> accessToken() async {
-    return storage.containsKey(key: supabasePersistSessionKey);
+    return storage.read(key: supabasePersistSessionKey);
   }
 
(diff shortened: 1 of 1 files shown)


dshukertjr (approved):
Thanks for the contribution!

PR #862, opened 2024-03-17
Ability to save requests offline and reexecute them later.

ConProgramming:
This is very useful for supporting offline mode in apps. Not sure if should be implemented like this, but this works for my use case.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -93,6 +93,98 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
     );
   }
 
+  Future<Map<String, dynamic>> jsonifyRequest() async {
+    final String? method = _method;
+
+    if (_count != null) {
+      if (_headers['Prefer'] != null) {
+        final oldPreferHeader = _headers['Prefer'];
+        _headers['Prefer'] = '$oldPreferHeader,count=${_count!.name}';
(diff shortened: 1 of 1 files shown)


dshukertjr:
If you want to execute the request in a later time, you can just not call `await` on the method. Then you can just call `await` when you actually want to execute it. 

```dart
final request = supabase.from('table').select();

... some time later
final data = await request;
```

ConProgramming:
I need to serialize the request to Json to save offline.  Saving it to a HiveBloc

dshukertjr:
@ConProgramming 
Cool. You can create your own extension method to implement this for your particular use case if you are not doing it already. Unfortunately, we will not add support for these methods on this SDK.

ConProgramming:
What's the recommended way to do that? I thought I would have to patch SDK

dshukertjr:
You can read about extension methods here:
https://dart.dev/language/extension-methods

ConProgramming:
Thanks

ConProgramming:
It seems extension methods won't work, because I cannot override private methods.

Is this possible without updating the sdk?

dshukertjr:
Right, I missed that. 

>Is this possible without updating the SDK?

I suppose not. You can fork the repo, add the feature, and add the package to your Flutter app. 
https://stackoverflow.com/questions/54022704/how-to-add-a-package-from-github-in-flutter

PR #865, opened 2024-03-21
Added some data retrieval methods to fetch data easily.

Aniketkhote:

feature

To fetch single latest or oldest data currently need to apply order() on created_at field and then single() method to get one also to fetch null, true or false we need to pass in isFilter.

### latest() and oldest()

Before

final data = await supabase.from('users').select().order('created_at, ascending: true).single();

Now

final data = await supabase.from('users').select().latest();
or
final data = await supabase.from('users').select().latest('id');

same for oldest.

### isNull() , isTrue() and isFalse()

Before

final data = await supabase.from('users').select().isFilter('data', null);

Now

final data = await supabase.from('users').select().isNull('data');
or
final data = await supabase.from('users').select().isTrue('data');
or
final data = await supabase.from('users').select().isFalse('data');

### DIFF
--- a/packages/postgrest/lib/src/postgrest_filter_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_filter_builder.dart
@@ -1,7 +1,7 @@
 part of 'postgrest_builder.dart';
 
 class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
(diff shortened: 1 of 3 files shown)


dshukertjr:
Thanks for the suggestion, but unfortunately we are not adding these new methods.

Issue #872, opened 2024-03-27
Timeout on `refreshSession`


ConProgramming:
I'm calling `await Supabase.instance.client.auth.refreshSession().timeout(const Duration(seconds: 30));` when my app switches from not having connection to having connection. This is timing out if the "Access token (JWT) expiry time" has passed. (It defaults to 1 hour, but can reproduce this issue easily if I change it to 2 minutes)

ConProgramming:
From debugging, looks like it gets stuck returning the existing future on the internal completer.

https://github.com/ConProgramming/supabase-flutter/blob/main/packages/gotrue/lib/src/gotrue_client.dart#L1071

dshukertjr:
Yeah, I'm working on a fix around the session refresh flow, which should fix this issue too. Hopefully I can ship the fix soon.

ConProgramming:
Worked on a PR. Testing it some, seems to work well https://github.com/supabase/supabase-flutter/pull/873

Obviously would have to remove logs to merge, but kept for readability. Let me know how this looks.

dshukertjr:
What is the purpose of calling refresh session when coming back online in the first place? You shouldn't need to explicitly call refresh session on your own as the SDK will handle refreshing the session whenever needed.

ConProgramming:
I was using it with a .timeout, then signing out if timed out. Because if I didn't then sdk calls were just hanging infinitely due to the internal refreshSession completer never completing.

dshukertjr:
What call was hanging on what occasion? Rather than trying to change the refreshSessoin method, we should fix the root issue. You shouldn't have to be calling refreshSession in the first place.

ConProgramming:
Mentioned at the top of this issue, calling refreshSession hands when my app switches from not having connection to having connection if the "Access token (JWT) expiry time" has passed (It defaults to 1 hour, but can reproduce this issue easily if I change it to 2 minutes)

dshukertjr:
@ConProgramming 
Yes, but why are you calling refreshSession in the first place?

ConProgramming:
If I dont call refreshSession, all sdk calls still hang either way

They are all hanging on the same completer

Calling refreshSession manually confirms that is the error

dshukertjr:
>If I dont call refreshSession, all sdk calls still hang either way

[...]

PR #873, opened 2024-03-28
Offline Support for Refreshing Sessions

ConProgramming:
Won't error out or freeze if refreshing a session fails, will just return null.

Flutter can watch Connectivity for a change, and then call refreshSession.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -1067,10 +1067,12 @@ class GoTrueClient {
         (value) => null,
         onError: (error, stack) => null,
       );
-      _refreshTokenCompleter!.future.timeout(const Duration(seconds: 10), onTimeout: () {
+       (_refreshTokenCompleter!.future as Future<void>).timeout(const Duration(seconds: 10), onTimeout: () {
         if (!_refreshTokenCompleter!.isCompleted) {
           _refreshTokenCompleter!.completeError(TimeoutException(), StackTrace.current);
         }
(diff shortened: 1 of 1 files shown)


dshukertjr (commented):
Thanks for the PR. As much as I love the lion emojies, let's get rid of them 🦁

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -581,7 +581,7 @@ class GoTrueClient {
 
   /// Force refreshes the session including the user data in case it was updated
   /// in a different session.
-  Future<AuthResponse> refreshSession() async {
+  Future<AuthResponse?> refreshSession() async {
```
dshukertjr, line 584: This would be a breaking change, and we cannot do it. Same with `setSession` and `recoverSession`.

ConProgramming: Could we expose a tryRefreshSession, that refreshSession then wraps, and then keeps the current method of throwing an exception?

ConProgramming:
> Thanks for the PR. As much as I love the lion emojies, let's get rid of them 🦁

Yeah left for observability. Would remove in final.

dshukertjr (changes requested):
Left a comment here, but let's take a step back and get to the bottom of the issue here.

https://github.com/supabase/supabase-flutter/issues/872#issuecomment-2024338624

dshukertjr:
Closing this PR as the original issue has been closed.

Issue #829, opened 2024-02-07
tokens it refreshes the token until the rate limit is reached


BharathKumar-99:
from yesterday iam getting a issue with tokens it refreshes the token until the rate limit is reached can any one help

iam not sure if this helps but here is my code 

 `  static Future initialize() => Supabase.initialize(
      url: KeyConstants.SUPABASE_URL!,
      anonKey: KeyConstants.SUPABASE_ANON_KEY!,
      debug: true);`

this the auth change stream code
`final supabaseAuth = SupaFlow.client.auth.onAuthStateChange
    .map<String>((response) => response.session!.user.id)
    .switchMap((id) => id.isEmpty
        ? Stream.value(null)
        : getFilteredRowStream(id).map((row) => log(row.toString())))
    .asBroadcastStream();`

iam refresh token should not triggers until the code 429 is thrown

I am using supabase_flutter: ^2.3.2

dshukertjr:
Again, we need you to follow the issue format for us to assist you. Please provide a detailed step-by-step of the actions you are performing and provide clear expected results and actual results you are getting with the issue you are facing.

BharathKumar-99:
iam not sure how to reproduce this as iam getting this error with out soing anything when app starts

dshukertjr:
So you are getting the error as a result of calling `Supabase.initialize()`? Could you copy and paste the content of the error here?

Issue #317, opened 2022-12-28
New token refresh algorithm  on gotrue-js


dshukertjr:
*Posting this without reading too much into the context, so might not be relevant for us, but wanted to get it out there for visibility. 

Gotrue-js is updating their token refresh algorithm, here. Might be something worth looking into!
https://github.com/supabase/gotrue-js/pull/564

Personally, I think the current implementation is okay though unless someone knows any issues.

matthewwong525:
So my supabase app supports offline usage (through flutter) and oftentimes if I get reports of my users being automatically "logged out". This is because the refresh token has trouble actually getting a refresh once the internet connection has been restored. Anyways, my solution to this was to periodically refresh the token (before it's expired) and set the expiration date to 7 days.

d-e-h-i-o:
Our users also get logged out sometimes. Often the following exception is thrown: 
`AuthException('Access token refresh retry limit exceeded.');`
`GoTrueClient._setTokenRefreshTimer (gotrue_client.dart:552)
GoTrueClient._callRefreshToken (gotrue_client.dart:606)`

Maybe this is expected when a short expiration time is set, and the internet connection not reliable.

Issue #860, opened 2024-03-15
[auth] Bug: Auth currentSession is null


iosephmagno:
**Describe the bug**
Hello @dshukertjr just reporting that we were hit but the null auth session bug.
The code is standard one:
```
 await Supabase.initialize(
      url: _url,
      anonKey: _publicAnonKey,
      debug: kDebugMode, // optional
    );
 debugPrint('$_debugPrefix Auth.currentSession is: ${auth.currentSession}');
```
`auth.currentSession` is null.

To fix issue, I had to uninstall app and reinstall it from scratch.
No errors were found on supabase console.

**To Reproduce**
Unfortunately, there are no clear steps to reproduce the bug. It occurs randomly out of blue. In our case, usually once every 2 months and doesnt autoresolve with app launch, user must delete app.

**Expected behavior**
Client should always be capable to auto-resolve any auth issue while fetching a new session from GoTrue server.

**Version (please complete the following information):**
On Linux/macOS
Please run `dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"` in your project directory and paste the output here.

```
── supabase_flutter 2.3.4
│   ├── supabase 2.0.8
│   │   ├── functions_client 2.0.0
│   │   ├── gotrue 2.5.1
│   │   ├── postgrest 2.1.1
│   │   ├── realtime_client 2.0.1
│   │   ├── storage_client 2.0.1
```

**Additional context**
Issue resolves by uninstalling the app, hence cache is involved. I mean the issue is caused by what the plugin has saved to cache and revert as soon as cache is cleared.

**Potential workaround**
Since client knows when the `currentSession` is null, can it clear the cache if `currentSession` is null after X retry?
Thats doesnt fix the root problem but should workaround the issue.

iosephmagno:
Forgot to mention, on android we cleared app's cache without deleting app and issue resolved also this way.

iosephmagno:
This also may help. After initialize()n we listen auth event and also print eventual error.
But when issue occurs we get no data from the listen and no error from `onError`.

```
// Listen on auth change event and take action
  auth.onAuthStateChange.listen((data) async {
    if (data.event == AuthChangeEvent.tokenRefreshed) {
     …
    }
    if (data.event == AuthChangeEvent.signedIn) {
    …
    }
    if (data.event == AuthChangeEvent.userUpdated) {
    …
    }
    if (data.event == AuthChangeEvent.initialSession) {
  …
[...]

Issue #827, opened 2024-02-06
Supabase throwing JWT expired. PostgrestException(message: JWT expired, code: PGRST301, details: Unauthorized, hint: null)


Aadim1:
A couple of hours ago, I started having problem I never had before. The supabase upon calling the api started throwing. `PostgrestException(message: JWT expired, code: PGRST301, details: Unauthorized, hint: null)`. My JWT expiration time was 3600 seconds. I got it down to 10 seconds, to see if the issue persists it did. I played around with from range 10 to 1000 to 3600, all of them started throwing JWT expired, after some time. I have awaited supabase at main(), and I also didn't have the authOptions. I also changed the authOptions to autoRefreshToken.

```
final sup = await Supabase.initialize(
    url: 'URL',
    anonKey:
        'ANON_KEY',
    authOptions: const FlutterAuthClientOptions(
      autoRefreshToken: true,
    ),
  );
```

I then thought maybe I was being rate limited, but I forced the session to refresh every 1 seconds with `supabase.auth.refreshSession();` and got a rate limit error `AuthException(message: Rate limit exceeded, statusCode: 429)`. This was working fine around 3 hours ago when I was last testing it. 

I am using supabase for auth, db, and realtime. 

```

try {
    final appointments = await supabase
        .from("<table>")
        .select('''
            ...
          ''')
    log('Appointment: $appointments ${lastAppointmentRefresh?.toUtc().toIso8601String()}');
    return appointments;
  } catch (e) {
    if (e is PostgrestException && e.code == "PGRST301" && retries > 0) {
      log("JWT EXpired...");
      await supabase.auth.refreshSession();
      return await fetchAllAppointmentNetwork(
          lastAppointmentRefresh, businessId, locationId,
          retries: retries - 1);
    }

    log('Fetch Appointment Error: $e');
    return null;
  }
  ```

When I catch the error, and manually refresh the token, everything works fine in the first try.

I don't know if it matters but this is how I am using auth stream

```

supabaseAuthStream?.cancel();
    supabaseAuthStream =
        supabase.auth.onAuthStateChange.listen((authState) async {
      final session = authState.session;
      final event = authState.event;
      if (session == null) {
        changeLoggedInState(false);
      } else {

        if (some_prefs_condition) {
          changeLoggedInState(false);
          return;
        }

        if (authState.event == AuthChangeEvent.tokenRefreshed) {
          supabase.auth.refreshSession();
        }

        changeLoggedInState(true);
      }
    });

[...]

PR #879, opened 2024-04-02
feat(gotrue, supabase_flutter): New auth token refresh algorithm

dshukertjr:

Change the way we refresh the auth tokens. It also creates two new auth methods `startAutoRefresh()` and `stopAutoRefresh()` to start and stop the token refresh process. 

We have a relatively complex algorithm to refresh the auth tokens. It can be hard to debug what is happening through out the token refresh flow. 

## What is the new behavior?

The flow is simplified, and easier to understand. 

1. Every 10 seconds the session is checked if it's expired. Call this "tick"
1. If expired, request a new token to the auth server
  - If the request fails due to a network issue, it retries with exponential backoff (i.e. 200ms, 400ms, 800ms ...)
  - Any additional token refresh attempt will also be bundled into the same `Completer`
  - The retry keeps going until the next tick
1. The completer is released (either succeded and returns the token or throws an exception) before the next tick.

In addition to the algorithm update, this PR has introduced the start and stop auto refresh methods, and uses them with the lifecycle events of Flutter. Specifically, when the app is in the background, it stops the SDK to attempt to refresh the token. This is because when the app is in the background, HTTP requests may or may not go through and the app could be killed any time. In order to reduce the chance of something wrong happening with the token refresh requests, the refresh requests should only be fired when the app is in the foreground. 

## Additional context

Closes https://github.com/supabase/supabase-flutter/issues/317
Related https://github.com/supabase/supabase-flutter/issues/171
Related https://github.com/supabase/supabase-flutter/issues/788
Related https://github.com/supabase/supabase-flutter/issues/827
Closes https://github.com/supabase/supabase-flutter/issues/829
Related https://github.com/supabase/supabase-flutter/issues/860
Closes https://github.com/supabase/supabase-flutter/issues/872

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -113,6 +113,9 @@ class GoTrueClient {
       client: this,
       fetch: _fetch,
     );
+    if (_autoRefreshToken) {
+      startAutoRefresh();
+    }
   }
 
   /// Getter for the headers
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -76,9 +76,6 @@ class SupabaseAuth with WidgetsBindingObserver {
           // ignore: invalid_use_of_internal_member
           .notifyAllSubscribers(AuthChangeEvent.initialSession);
     }
-    if (_autoRefreshToken) {
-      Supabase.instance.client.auth.startAutoRefresh();
-    }
   }
 
   /// Recovers the session from local storage.


packages/gotrue/lib/src/types/auth_exception.dart
```diff
@@ -24,3 +24,33 @@ class AuthException implements Exception {
 class AuthPKCEGrantCodeExchangeError extends AuthException {
   AuthPKCEGrantCodeExchangeError(String message) : super(message);
 }
+
+class AuthSessionMissingError extends AuthException {
```
Vinzent03, line 28: I would rename all these classes to Exceptions. From the documentations: [Exception](https://api.dart.dev/stable/3.3.3/dart-core/Exception-class.html) and [Error](https://api.dart.dev/stable/3.3.3/dart-core/Error-class.html). Exceptions should be caught, which is the case here. Errors are more for wrong code. Additionally, they implement `Exception` and not `Error`.

packages/supabase_flutter/lib/src/supabase_auth.dart
```diff
@@ -107,33 +108,15 @@ class SupabaseAuth with WidgetsBindingObserver {
   void didChangeAppLifecycleState(AppLifecycleState state) {
     switch (state) {
       case AppLifecycleState.resumed:
-        _recoverSupabaseSession();
+        Supabase.instance.client.auth.startAutoRefresh();
+      case AppLifecycleState.detached:
+      case AppLifecycleState.inactive:
+      case AppLifecycleState.paused:
+        Supabase.instance.client.auth.stopAutoRefresh();
```
Vinzent03, line 115: What's the reason to stop auth refresh here? From what I read in the [documentation](https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html) the app may still be running in the background.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -917,33 +924,139 @@ class GoTrueClient {
...
+    );
 
-      return AuthResponse(session: session);
+    await Future.delayed(Duration.zero);
+    await _autoRefreshTokenTick();
+  }
+
+  /// Stops an active auto refresh process running in the background (if any).
+  ///
+  /// If you call this method any managed visibility change callback will be
```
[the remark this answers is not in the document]
dshukertjr: Sorry, this is taken from the JS SDK and I hadn't removed it. It's irrelevant for Flutter SDK.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -917,33 +924,137 @@ class GoTrueClient {
...
+    );
+
+    await Future.delayed(Duration.zero);
+    await _autoRefreshTokenTick();
+  }
+
+  /// Stops an active auto refresh process running in the background (if any).
+  ///
+  /// If you call this method any managed visibility change callback will be
+  /// removed and you must manage visibility changes on your own.
```
dshukertjr, line 980:
```suggestion
```

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -987,139 +1100,66 @@ class GoTrueClient {
     return OAuthResponse(provider: provider, url: oauthUrl);
   }
 
-  void _saveSession(Session session) async {
+  /// set currentSession and currentUser
+  /// process to _startAutoRefreshToken if possible
```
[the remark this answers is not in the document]
dshukertjr: Same here. Forgot to remove the comments from the JS SDK. 
```suggestion
```

packages/supabase_flutter/lib/src/supabase_auth.dart
Vinzent03: Oh sorry, seems like I didn't read that.
But does it matter if a refresh request fails? The app could still be running, so the application could send requests to supabase with outdated access token. If the refresh fails, then requests from the application will also fail, which is fine.

dshukertjr: No problem at all!

>The app could still be running, so the application could send requests to supabase with outdated access token. If the refresh fails, then requests from the application will also fail, which is fine.

Actually, `stopAutoRefresh()` only stops periodically checking the token if it's expired then refresh if expired, and if a user would to make a request after `stopAutoRefresh()` has been called, the auth HTTP client will still refresh the token as needed. 

All it's doing is avoiding a slim chance of the auth refresh request firing, which reaches the server, but before the response comes back the connection is lost. This is where the app could be left with an expired JWT. This could very much happen when the app is in the foreground as well, but if the app is running for a long time in the background, there might be more chances of something like this happening. 

It was advice from our Auth engineer and we also recommend it for React native apps. On step 4 there is pretty much the same code:
https://supabase.com/docs/guides/auth/quickstarts/react-native

Vinzent03: Oh yeah I forgot that the auth http client still exists. Then you are right this is fine.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -987,139 +1090,65 @@ class GoTrueClient {
...
   }
 
   /// Generates a new JWT.
   ///
   /// To prevent multiple simultaneous requests it catches an already ongoing request by using the global [_refreshTokenCompleter].
   /// If that's not null and not completed it returns the future of the ongoing request.
   ///
-  /// To call [_callRefreshToken] during a running request [ignorePendingRequest] is used to bypass that check.
-  ///
   /// When a [ClientException] occurs [_setTokenRefreshTimer] is used to schedule a retry in the background, which emits the result via [onAuthStateChange].
```
Vinzent03, line 1109: This comment is outdated now. This method doesn't retry anything now and just reports any error and may sign out.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -917,33 +924,129 @@ class GoTrueClient {
...
+    } catch (error) {
+      notifyException(error);
+      rethrow;
+    }
+  }
+
+  /// Starts an auto-refresh process in the background. Close to the time of expiration a process is started to
+  /// refresh the session. If refreshing fails it will be retried for as long as necessary.
+  ///
+  /// If you set the `autoRefreshToken` to `true`, you don't need to call this function, it will be called for you.
```
Vinzent03, line 964: This is not really true. Only when using the `supabase_flutter` package this method is called, nowhere else. Additionally, `autoRefreshToken` should be respected when calling `startAutoRefresh` in `supabase_flutter` as well.

Vinzent03 (changes requested):

packages/gotrue/lib/src/gotrue_client.dart
dshukertjr: Great catch! Pushed a fix.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -987,139 +1090,65 @@ class GoTrueClient {
...
-    } catch (error, stack) {
-      if (error is AuthException) {
-        if (error.message.startsWith('Invalid Refresh Token:')) {
-          await signOut();
-        }
-      }
-      if (!_refreshTokenCompleter!.isCompleted) {
-        _refreshTokenCompleter!.completeError(error, stack);
-      }
-      _onAuthStateChangeController.addError(error, stack);
```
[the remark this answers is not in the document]
dshukertjr: Great catch! I added it back!

packages/gotrue/lib/src/gotrue_client.dart
Vinzent03: I think the auto refresh should be started when only using the supabase or gotrue  package as well. Not only when using the flutter package

dshukertjr: Makes sense. Let me change the code right now.

dshukertjr: @Vinzent03 Updated it to call the startAutoRefresh from within gotrue.

Vinzent03 (approved):

packages/supabase_flutter/lib/src/supabase_auth.dart
iosephmagno: @dshukertjr any specific reason why handling the socket exception wan't enough?

on SocketException catch (e) {
      debugPrint('Exception thrown while trying to refresh token: $e');
   }

Was thinking... isn't it best to refresh token within a WorkManager periodic task? It will run several times per day, return if DateTime.now < expiryDate, and invoke api otherwise. What do you think?

PR #881, opened 2024-04-02
fix: Don't send access token  in rest broadcast

### LINKED ISSUE #867: Invalid API key on sendBroadcastMessage method on realtime.
I get this error with the  when using the sendBroadcastMessage function, here is my body 
`{"message":"Invalid API key","hint":"Double check your Supabase`anon`  or  `service_role`  API key."}`

**My code**
This is sending a broadcast messages.

```
void sendBroadcastMessage(
    String chatRoomId,
    TextEditingController textController,
  ) async {
    final message = textController.text;
    if (message.isEmpty) return;
    final channel = supabase.channel('chat_$chatRoomId').subscribe();

    final res = await channel.sendBroadcastMessage(event: 'chat', payload: {
      'message': message,
      'user_id': Supabase.instance.client.auth.currentUser!.id,
      'timestamp': DateTime.now().millisecondsSinceEpoch,
      'message_type': 'text',
    });
    print(res);
  }
```
`res`= 
image

Screenshot of a network debugger.
image

Version:
dart pub deps
```

└── supabase_flutter 2.3.4
    ├── supabase 2.0.8
    │   ├── functions_client 2.0.0
    │   ├── gotrue 2.5.1
    │   ├── postgrest 2.1.1
    │   ├── realtime_client 2.0.1
    │   ├── storage_client 2.0.1
```
flutter --version

```
Flutter 3.19.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision ba39319843 (2 weeks ago) • 2024-03-07 15:22:21 -0600
Engine • revision 2e4ba9c6fb
Tools • Dart 3.3.1 • DevTools 2.31.1
```
**Additional context**

All functions related to the Supabase are actively working on my project

None of the problems with **apiKey** and **anon** are experienced when launching my application, the only problem seems to be this function.

Vinzent03:

Bug fix

The access token is sent as `apikey`, but from what I see it doesn't need/take authorization at all.

## What is the new behavior?

The `apikey` passed in the default header in realtime client initialization in the supabase package is used. That's the anon apikey and not the user's one. 

## Additional context

I noticed that it was enough to remove the apikey field in the send method, because the `apikey` field gets also set in the `AuthHttpClient` if it's not already set, but I thought it's either way good to add it to the headers directly as well. realtime-js does this as well.

close #867

### DIFF
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -483,11 +483,7 @@ class RealtimeChannel {
     }
 
     if (!canPush && type == RealtimeListenTypes.broadcast) {
-      final headers = {
-        'Content-Type': 'application/json',
-        'apikey': socket.accessToken ?? '',
-        ...socket.headers
-      };
+      final headers = {'Content-Type': 'application/json', ...socket.headers};
       final body = {
         'messages': [
           {
--- a/packages/realtime_client/test/channel_test.dart
+++ b/packages/realtime_client/test/channel_test.dart
@@ -248,6 +248,7 @@ void main() {
       mockServer = await HttpServer.bind('localhost', 0);
       socket = RealtimeClient(
         'ws://${mockServer.address.host}:${mockServer.port}/realtime/v1',
+        headers: {'apikey': 'supabaseKey'},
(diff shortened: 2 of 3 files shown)


dshukertjr (approved):
Looks great! 🙌

Issue #874, opened 2024-03-28
[auth] Add support for `signInAnonymously()`


pedromassango:
**Is your feature request related to a problem? Please describe.**
The JavaScript SDK has support for [signInAnonymously()](https://supabase.com/docs/reference/javascript/auth-signinanonymously), this is a request to add the same feature on the Flutter sdk.

**Describe the solution you'd like**
N/A

**Describe alternatives you've considered**
There is no alternatives.

**Additional context**
N/A

dshukertjr:
I like how closely you keep an eye on the js sdk. It will come to the Flutter SDK soon as well.

pedromassango:
> I like how closely you keep an eye on the js sdk. It will come to the Flutter SDK soon as well.

Is the work being done in public? I can't ship without this feature / random email not an option

dshukertjr:
We ask you to wait patiently. It will arrive as soon as possible. Note that due to the current limitation of the auth server of Supabase, you will only be able to upgrade the anonymous users to a authenticated users through email + password or the web based OAuth login. Open ID connect (native Google and Apple sign-in) will not yet be supported.

pedromassango:
Sure. We can live with that!

muezz:
Hey @dshukertjr, do you have an ETA on this? Coincidentally, I happen to be working on a project which would greatly benefit from this. If it is going to be within a few days, I would rather just wait for it.

antonkarliner:
+++ also waiting to implement this in flutter 🙏🏼

moriosoft:
Waiting as well. Looking forward to it :)

PR #883, opened 2024-04-05
feat(gotrue): Add `signInAnonymously()` method

dshukertjr:

Adds `signInAnonymously()` method to support Anonymous sign-in.

This PR should be merged after https://github.com/supabase/supabase-flutter/pull/879

Closes https://github.com/supabase/supabase-flutter/issues/874


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -813,8 +854,11 @@ class GoTrueClient {
         await admin.signOut(accessToken, scope: scope);
       } on AuthException catch (error) {
         // ignore 401s since an invalid or expired JWT should sign out the current session
+        // ignore 403s since user might not exist anymore
         // ignore 404s since user might not exist anymore
-        if (error.statusCode != '401' && error.statusCode != '404') {
+        if (error.statusCode != '401' &&
+            error.statusCode != '403' &&
```
dshukertjr, line 860: This edit has been made, because for deleted users on the latest Supabase auth server, the following error is returned:
`AuthException(message: User from sub claim in JWT does not exist, statusCode: 403)`

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -132,6 +132,40 @@ class GoTrueClient {
...
+
+      final session = authResponse.session;
+      if (session != null) {
+        _saveSession(session);
+        notifyAllSubscribers(AuthChangeEvent.signedIn);
+      }
+
+      return authResponse;
+    } catch (error) {
+      notifyException(error);
```
Vinzent03, line 164: The other sign in methods don't have a try catch, which I think is better, because this method is only directly called by the user, so exceptions don't need to go to the stream.

Vinzent03 (changes requested):

packages/gotrue/lib/src/gotrue_client.dart
dshukertjr: Makes sense!

packages/gotrue/test/client_test.dart
```diff
@@ -80,6 +80,14 @@ void main() {
       );
     });
 
+    test('anonymous sign-in', () async {
```
Vinzent03, line 83: You could add a check to the test that the user  anonymous is.

Vinzent03 (changes requested):

Vinzent03 (approved):

Issue #894, opened 2024-04-19
Add support for SSE response on functions client. 


dshukertjr:
**Is your feature request related to a problem? Please describe.**
Yes. Currently, there seems to be no possibility to invoke an Edge Function that returns a server-sent events using the client library.

**Describe the solution you'd like**
I would like to see the implementation of an additional function, which could return a `Stream` of server-sent events instead of standard promise.

**Describe alternatives you've considered**
One alternative I considered was to simply await the full result. However, this would leave the user waiting for a long period, which is not an ideal user experience.

**Additional context**
The motivation is a project, where we are looking to implement a ChatGPT interface within our Flutter application, and we need to stream the response of the API call to OpenAI to the user in real-time.

Implementing this functionality on the backend as an Edge Function was straightforward enough - I followed the tutorial on Streaming Data in Edge Functions by Supabase (https://www.youtube.com/watch?v=9N66JBRLNYU).

Copied from https://github.com/supabase/functions-js/issues/67

nietsmmar:
Because this feature does not exist yet I used a fork of `dart_sse_client` which provides web-support and used it together with `fetch_client` library.

This is super hacky but works for (most) part. But now since the upgrade from 

`supabase-flutter` package > 2.4.0
`gotrue` package > 2.5.1

it does not work anymore. I always get `415 Unsupported Media Type` as a response from my Edge-Function. I think it is some change in `gotrue` package to version `2.6.0` although I couldn't figure out which. 

It seems like this issue here is just a change of two lines, regarding the js-PR. I would really appreciate to finally be able to use streaming in flutter from the official supabase package. As of now it doesn't even work with the workaround anymore.

Thanks for all the great work.

dshukertjr:
@nietsmmar I did some quick research, and for the Dart library, it's not going to be just updating two lines of code, but a few more than that. I want to properly test it out, and with some other tasks that I have, it might take some days for me to get to this one. Sorry for the inconvenience.

nietsmmar:
@dshukertjr Thank you for checking it out! Really appreciate it! Looking forward to the fix :)

oliverbytes:
I'm eagerly waiting as well. Thanks so much to anyone doing the work.

nietsmmar:
@dshukertjr 
I initialized it now like this:

```
final client = FetchClient(mode: RequestMode.cors);
  await Supabase.initialize(
    url: config.api.url,
    anonKey: config.api.anonKey,
[...]

PR #905, opened 2024-04-26
feat(functions_client): Add SSE support to invoke method

dshukertjr:

Adds support for streaming response of an edge function that returns Server Sent Events. 

Fixes https://github.com/supabase/supabase-flutter/issues/894

There is no support for SSE, and developers have to use other libraries to listen to SSE. 

## What is the new behavior?

Users can listen to SSE like the following:
```dart
final response = await supabase.functions.invoke('sse-function');
response.data
    .transform(const Utf8Decoder())
    .listen((val) {
      print(val);
    });
```

## Additional Context

On the web, you would need to pass a custom HTTP client like this:

```dart
final client = FetchClient(mode: RequestMode.cors);
await Supabase.initialize(
  url: supabaseUrl,
  anonKey: supabaseKey,
  debug: false,
  httpClient: kIsWeb ? client : null,
);
```

### DIFF
--- a/packages/functions_client/lib/functions_client.dart
+++ b/packages/functions_client/lib/functions_client.dart
@@ -1,4 +1,6 @@
 library functions_client;
 
+export 'package:http/http.dart' show ByteStream;
+
 export 'src/functions_client.dart';
 export 'src/types.dart';
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -43,6 +43,20 @@ class FunctionsClient {
   /// [headers]: object representing the headers to send with the request
   ///
   /// [body]: the body of the request
+  ///
+  /// ```dart
+  /// // Call a standard function
+  /// final response = await supabase.functions.invoke('hello-world');
+  /// print(response.data);
+  ///
+  /// // Listen to Server Sent Events
+  /// final response = await supabase.functions.invoke('sse-function');
+  /// response.data
+  ///     .transform(const Utf8Decoder())
+  ///     .listen((val) {
+  ///       print(val);
+  ///     });
+  /// ```
   Future<FunctionResponse> invoke(
     String functionName, {
     Map<String, String>? headers,
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -3,7 +3,6 @@ import 'dart:convert';
 import 'package:functions_client/src/constants.dart';
 import 'package:functions_client/src/types.dart';
 import 'package:http/http.dart' as http;
-import 'package:http/http.dart';
 import 'package:yet_another_json_isolate/yet_another_json_isolate.dart';
 
 class FunctionsClient {
@@ -52,7 +51,6 @@ class FunctionsClient {
   }) async {
     final bodyStr = body == null ? null : await _isolate.encode(body);
 
-    late final Response response;
     final uri = Uri.parse('$_url/$functionName');
 
     final finalHeaders = <String, String>{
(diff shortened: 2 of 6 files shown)


packages/functions_client/lib/src/functions_client.dart
```diff
@@ -52,67 +65,42 @@ class FunctionsClient {
...
           ? ""
-          : await _isolate.decode(utf8.decode(response.bodyBytes)),
-      'application/octet-stream' => response.bodyBytes,
-      _ => utf8.decode(response.bodyBytes),
-    };
+          : await _isolate.decode(utf8.decode(bodyBytes));
+    } else if (responseType == 'application/octet-stream') {
+      data = await response.stream.toBytes();
+    } else if (responseType == 'text/event-stream') {
+      data = response.stream;
```
dshukertjr, line 98: Here, I thought about adding `.transform(const Utf8Decoder())` so that the developers don't have to add them by themselves, but `Utf8Decoder()` has a `allowMalformed` option, and I wanted to give the developers full control of what to do with the option, so I'm just returning raw ByteStream. 
```dart
      data = response.stream.transform(const Utf8Decoder());
```

grdsdev (approved):

packages/functions_client/lib/src/functions_client.dart
```diff
@@ -52,67 +65,42 @@ class FunctionsClient {
...
-        break;
-    }
+    final request = http.Request(method.name, uri);
 
+    finalHeaders.forEach((key, value) {
+      request.headers[key] = value;
+    });
+    if (bodyStr != null) request.body = bodyStr;
+    final response = await (_httpClient?.send(request) ?? request.send());
     final responseType = (response.headers['Content-Type'] ??
```
Vinzent03, line 82: From the [documentation](https://pub.dev/documentation/http/latest/http/BaseResponse/headers.html) it says, that all headers are converted to lowercase. So the first check should never succeed.

Vinzent03 (approved):
The documentation for `data`  in the `FunctionResponse` should be updated as well.

Vinzent03 (changes requested):
(I accidentally approved)

packages/functions_client/lib/src/functions_client.dart
dshukertjr: Hmm, I deleted the `response.headers['Content-Type'] ??` and ran the test, but it seems like the tests are failing when the mock server returns `Content-Type` header in upper case.

packages/functions_client/lib/functions_client.dart
```diff
@@ -1,4 +1,6 @@
 library functions_client;
 
+export 'package:http/http.dart' show ByteStream;
```
dshukertjr, line 3: Added this, because I thought people might want to type cast the `response.data` as `ByteStream`.

packages/functions_client/lib/src/functions_client.dart
Vinzent03: I still think the documentation is correct, but we just skip the translation when mocking the response object directly. So just keep it.

Vinzent03 (approved):
What about adding the missing support for the stream in web to the doc as well? I already see an issue coming in here...

nietsmmar:
> ## Additional Context
> On Flutter Web, the stream emits data only when the server is done sending the entire events. See the following issue for more info [dart-lang/http#337](https://github.com/dart-lang/http/issues/337)

Thanks for your effort!

But is there no possibility for it to work in web? I managed to make it work in web (with older gotrue/supabase-flutter version) using `sse_stream` and especially `fetch_client`. Could there be an option to use `fetch_client` as alternate `http` client as darts `HttpClient` does not support it yet. 

Maybe like it is done in here: https://github.com/pratikbaid3/flutter_client_sse/issues/19#issuecomment-1925033427

There there is the option to provide an `httpClientProvider`.

This would make it possible to use streaming in web too.

dshukertjr:
@nietsmmar Actually, just by passing fetch client, I was able to stream SSE on the web!
```dart
final client = FetchClient(mode: RequestMode.cors);
await Supabase.initialize(
  url: supabaseUrl,
  anonKey: supabaseKey,
  debug: false,
  httpClient: kIsWeb ? client : null,
);
```

PR #907, opened 2024-04-27
chore: Add comment about using SSE on the web

dshukertjr:

I meant to add this comment in https://github.com/supabase/supabase-flutter/pull/905, but I accidentally merged it before pushing it.

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -57,7 +57,7 @@ class FunctionsClient {
   ///       print(val);
   ///     });
   /// ```
-  /// To stream SSE on the web, you can use custom HTTP client that are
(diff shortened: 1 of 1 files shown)


packages/functions_client/lib/src/functions_client.dart
```diff
@@ -57,6 +57,16 @@ class FunctionsClient {
   ///       print(val);
   ///     });
   /// ```
+  /// To stream SSE on the web, you can use custom HTTP client that are
```
Vinzent03, line 60:
```suggestion
  /// To stream SSE on the web, you can use a custom HTTP client that is
```

Vinzent03 (changes requested):

Vinzent03 (approved):

PR #915, opened 2024-05-06
feat(supabase_flutter): Update app_links dependency to v6.0.1

flowhorn:

With app_links version 6.0.0 there are breaking changes in the app_links api ([https://pub.dev/packages/app_links/changelog](https://pub.dev/packages/app_links/changelog)):

`Breaking: Renamed getInitialApp* and getLatestApp* methods to getInitialLink* and getLatestLink*.`

I adjusted this in the PR and bumped the app_links package to ^6.0.1

## Additional context

I don't know whether you directly want to bump the app_links version, as there were multiple breaking version releases in the last 2 months. Just wanted to make this small PR as I personally bumped the version for myself.

### DIFF
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -189,7 +189,7 @@ class SupabaseAuth with WidgetsBindingObserver {
     _initialDeeplinkIsHandled = true;
 
     try {
-      final uri = await _appLinks.getInitialAppLink();
+      final uri = await _appLinks.getInitialLink();
       if (uri != null) {
         await _handleDeeplink(uri);
       }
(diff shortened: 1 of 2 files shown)


dshukertjr:
Yeah, this would be a breaking change on supabase_flutter for anyone using it with app_links, so we would wait until the next breaking change to introduce this update. 

Maybe we can explore handling deep links without app_links dependencies? https://docs.flutter.dev/ui/navigation/deep-linking

Vinzent03:
Can't we try to call the old methods with a try catch, and fallback to the new one? Definitely isn't the best way, but it should work.

dshukertjr:
@Vinzent03 I really don't think it's possible, but you have surprised me so many times in the past. Could you show me how that would work?

PR #918, opened 2024-05-06
fix: Keep compatible with app_links 6.0.0 and 5.0.0

Vinzent03:

Bug fix, dependency version bump

app_links 6.0.0 is not supported

## What is the new behavior?

5.0.0 is still supported and 6.0.0 now as well, despite the breaking change, by trying both methods with error catching.
Additionally, the supabase_flutter packages is now tested with downgraded packages versions as well.

## Additional context

#915 wanted to drop support for 5.0.0

### DIFF
--- a/.github/workflows/supabase_flutter.yml
+++ b/.github/workflows/supabase_flutter.yml
@@ -5,25 +5,25 @@ on:
     branches:
       - main
     paths:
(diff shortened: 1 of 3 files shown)


dshukertjr (approved):
Ah, casting it to dynamic! Nice trick! Thanks for this!

Issue #917, opened 2024-05-06
Matching on null fails to return anything


arc4randall:
**Describe the bug**
When performing a select, I would like to include the equivalent of "where columnName is null". There are two options for where clause in Supabase flutter:
1. eq: This doesn't accept optional or null values. So it's impossible to do

`Supabase.instance.client.from('my_table').select('column_one, column_two').eq('column_three, null);`

2. match: This accepts optional and null values but it doesn't actually work.

`Supabase.instance.client.from('my_table').select('column_one, column_two').eq('column_three, null);`

Returns 0 values even though running the equivalent query in Supabase SQL Editor returns results. Select eq with a value for column_three also returns all of the values. Only null returns no values.

**To Reproduce**
1. Setup a table with a column with nullable values
2. Setup a Flutter app
3. Attempt to select from the table
6. See 0 results

**Expected behavior**
I should be able to select where null. The workaround now is that I have to fetch all of the values and filter for null on the client.

**Version (please complete the following information):**
On Linux/macOS
Please run `dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"` in your project directory and paste the output here.

── supabase_flutter 2.5.2
│   ├── supabase 2.1.2
│   │   ├── functions_client 2.1.0
│   │   ├── gotrue 2.6.1
│   │   ├── postgrest 2.1.1
│   │   ├── realtime_client 2.0.4
│   │   ├── storage_client 2.0.1

**Additional context**

Vinzent03:
You should use the [isFilter](https://supabase.com/docs/reference/dart/is) for that case.

arc4randall:
> You should use the [isFilter](https://supabase.com/docs/reference/dart/is) for that case.

Thanks, this worked, but I don't think we should close this ticket because match accepts null values but they don't actually work.

dshukertjr:
Yeah, it might make sense to add an assersion for this.

arc4randall:
Sorry this actually is broken again. isFilter was working but now it is returning this error


```
code:
"PGRST100"
details:
"unexpected "P" expecting null or trilean value (unknown, true, false)"
hashCode:
738743739
hint:
null
message:
""failed to parse filter (is.Payments)" (line 1, column 4)"
```

In my case I am passing in a nullable string. The string contains the word "Payments" so that's why the error says that. There is now a bug with match as well as isFilter. I now have to resort to fetching all of the objects and doing a local filter.

[...]

PR #919, opened 2024-05-07
fix(postgrest): Update parameter type of `match()` filter so that null values cannot be passed. 

dshukertjr:

Fixes https://github.com/supabase/supabase-flutter/issues/917

Currently, users can pass `null` to `match()` filter. 

## What is the new behavior?

Users will not be able to pass `null` to `match()`.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_filter_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_filter_builder.dart
@@ -66,6 +66,8 @@ class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
   ///     .select()
   ///     .eq('username', 'supabot');
   /// ```
+  ///
+  /// If you want to filter with null equality, use [isFilter] instead.
   PostgrestFilterBuilder<T> eq(String column, Object value) {
     final Uri url;
     if (value is List) {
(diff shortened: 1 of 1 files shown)


bdlukaa (approved):

Vinzent03:
But the type of a map may be easily `Map<dynamic,dynamic>` despite having no null entries. So a working code might break now. I would prefer to note it in the documentation and use an assert over the actual values instead.

dshukertjr:
Fair enough.

packages/postgrest/lib/src/postgrest_filter_builder.dart
```diff
@@ -478,8 +480,10 @@ class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
   /// });
   /// ```
   PostgrestFilterBuilder<T> match(Map query) {
+    assert(!query.values.contains(null),
+        '`null` equality does not work with match. Use isFilter instead.');
     var url = _url;
-    query.forEach((k, v) => url = appendSearchParams('$k', 'eq.$v', url));
+    query.forEach((k, v) => url = appendSearchParams(k, 'eq.$v', url));
```
Vinzent03, line 486: Might be good to undo this.
```suggestion
    query.forEach((k, v) => url = appendSearchParams('$k', 'eq.$v', url));
```

Vinzent03 (changes requested):

bdlukaa:
I like the first proposed solution better. It gives the developers type-safety, causing less hassle. It might be good to have this shipped in the next major release.

dshukertjr:
>I like the first proposed solution better. It gives the developers type-safety, causing less hassle.

I like it too. I think sometimes we try to be a bit too picky about what defines a breaking change. I feel like the world does revolve around these minor "technically breaking changes" being shipped as a bug fix all the time. Flutter itself ships "breaking changes" without bumping the major versions all the time, and as long as it's under moderation, I don't think we should hesitate to ship these changes. 

I'd say we just ship it with the original change.

Vinzent03 (approved):
I don't think you can compare Flutter breaking changes with package breaking changes. Packages should use semver, Flutter doesn't.
I'm pretty sure three are code basis, which throw a static analysis error now, which definitely is a breaking change. But sure, our other bug fixes may sometimes be more breaking, and they  don't even show up when compiling.
I wouldn't make this breaking change, but you decide.

PR #920, opened 2024-05-08
fix(postgrest): Update parameter type of `isFilter()` to only allow boolean or null

dshukertjr:

Fixes https://github.com/supabase/supabase-flutter/issues/917

Currently any object can be passed as the value parameter of `isFilter()`

## What is the new behavior?

Only null and boolean values can be passed to `isFilter()`


Vinzent03 (approved):
Same breaking change warning as in #919 Just to note for the future.

PR #922, opened 2024-05-09
feat(goture): Allow OAuthProvider.kakao for signInWithIdToken

nectarine:

Allow OAuthProvider.kakao for signInWithIdToken in gotrue_client

Only google and apple are allowed.

## What is the new behavior?

https://github.com/supabase/auth/pull/1381
Since supabase.auth already allows an OIDC with kakao, the client should accept the OAuthProvider.kakao with signInWithIdToken


dshukertjr (approved):
Thanks for this! It will be included in the next release, which should happen in the next few days.

PR #926, opened 2024-05-13
feat(functions): Invoke function with custom query params

### LINKED ISSUE #924: Allow to pass query params when invoking function
When invoking a function I want to be able to pass custom query params.

**Describe the solution you'd like**
```dart
final res = await supabase.functions
    .invoke(
        'hello',
        query: { 'key' : 'value' }
    )
```

We should be able to edit the Uri [here](https://github.com/supabase/supabase-flutter/blob/main/packages/functions_client/lib/src/functions_client.dart#L78) to include the new query parameters accepted on the `invoke()` method.

**Additional context**
https://github.com/supabase/supabase-swift/pull/376

Cteq3132:

Adding query parameters support for edge functions. 
Closes #924

## Additional context

Largely inspired from https://github.com/supabase/supabase-swift/pull/376.

cc @dshukertjr

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -71,11 +71,13 @@ class FunctionsClient {
     String functionName, {
     Map<String, String>? headers,
     Map<String, dynamic>? body,
+    Map<String, String>? queryParameters,
     HttpMethod method = HttpMethod.post,
   }) async {
     final bodyStr = body == null ? null : await _isolate.encode(body);
 
-    final uri = Uri.parse('$_url/$functionName');
+    final uri = Uri.parse('$_url/$functionName')
+        .replace(queryParameters: queryParameters);
 
     final finalHeaders = <String, String>{
       ..._headers,
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -71,7 +71,7 @@ class FunctionsClient {
     String functionName, {
     Map<String, String>? headers,
     Map<String, dynamic>? body,
-    Map<String, String>? queryParameters,
+    Map<String, dynamic>? queryParameters,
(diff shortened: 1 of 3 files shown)


packages/functions_client/lib/src/functions_client.dart
```diff
@@ -71,11 +71,13 @@ class FunctionsClient {
     String functionName, {
     Map<String, String>? headers,
     Map<String, dynamic>? body,
+    Map<String, String>? queryParameters,
```
dshukertjr, line 74: Should we align the naming here with the Swift library and call it `query` instead of `queryParameters`?

dshukertjr (commented):
Thanks for this PR and for adding tests for it as well! I just had one comment regarding the naming of the parameter, but other than that it looks great!

packages/functions_client/lib/src/functions_client.dart
Vinzent03, line 74: From the [Uri constructor documentation](https://api.dart.dev/stable/3.3.4/dart-core/Uri/Uri.html)
> A value in the map must be either null, a string, or an [Iterable](https://api.dart.dev/stable/3.3.4/dart-core/Iterable-class.html) of strings. An iterable corresponds to multiple values for the same key, and an empty iterable or null corresponds to no value for the key.

So I think we could adapt the same type and use `Map<String,dynamic>` instead of `Map<String,String>` for the query parameters.

Vinzent03: I think we should keep the wording of the dart ecosystem, which uses `queryParameters` in the `Uri` class for a map and `query` for the string representation.

Cteq3132: Yes that's a good point, I'm updating this part right now.

dshukertjr: Sure, fair enough.

PR #932, opened 2024-05-16
feat: Allow setting `timeout` of `RealtimeClient`.

nns52k:

Allow the ctor `SupabaseClient` to set `timeout` to a `RealtimeClient`.

Due to the fact the `timeout` is a final field in `RealtimeClient`, we cannot change its value after the `RealtimeClient` is created, and because the `RealtimeClient` is created in the ctor of `SupabaseClient`, there is no way for users to set the timeout.

## What is the new behavior?

The new behavior allow us to set the timeout of the `RealtimeClient` like this:
``` dart
final SupabaseClient supabase = SupabaseClient(
  'https://your.supabase.address/',
  'your_anon_key',
  realtimeClientOptions: RealtimeClientOptions(timeout: Duration(minutes: 1)),
);
```

## Additional context

My app is connecting Supabase through Tor proxy. It's normal to cost 10 or more seconds to build a WebSocket connection. However, the current default timeout of `RealtimeClient` is 10 seconds, which is too short, and very often leads to result of no WebSocket connection is created, and therefore no realtime feature is provided.

### DIFF
--- a/packages/realtime_client/lib/realtime_client.dart
+++ b/packages/realtime_client/lib/realtime_client.dart
@@ -1,4 +1,4 @@
-export 'src/constants.dart' show RealtimeLogLevel;
+export 'src/constants.dart' show RealtimeConstants, RealtimeLogLevel;
 export 'src/realtime_channel.dart';
 export 'src/realtime_client.dart';
 export 'src/realtime_presence.dart';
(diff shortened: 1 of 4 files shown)


dshukertjr (approved):
This is awesome! Thanks for the update.

PR #945, opened 2024-06-08
fix(gotrue): Signing in does not remove the session unless the operation succeeds.

dshukertjr:

Bug fix

Currently, the SDK removes the session whenever a user tries to sign-in / sign-up. If the operation fails, the user is signed out from the original session they had. This PR fixes the behavior so that the session is not removed.

## What is the new behavior?

The user's session is not removed at the beginning of calling sign-in / sign-up methods, and they will keep the existing session if the sign-in / sign-up operation fails. 

## Additional context

JS PR: https://github.com/supabase/auth-js/pull/915

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -140,8 +140,6 @@ class GoTrueClient {
     Map<String, dynamic>? data,
     String? captchaToken,
   }) async {
-    _removeSession();
-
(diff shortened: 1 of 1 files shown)


kangmingtay (approved):

dshukertjr:
Just noticed that there are a bunch of errors on the tests. Looking into it right now.

Vinzent03 (approved):

Issue #937, opened 2024-05-28
Handling initial deeplink in password recovery


bruiken:
**Is your feature request related to a problem? Please describe.**
When implementing the password recovery flow, we noticed that the initial deeplink handling does not yield for much customisability. The function ` _handleInitialUri()` is called as await from the initialisation function of the `Supabase` client. 
If that initial deeplink is concerning a password recovery, then it seems to be impossible to catch that event in the `supabase.auth.onAuthStateChange` listener. The same holds for the error cases, which are never even brought to that stream (this has been pointed out by #664 before too). Yet in that case it wasn't about a password recovery but a magic link.
For magic links this all is not that big of an issue, at the end of the day the user is authenticated and there is a session. For password recovery it to us seems impossible to direct the user to the "reset your password" screen when the app was fully closed.

**Describe the solution you'd like**
We'd like to see a solution where it is made possible to detect password recovery and possible errors from initial deeplink handling. A solution in this most definitely extends to any event, not just password recovery. 

**Describe alternatives you've considered**
- We could use the new functionality to block the deeplink handling by supabase and do it all manually. 
- We could not support the case where the user force closes the app after pressing reset password.

**Additional context**
Besides this, it's also quite difficult to catch errors specific to deeplink handling, our code now contains a number of checks based on the `AuthException.message` field. It would seem like a good idea to create more different error subclasses of `AuthException` so it is possible to do checks that are not reliant on error message changes. 

If directions are given for how to possibly fix this, I'd be okay with submitting a PR. One possible avenue could be to allow giving the `supabase.auth.onAuthStateChange` in the `Supabase.initialize` function. I don't think that would be nice behaviour as listeners in different places get access to different events. Maybe a more specific event handler for the initial deeplink handling? Or then some way to get access to past events.

bruiken:
After a bit further investigation in how we can do this, it seems this issue could be resolved by using `AppLinks.allUriLinkStream` instead of `AppLinks.uriLinkStream`. The former includes the initial link. This is different from the current implementation because this initial link will not be awaited and can thus be caught by our listeners.

A different solution for us would be if the deeplink integration had a bit more customisability. Allowing e.g. an enum option such as `enum DeeplinkHandlingOption { all, none, nonInitial }`. I don't really like this because this would just be duplicate implementation.

bruiken:
[...]

PR #965, opened 2024-07-05
fix: make auth state change callbacks available in initialisation options (and fix duplicate initial deeplink handling)

bruiken:

This change allows giving callbacks for both data and errors from auth state changes. These listeners are applied in the initialisation of `SupabaseAuth`. 

In practice, this allows getting state change updates that happen _during_ the initialisation of Supabase.

Secondly, this fixes handling the initial deeplink multiple times when using AppLinks >= 6.0.0. In that version a change was made such that the initial deeplink is also exposed in the subscription that was already in place here.

It is currently not possible to get updates from the subscription during the initialisation phase. Trying to assign a listener to the subscription fails because initialisation has not happened yet. 

See #937 for elaborate information and what issue this actually causes: failing to observe `passwordRecovery` events from the initial deeplink. 

## What is the new behavior?

This adds options for adding listeners to the initialisation, there already was a subscription, now the handlers for both `onData` and `onError` also optionally call those listeners. This is commit 1fc399e

To solve the solution of duplicate initial deeplink handling, we can simply remove the handling when we know that we are using AppLinks >= 6.0.0. This is commit 494f371

## Additional context

I don't think this is the cleanest solution, but also didn't see a good way of getting a listener attached this early in the process. I especially dislike that there are now multiple ways you can attach such as listener. However, I think disallowing the original subscribing solution is bad because, (1) it is a breaking change and (2) it seems like a legitimate use case to me to subscribe after the initialisation phase, depending on what people are trying to do.

### DIFF
--- a/packages/supabase_flutter/lib/src/flutter_go_true_client_options.dart
+++ b/packages/supabase_flutter/lib/src/flutter_go_true_client_options.dart
@@ -3,6 +3,13 @@ import 'package:supabase_flutter/supabase_flutter.dart';
 class FlutterAuthClientOptions extends AuthClientOptions {
   final LocalStorage? localStorage;
 
+  /// Callback for receiving state change updates from the auth subscription.
+  final void Function(AuthState data)? onAuthStateChange;
+
+  /// Callback for receiving errors from the auth state change subscription.
+  /// Both the error and the stacktrace will be given as arguments.
+  final Function? onAuthError;
+
   /// If true, the client will start the deep link observer and obtain sessions
   /// when a valid URI is detected.
   final bool detectSessionInUri;
@@ -12,6 +19,8 @@ class FlutterAuthClientOptions extends AuthClientOptions {
     super.autoRefreshToken,
     super.pkceAsyncStorage,
     this.localStorage,
+    this.onAuthStateChange,
+    this.onAuthError,
     this.detectSessionInUri = true,
   });
 
@@ -21,6 +30,8 @@ class FlutterAuthClientOptions extends AuthClientOptions {
     LocalStorage? localStorage,
     GotrueAsyncStorage? pkceAsyncStorage,
     bool? detectSessionInUri,
+    void Function(AuthState data)? onAuthStateChange,
+    Function? onAuthError,
   }) {
     return FlutterAuthClientOptions(
       authFlowType: authFlowType ?? this.authFlowType,
@@ -28,6 +39,8 @@ class FlutterAuthClientOptions extends AuthClientOptions {
       localStorage: localStorage ?? this.localStorage,
       pkceAsyncStorage: pkceAsyncStorage ?? this.pkceAsyncStorage,
       detectSessionInUri: detectSessionInUri ?? this.detectSessionInUri,
+      onAuthStateChange: onAuthStateChange ?? this.onAuthStateChange,
+      onAuthError: onAuthError ?? this.onAuthError,
     );
   }
 }
--- a/packages/supabase_flutter/lib/src/supabase_auth.dart
+++ b/packages/supabase_flutter/lib/src/supabase_auth.dart
@@ -194,10 +194,12 @@ class SupabaseAuth with WidgetsBindingObserver {
         // before app_links 6.0.0
         uri = await (_appLinks as dynamic).getInitialAppLink();
       } on NoSuchMethodError catch (_) {
-        // Needed to keep compatible with 5.0.0 and 6.0.0
+        // The AppLinks package contains the initial link in the uriLinkStream
+        // starting from version 6.0.0. Before this version, getting the
+        // initial link was done with getInitialAppLink. Being in this catch
+        // handler means we are in at least version 6.0.0, meaning we do not
+        // need to handle the initial link manually.
         // https://pub.dev/packages/app_links/changelog
-        // after app_links 6.0.0
-        uri = await (_appLinks as dynamic).getInitialLink();
       }
       if (uri != null) {
         await _handleDeeplink(uri);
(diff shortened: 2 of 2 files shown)


packages/supabase_flutter/lib/src/flutter_go_true_client_options.dart
```diff
@@ -3,6 +3,13 @@ import 'package:supabase_flutter/supabase_flutter.dart';
 class FlutterAuthClientOptions extends AuthClientOptions {
   final LocalStorage? localStorage;
 
+  /// Callback for receiving state change updates from the auth subscription.
+  final void Function(AuthState data)? onAuthStateChange;
+
+  /// Callback for receiving errors from the auth state change subscription.
+  /// Both the error and the stacktrace will be given as arguments.
+  final Function? onAuthError;
+
```
dshukertjr, line 12: I don't think we want to introduce these. Let's remove them.

dshukertjr (commented):
Thanks for the PR. Let's keep the PR lean and focus on fixing the way this library handles the deep links.

packages/supabase_flutter/lib/src/flutter_go_true_client_options.dart
bruiken: Would you be okay with adding these in a different scope? The whole problem with the deeplink handling is the impossibility to catch the `passwordRecovery` event. To make this possible we need to have a listener before these events are pushed to the auth stream. 

This is an attempt to do so, as these listeners callbacks will be called even during initialization (= the phase where events like `passwordRecovery` are emitted).

dshukertjr: As long as we remove the duplicate initial session that gets fired after password recovery event, we should be good.

bruiken: Okay I'll revert the last commit! Do you have any tips on how to get the tests working again? I see quite a nice solution with mocking the method channels. Yet in AppLinks 6.0.0 the initial link handling turned into platform specific implementations, removing the ability to mock this in Dart 🤔

dshukertjr: Is it possible to create multiple mocks and test cases that mocks the behavior of web and non-web platforms?

bruiken: Hmm don't think so as the difference in usage is only dependent on the app_links package version. Tried to address this problem in latest commit, mocking both the event channel and the method channel with the initial link.

bruiken:
The test for v3.10.x now fails because of an undefined method, didn't inspect further yet.

Test for v3.x fails because it uses app_links <6.0.0, which is allowed by the pubspec. Could be fixed with the same hack in the codebase, checking function availability that causes an exception in versions >=6.0.0. Could also be fixed by allowing the count of the function call to be >=1, but that would be bad because we truly expect it only once.

packages/supabase_flutter/lib/src/flutter_go_true_client_options.dart
bruiken: Ending this comment chain here, not relevant to these changes

bruiken:
[7d12a9f](https://github.com/supabase/supabase-flutter/pull/965/commits/7d12a9fc12ea93f9728e986d1fb8faaee8206600) should fix tests for newest flutter, still haven't looked into fixing v3.10.x tests.

bruiken:
Commit 81d7314 hopefully fixes it for the older versions of flutter too, I now start seeing why certain code is written in certain ways, to keep supporting v3.10.*. 

Bit hard to test locally, any tips for that?

packages/supabase_flutter/test/supabase_flutter_test.dart
```diff
@@ -93,11 +94,32 @@ void main() {
...
+    /// can find the getInitialAppLink function.
+    bool appLinksExposesInitialLinkInStream() {
+      try {
+        // before app_links 6.0.0
+        (AppLinks() as dynamic).getInitialAppLink;
+        return false;
+      } on NoSuchMethodError catch (_) {
+        return true;
+      }
+    }
```
dshukertjr, line 110: Instead of checking the current environment to see if app_links 6.0.0 is used or not, could we create two test cases, one where the behavior of app_links v6.0.0 is used, and one where app_link version prior to that is used?

dshukertjr (commented):
Thanks for the amazing work on this PR! Just had one comment to improve the test, but it looks awesome!

packages/supabase_flutter/test/supabase_flutter_test.dart
bruiken: The GH Actions are already set up in a way that the tests are run both for the new app_links version and for the old one.

What could be possible is to split up the test cases and have one of them be skipped based on the environment (app_links version). We could then still use this same logic to decide which one is to be skipped and which one to be run. That might make a little more sense than having one test that does different things based on the environment. 

Would that solve it for you?

dshukertjr: Could you elaborate on how GitHub actions are already set up in a way that tests are for both of the new and old app_links versions?

dshukertjr: Ah, alright! Could we maybe keep a comment mentioning that the CI is setup to test new and old versions of app_links then so that it's obvious?

packages/supabase_flutter/test/supabase_flutter_test.dart
```diff
@@ -93,11 +94,32 @@ void main() {
 
   group('Deep Link with PKCE code', () {
     late final PkceHttpClient pkceHttpClient;
+    late final bool mockEventChannel;
+
+    /// Check if the current version of AppLinks uses an explicit call to get
+    /// the initial link. This is only the case before version 6.0.0, where we
+    /// can find the getInitialAppLink function.
+    bool appLinksExposesInitialLinkInStream() {
```
dshukertjr, line 102: Something like this?
```suggestion
    /// can find the getInitialAppLink function.
    /// 
    /// CI pipeline is set so that it tests both app_links newer and older than v6.0.0
    bool appLinksExposesInitialLinkInStream() {
```

dshukertjr (approved):
Thanks for the amazing work!

Issue #969, opened 2024-07-08
Unchecked null access in AuthHttpClient


AtjonTV:
**Describe the bug**
Sentry caught this exception right after an `AuthApiException`, so seemingly during a session refresh.

**To Reproduce**
I am not sure how this can be reproduced, but this exact error in the same constellation was captured twice within 9 days.

The first was on 28th June, second on 7th July.

**Expected behavior**
Graceful handling of possibly-null variables.

**Screenshots**
Failed token refresh:
```
AuthApiException: AuthException(message: Invalid Refresh Token: Already Used, statusCode: 400)
  File "gotrue_client.dart", line 1206, in GoTrueClient.notifyException
  File "gotrue_client.dart", line 995, in GoTrueClient.recoverSession
  File "<asynchronous suspension>"
  File "supabase_auth.dart", line 90, in SupabaseAuth.recoverSession
  File "<asynchronous suspension>"
  File "cancelable_operation.dart", line 425, in CancelableCompleter.complete.<fn>
```

Followed by:
```
_TypeError: Null check operator used on a null value
  File "auth_http_client.dart", line 20, in AuthHttpClient.send
  File "<asynchronous suspension>"
  File "base_client.dart", line 93, in BaseClient._sendUnstreamed
  File "<asynchronous suspension>"
  File "postgrest_builder.dart", line 130, in PostgrestBuilder._execute
  File "<asynchronous suspension>"
  File "postgrest_builder.dart", line 372, in PostgrestBuilder.then
```

**Version (please complete the following information):**
```
├── supabase_flutter 2.5.2
│   ├── supabase 2.1.2
│   │   ├── functions_client 2.1.0
│   │   ├── gotrue 2.6.1
│   │   ├── postgrest 2.1.1
│   │   ├── realtime_client 2.0.4
│   │   ├── storage_client 2.0.1
```

**Additional context**
I know that the app does not have the latest versions of the dependencies, but the failing code still exists on top-of-main: https://github.com/supabase/supabase-flutter/blob/c65cb6123639d38b2f7fa69b009f29baa8926b01/packages/supabase/lib/src/auth_http_client.dart#L18C23-L22C45

themightychris:
I always see this the first time I run my app after installing from an APK downloaded from the web. Doesn't happen if I install the same APK via command line or on subsequent runs.

themightychris:
[...]

PR #976, opened 2024-07-19
fix: Add error messages for when the session or expiresAt was missing while making API requests. 

dshukertjr:

This PR  properly handles null session and null `expiresAt` for a slim chance of them being null while making HTTP requests. 

Closes https://github.com/supabase/supabase-flutter/issues/969

### DIFF
--- a/packages/supabase/lib/src/auth_http_client.dart
+++ b/packages/supabase/lib/src/auth_http_client.dart
@@ -14,23 +14,15 @@ class AuthHttpClient extends BaseClient {
       try {
         await _auth.refreshSession();
       } catch (error) {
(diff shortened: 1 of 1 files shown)


Vinzent03 (commented):
The session being null in the catch case mainly happens because the session get's removed in `refreshSession`, right?.(And it could in rare cases happen, because of parallelization). You now force them to exists in the catch case. This causes no request to happen, although they might don't even need a session. Isn't it better to send the request even with no session?

dshukertjr:
@Vinzent03 Yeah, good call.  Let's do that.

Vinzent03 (approved):

PR #977, opened 2024-07-20
feat: Support MultipartRequest in functions invoke

### LINKED ISSUE #933: Does Supabase Flutter `functions.invoke` support FormData input?

Does the Flutter client's `functions.invoke` method support FormData as input? My understanding is that it currently does not, since:

https://github.com/supabase/supabase-flutter/blob/408e816ac66442eda90db3e83900b05c10dd604f/packages/functions_client/lib/src/functions_client.dart#L87

would need to be

```dart
http.MultipartRequest
```

instead. I could be wrong though. If the current Flutter client already supports FormData, can someone show me an example? I am mostly considering this example on Edge Function file upload: https://github.com/orgs/supabase/discussions/11123. I can see how it is being done in JS/TS since the JS client can handle FormData.

I have been using `Dio.post` as a workaround, but it would be nice to keep everything within the Supabase client.

Thank you.

Vinzent03:

feature

The `body` can only be of type `Map<String,dynamic>` and no files can be uploaded.

## What is the new behavior?

The `body` can be any `Object`, which is encodable to String. There's now a `files` parameter to send files with a `MultipartRequest`.

## Additional context

At the moment everything is bundled in one method like in functions-js. One could argue to create an extra method to be more type safe with `body` being `Map<String,String>`.

close #933

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'package:functions_client/src/constants.dart';
 import 'package:functions_client/src/types.dart';
 import 'package:http/http.dart' as http;
+import 'package:http/http.dart' show MultipartRequest;
 import 'package:yet_another_json_isolate/yet_another_json_isolate.dart';
 
 class FunctionsClient {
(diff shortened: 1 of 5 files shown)


dshukertjr (commented):
This is great! We probably need to export `MultipartFile` so that it's available to the users, right?

Vinzent03:
It's from the http package, which I guess most people have already installed and can import themself. But exporting just that class won't hurt, so I will add that tomorrow.

packages/functions_client/lib/src/functions_client.dart
```diff
@@ -84,12 +90,36 @@ class FunctionsClient {
...
+    } else {
+      final bodyRequest = http.Request(method.name, uri);
+
+      final String? bodyStr;
+      if (body == null) {
+        bodyStr = null;
+      } else if (body is String) {
+        bodyStr = body;
+      } else {
+        bodyStr = await _isolate.encode(body);
```
bdlukaa, line 113: It'd be good to add tests for `body` with different types, like `int`, `double`, `Map` and `List`.

dshukertjr:
>It's from the http package, which I guess most people have already installed and can import themself. But exporting just that class won't hurt, so I will add that tomorrow.

Awesome 👍 Just sharing my two cents, but I think it's common for people to not have the HTTP package installed if they are using Supabase.

bdlukaa (approved):

Vinzent03:
Since more types for the body are allowed now, I've added a basic content header detection. For string inputs, it sets the header to text/plain and for `Uint8List` as an example for binary data to octet stream.

dshukertjr (approved):
Love the automatic content type detection!

PR #985, opened 2024-07-23
fix(functions_client): Add `toString` to `FunctionException`

dshukertjr:

Add `toString` method to FunctionException so that the exception would show some context, hence making it easier to debug.

### DIFF
--- a/packages/functions_client/lib/src/types.dart
+++ b/packages/functions_client/lib/src/types.dart
@@ -32,7 +32,7 @@ class FunctionException {
   final dynamic details;
   final String? reasonPhrase;
 
-  FunctionException({required this.status, this.details, this.reasonPhrase});
+  const FunctionException({required this.status, this.details, this.reasonPhrase});
 
   @override
   String toString() =>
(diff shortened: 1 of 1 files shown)


packages/functions_client/lib/src/types.dart
```diff
@@ -32,4 +33,8 @@ class FunctionException {
   final String? reasonPhrase;
 
   FunctionException({required this.status, this.details, this.reasonPhrase});
```
bdlukaa, line 35: This could be `const`.
```suggestion
  const FunctionException({required this.status, this.details, this.reasonPhrase});
```

packages/functions_client/lib/src/types.dart
```diff
@@ -1,3 +1,4 @@
+// ignore_for_file: public_member_api_docs, sort_constructors_first
```
Vinzent03, line 1: What's the reason for these ignores here?

dshukertjr: Sorry, not sure how it slipped in. Deleting it!

Vinzent03 (approved):

bdlukaa (approved):

Issue #989, opened 2024-07-26
Object cannot be assigned to the parameter type Map<dynamic, dynamic>


PcolBP:
**Describe the bug**
After newest version 2.3.0 I cannot build my web application, had to force downgrade it to 2.2.0 since Supabase uses automatically 2.3.0

**To Reproduce**
Build Web Application

**Expected behavior**
```
Error: The argument type 'Object' can't be assigned to the parameter type 'Map<dynamic, dynamic>'.
 - 'Object' is from 'dart:core'.
 - 'Map' is from 'dart:core'.
        bodyStr = await _isolate.encode(body);
```

**Version (please complete the following information):**
On Linux/macOS
Please run `dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"` in your project directory and paste the output here.
```console
dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"
│   ├── functions_client 2.2.0
│   ├── supabase_flutter 2.5.1
│   │   ├── supabase 2.2.5
│   │   │   ├── functions_client...
│   │   │   ├── gotrue 2.8.3
│   │   │   ├── postgrest 2.1.2
│   │   │   ├── realtime_client 2.2.0
│   │   │   ├── storage_client 2.0.2
```

**Additional context**
Flutter Doctor
```console
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.22.3, on macOS 14.5 23F79 darwin-arm64 (Rosetta), locale pl-PL)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 15.0.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2024.1)
[✓] VS Code (version 1.91.1)
[✓] Connected device (4 available)
[✓] Network resources

• No issues found!
```

dshukertjr:
I can confirm this happening. Will ship a fix ASAP. Thanks for reporting.

dshukertjr:
For now, the latest version has been retracted.

PR #990, opened 2024-07-26
fix: Make sure the package can be built on Flutter web

dshukertjr:

Fixes an issue where the package is not able to build on Flutter web

Closes https://github.com/supabase/supabase-flutter/issues/989

### DIFF
--- a/packages/supabase_flutter/example/pubspec.yaml
+++ b/packages/supabase_flutter/example/pubspec.yaml
@@ -21,21 +21,3 @@ dev_dependencies:
 
 flutter:
   uses-material-design: true
-melos_managed_dependency_overrides: functions_client,gotrue,postgrest,realtime_client,storage_client,yet_another_json_isolate
-dependency_overrides:
-  functions_client:
-    path: ../../functions_client
-  gotrue:
-    path: ../../gotrue
-  postgrest:
-    path: ../../postgrest
-  realtime_client:
-    path: ../../realtime_client
-  storage_client:
-    path: ../../storage_client
-  supabase:
-    path: ../../supabase
-  yet_another_json_isolate:
-    path: ../../yet_another_json_isolate
-  supabase_flutter:
-    path: ../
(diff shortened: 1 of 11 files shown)


dshukertjr:
Once we publish a new yet_another_json_isolate with this update, the currently retracted functions_client v2.3.0 will be restored.

PcolBP (approved):

packages/supabase_flutter/example/pubspec.yaml
```diff
@@ -21,3 +21,21 @@ dev_dependencies:
 
 flutter:
   uses-material-design: true
+melos_managed_dependency_overrides: functions_client,gotrue,postgrest,realtime_client,storage_client,yet_another_json_isolate
```
Vinzent03, line 24: I'm unsure about these changes here. Did you run any command to produce these lines? Is there no way to include the example in the bootstrap to avoid modifying the overrides in the pubspec.yaml?

dshukertjr: What is the concern? For now, this solves our issue, ensuring that the package works on Flutter web. I can see that this can cause slight inconvenience for anyone viewing the example directly, but that's not the top priority issue that we have right now, and we can take our time to figure out if there is a better way of overriding the dependencies.

dshukertjr: Okay, I have updated the melos config instead of the pubspec.yaml file of the example.

dshukertjr: And of course now the tests are failing 😂

Vinzent03: I wouldn't add the packages to the bootstrap section. Just the oneline under packages

dshukertjr: The CI is probably failing, because pub get for the example app is running before some of the other packages. I have to run now for a bit, but will take a look when I come back.

Vinzent03: The issue was, that melso tried to use the flutter cli, because the example is a flutter package and was not excluded like the supabase_flutter package everywhere manually.

Vinzent03 (approved):

dshukertjr:
Nice! Thanks for the fix @Vinzent03 !

PR #992, opened 2024-07-29
fix: Upgrade `web_socket_channel` for supporting `web: ^1.0.0` and therefore WASM compilation on web

geoextra:

Upgrade the  `web_socket_channel` to `^3.0.1` which mostly serves as an upgrade path for the transitive `web` dependency which otherwise causes `web_socket_channel` to downgrade and prevent WASM support on the web platform.

When having `supabase` or `supabase_flutter` as a project dependeny while relying on `web: ^1.0.0` the  `web_socket_channel` dependeny would downgrade and WASM compilation isn't possible anymore.

## What is the new behavior?

This behavior seems to be migrated.

## Additional context

The breaking changes don't affect the codebase. Tests passed locally. Additionally, one linting error got fixed.

### DIFF
--- a/packages/supabase/lib/src/auth_user.dart
+++ b/packages/supabase/lib/src/auth_user.dart
@@ -2,33 +2,18 @@ import 'package:gotrue/gotrue.dart' show User;
 
 class AuthUser extends User {
   AuthUser({
-    required String id,
-    required Map<String, dynamic> appMetadata,
-    required Map<String, dynamic> userMetadata,
-    required String aud,
-    required String? email,
-    required String? phone,
-    required String createdAt,
-    String? confirmedAt,
-    String? emailConfirmedAt,
-    String? phoneConfirmedAt,
-    String? lastSignInAt,
-    required String role,
-    required String updatedAt,
-  }) : super(
-          id: id,
-          appMetadata: appMetadata,
-          userMetadata: userMetadata,
-          aud: aud,
-          email: email,
-          phone: phone,
-          createdAt: createdAt,
-          // ignore: deprecated_member_use
-          confirmedAt: confirmedAt,
-          emailConfirmedAt: emailConfirmedAt,
-          phoneConfirmedAt: phoneConfirmedAt,
-          lastSignInAt: lastSignInAt,
-          role: role,
-          updatedAt: updatedAt,
-        );
+    required super.id,
+    required super.appMetadata,
+    required Map<String, dynamic> super.userMetadata,
+    required super.aud,
+    required super.email,
+    required super.phone,
+    required super.createdAt,
+    super.confirmedAt,
+    super.emailConfirmedAt,
+    super.phoneConfirmedAt,
+    super.lastSignInAt,
+    required String super.role,
+    required String super.updatedAt,
+  });
 }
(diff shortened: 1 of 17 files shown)


dshukertjr (changes requested):
Thanks for this PR! 

Can we update the `web_socket_channel` dependencies to `'>=2.3.0 <4.0.0'` so that this doesn't become a breaking change?

geoextra:
Seems like I was a bit too fast with upgrading the linting dependecy which are not compatible with Dart version < 3.1 and fails tests, I will adjust to a non-breaking version.

packages/supabase/lib/src/auth_user.dart
```diff
@@ -2,33 +2,18 @@ import 'package:gotrue/gotrue.dart' show User;
...
-          confirmedAt: confirmedAt,
-          emailConfirmedAt: emailConfirmedAt,
-          phoneConfirmedAt: phoneConfirmedAt,
-          lastSignInAt: lastSignInAt,
-          role: role,
-          updatedAt: updatedAt,
-        );
+    required super.id,
+    required super.appMetadata,
+    required Map<String, dynamic> super.userMetadata,
```
dshukertjr, line 7:
```suggestion
    required super.userMetadata,
```

packages/supabase/pubspec.yaml
```diff
@@ -19,6 +19,6 @@ dependencies:
   yet_another_json_isolate: 2.0.1
 
 dev_dependencies:
-  lints: ^2.1.1
+  lints: ^3.0.0
   test: ^1.17.9
-  web_socket_channel: ^2.2.0
+  web_socket_channel: '>=2.3.0 <4.0.0'
```
dshukertjr, line 24:
```suggestion
  web_socket_channel: '>=2.2.0 <4.0.0'
```

packages/supabase/lib/src/supabase_query_builder.dart
```diff
@@ -11,22 +10,18 @@ class SupabaseQueryBuilder extends PostgrestQueryBuilder {
   SupabaseQueryBuilder(
     String url,
     RealtimeClient realtime, {
-    Map<String, String> headers = const {},
-    required String schema,
+    Map<String, String> super.headers = const {},
+    required String super.schema,
```
dshukertjr, line 14:
```suggestion
    super.headers = const {},
    required super.schema,
```

packages/supabase/lib/src/supabase_query_builder.dart
```diff
@@ -11,22 +10,18 @@ class SupabaseQueryBuilder extends PostgrestQueryBuilder {
...
     required String table,
-    Client? httpClient,
+    super.httpClient,
     required int incrementId,
-    required YAJsonIsolate isolate,
+    required YAJsonIsolate super.isolate,
```
dshukertjr, line 18:
```suggestion
    required super.isolate,
```

packages/supabase/lib/src/auth_user.dart
```diff
@@ -2,33 +2,18 @@ import 'package:gotrue/gotrue.dart' show User;
...
+    required super.aud,
+    required super.email,
+    required super.phone,
+    required super.createdAt,
+    super.confirmedAt,
+    super.emailConfirmedAt,
+    super.phoneConfirmedAt,
+    super.lastSignInAt,
+    required String super.role,
+    required String super.updatedAt,
```
dshukertjr, line 17:
```suggestion
    required super.role,
    required super.updatedAt,
```

dshukertjr (approved):
Thanks for this update! This will be included in the next release in a few days!

PR #995, opened 2024-08-02
fix: Added missing error codes for AuthException

lawinski:

Add support for auth error codes

There are no error code in auth exception

## What is the new behavior?

Error codes added to AuthException


dshukertjr:
Thanks for getting started on this one! Here is the JS PR for the same feature. If you could match the implementation as much as possible, that would be great!
https://github.com/supabase/auth-js/pull/855

lawinski:
Applied adjustments to the logic from JS implementation:
- Modified the version management by wrapping the logic into a dedicated class and storing API versions in another. This allows for more type-safe access and consolidates all logic in one place. The core logic remains unchanged.
- Added tests for GotrueFetch to cover different responses (MockedHttpClient added for this purpose).
- Changed AuthErrorCode to ErrorCode.

Doubts:
- Do not introduce exceptions that do not already exist (e.g., AuthInvalidTokenResponseError, AuthImplicitGrantRedirectError, AuthInvalidCredentialsError).
- Setting up a default version header if it does not exist in the request method prevents the user from using the initial API version until a valid one is provided. I’m unsure if this is a good assumption, but I guess you have already considered this. :)

PS. If the logic should match the JS implementation exactly, please let me know. If we are heading in that direction, I would appreciate guidance on the process of implementing new functionalities. Does this mean we always have to wait for the implementation in JS first?

packages/gotrue/lib/src/fetch.dart
```diff
@@ -28,9 +31,12 @@ class GotrueFetch {
...
   }
 
-  String? _getErrorCode(dynamic error) {
+  String? _getErrorCode(dynamic error, String key) {
     if (error is Map) {
-      return error['error_code']?.toString();
+      final dynamic errorCode = error[key];
+      if (errorCode is String) {
+        return errorCode;
+      }
```
bdlukaa, line 39: Would there be any cases where the error code is a number? HTTP Error 400, for example.

lawinski: It should not.

For the old response, the `error_code` is declared as a `string`:
- https://github.com/supabase/auth/blob/701a779cf092e777dd4ad4954dc650164b09ab32/internal/api/errors.go#L103

For the 2024-01-01 version of the API, it will be an `ErrorCode`, which is also a `string`:
- https://github.com/supabase/auth/blob/701a779cf092e777dd4ad4954dc650164b09ab32/internal/api/errors.go#L182

Anyway, I could be missing something :)

dshukertjr (commented):
@lawinski Thanks for the amazing work! This PR is awesome! I edited the code in a way that I think would be more aligned with the JS SDK and might be easier to maintain, but would love to hear what you think!

> If the logic should match the JS implementation exactly, please let me know. If we are heading in that direction, I would appreciate guidance on the process of implementing new functionalities. 

Generally yes. The closer the implementation is to the JS SDK, the easier to maintain, and also could avoid having to deviate the SDK from the JS one because of future changes that were not anticipated. 

>Does this mean we always have to wait for the implementation in JS first?

Also generally yes.

packages/gotrue/lib/src/types/error_code.dart
```diff
@@ -0,0 +1,90 @@
...
+  anonymousProviderDisabled('anonymous_provider_disabled'),
+  hookTimeout('hook_timeout'),
+  hookTimeoutAfterRetry('hook_timeout_after_retry'),
+  hookPayloadOverSizeLimit('hook_payload_over_size_limit'),
+  hookPayloadUnknownSize('hook_payload_unknown_size'),
+  requestTimeout('request_timeout'),
+  mfaPhoneEnrollDisabled('mfa_phone_enroll_not_enabled'),
+  mfaPhoneVerifyDisabled('mfa_phone_verify_not_enabled'),
+  mfaTotpEnrollDisabled('mfa_totp_enroll_not_enabled'),
+  mfaTotpVerifyDisabled('mfa_totp_verify_not_enabled');
```
dshukertjr, line 80: I couldn't seem to find these error codes, but do you remember where you got them from?

lawinski: I got them from the Supabase Auth error codes documentation:
https://github.com/supabase/auth/blob/master/internal/api/errorcodes.go

Aren't these all the available errors?

lawinski:
@dshukertjr I'm okay with the changes :) LGTM

packages/gotrue/lib/src/fetch.dart
```diff
@@ -50,24 +63,44 @@ class GotrueFetch {
...
-            .isEmpty) {
+    String? errorCode;
+
+    final responseApiVersion = ApiVersion.fromResponse(error);
+
+    if (responseApiVersion?.isSameOrAfter(ApiVersions.v20240101) ?? false) {
+      errorCode = _getErrorCode(data, 'code');
+    } else {
+      errorCode = _getErrorCode(data, 'error_code');
+    }
```
bdlukaa, line 74: This could still fail somehow, specially because the version is set in the client if it is missing in the response headers. What about something like:

```suggestion
    String? errorCode = _getErrorCode(data, 'code') ?? _getErrorCode(data, 'error_code');
```

bdlukaa (commented):
LGTM!

packages/gotrue/lib/src/types/error_code.dart
dshukertjr: Ah, I see. I will check with the auth team about these then. Come to think about it, not sure if we need the list of error codes in the client libs in the first place 😂 We will keep it in for now, but might take it out in the future as it's not really doing anything.

packages/gotrue/lib/src/fetch.dart
dshukertjr: Hmm, in what scenario would the version in the response headers be missing though, and how would it "fail"? In that case, wouldn't it just go to `else` in the above code block returning `_getErrorCode(data, 'error_code')`?

bdlukaa: The gotrue library is not restricted to Supabase, and it is open source and its source can be changed.

Yes, it fallbacks to the else clause, but it would imply that the first clauze isn't tested in some case.

But we can keep the version check, it is fine for me.

dshukertjr: Thanks for the catch! I think at this point we can consider this library dedicated to Supabase, so let's go with the current code 👍

dshukertjr (approved):

PR #999, opened 2024-08-07
feat: Add third-party auth support

dshukertjr:

Adds support for the accessToken option on the Supabase client which can be used to provide a third-party authentication (e.g. Auth0, Clerk, Firebase Auth, ...) access token or ID token to be used instead of Supabase Auth.

When set, supabase.auth.xyz cannot be used and an error will be thrown.

js PR here: https://github.com/supabase/supabase-js/pull/1004

### DIFF
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -59,7 +59,7 @@ class SupabaseClient {
   late final SupabaseStorageClient storage;
   late final RealtimeClient realtime;
   late final PostgrestClient rest;
-  late StreamSubscription<AuthState> _authStateSubscription;
+  StreamSubscription<AuthState>? _authStateSubscription;
   late final YAJsonIsolate _isolate;
   final Future<String> Function()? accessToken;
 
@@ -247,7 +247,7 @@ class SupabaseClient {
   }
 
   Future<void> dispose() async {
-    await _authStateSubscription.cancel();
+    await _authStateSubscription?.cancel();
     await _isolate.dispose();
   }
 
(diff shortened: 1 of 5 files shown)


packages/supabase_flutter/lib/src/supabase.dart
```diff
@@ -75,6 +75,7 @@ class Supabase {
     PostgrestClientOptions postgrestOptions = const PostgrestClientOptions(),
     StorageClientOptions storageOptions = const StorageClientOptions(),
     FlutterAuthClientOptions authOptions = const FlutterAuthClientOptions(),
+    Future<String> Function()? accessToken,
```
bdlukaa, line 78: Maybe create a typedef for this?

bdlukaa (approved):

packages/supabase/lib/src/supabase_client.dart
```diff
@@ -122,18 +133,30 @@ class SupabaseClient {
         },
         _httpClient = httpClient,
         _isolate = isolate ?? (YAJsonIsolate()..initialize()) {
-    auth = _initSupabaseAuthClient(
+    _authInstance = _initSupabaseAuthClient(
```
Vinzent03, line 136: When a custom access token is passed, no auth instance should be created, right? From what I see, this is still the case, but just because it's marked as `late`. I would prefer a clearer structure to ensure no auth instance is created.

packages/supabase_flutter/lib/src/supabase.dart
dshukertjr: Thanks for the feedback! I know this is highly opinionated, but I really feel that using typedef generally decreases the readability of the code. Especially since a property named `accessToken` doesn't sound like a function.

packages/supabase/lib/src/supabase_client.dart
dshukertjr: We can always come back and change the implementation of it. For now, I'm going to merge this PR as this is not really critical.

PR #1003, opened 2024-08-12
feat: Add setHeader method on postgrest builder

Vinzent03:

feature

To change the headers for one rpc or query call the instance wide headers objects needs to be changed.

## What is the new behavior?

After calling `.rpc` or `.from` individual header fields can be set with `.setHeader(key, value)`

## Additional context

js-pr: https://github.com/supabase/postgrest-js/pull/550
issue: #179

### DIFF
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -93,6 +93,12 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
     );
   }
 
+  PostgrestBuilder<T, S, R> setHeader(String key, String value) {
+    return _copyWith(
+      headers: {..._headers, key: value},
+    );
+  }
+
   Future<T> _execute() async {
     final String? method = _method;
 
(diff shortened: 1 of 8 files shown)


dshukertjr:
What's the reason why we implemented `setHeader` method on the query builder and the filter builder instead of having it on the postgrest builder like in JS?

Vinzent03:
The js implementation is able to return the same type again, as it's not immutable like in Dart. By implementing it only on `PostgrestBuilder`, the return type would be `PostgrestBuilder`, which doesn't allow any further filtering after. To be able to call it at every stage in the query building, we would have to implement it in every class. We could do that, but I've found it to be most appealing when called at the start of the query. So I only implemented it for Query- and FilterBuilder, because these are the return types of `.from` and `.rpc`.
So I designed it to be `.from().setHeader()` and `.rpc().setHeader()`. 
Though I see that it could be nice to be more free in where to place the `setHeader()`. So I could implement it in the remaining classes as well.

dshukertjr:
Thanks for the detailed explanation! Yeah, if it's not too much trouble, maybe we can add the same method in other classes as well to make it more flexible?

dshukertjr (approved):
🚀

PR #1005, opened 2024-08-15
feat: Broadcast auth events to other tabs on web

### LINKED ISSUE #979: Users are not getting logged out across tabs on web
**Describe the bug**
Currently, logging a user out in one tab of a web browser does not log the user out in other tabs.

**To Reproduce**
1. Log in in one tab
2. Open a second tab -> The user is logged in in the second tab
3. Log out in the first tab -> An `AuthChangeEvent.signedOut` event is triggered in the first tab but not in the second tab

Actions that require a signed-in user are still possible in the second tab.

**Expected behavior**

The sign-out event should get triggered in all tabs. 

├── supabase_flutter 2.5.5
│   ├── supabase 2.2.1
│   │   ├── functions_client 2.2.0
│   │   ├── gotrue 2.8.0
│   │   ├── postgrest 2.1.2
│   │   ├── realtime_client 2.1.0
│   │   ├── storage_client 2.0.2

Vinzent03:

feature

When singing in or out in a tab on flutter web, the other tabs of the same domain are untouched.

## What is the new behavior?

Every auth event and their session is distributed to the other tabs via the web broadcast channel feature. The js library does this as well.

I've noticed that auth-js doesn't contain the `userDeleted` event, but found no use in our library as well. It got added [here](https://github.com/supabase/gotrue-dart/commit/42def9a4b8177dc3c21434dcbbc559d35881ebac), but I don't see why. So I marked it as deprecated.

Furthermore, from what I see, the auth refresh token and streams got nowhere closed, so I added a `dispose()` method and call it from the dispose method in the `supabase` package.

## Additional context

close #979


Vinzent03:
I couldn't find a way to do automated testing for this. I only tested this manually with a little modified flutter example app. I didn't test the compatibility with the js package as well. Maybe you have a ready to go setup for a flutter and js app on the same domain.

dshukertjr (commented):
Do you have a screen recording that shows how this PR works?

Vinzent03:
I do not currently, but I'm unsure how that helps you. I've added logging to the onAuthStateChange stream and saw that when I log in in one tab, the other tabs received the same `AuthState` event.

dshukertjr:
>I'm unsure how that helps you

I want to verify the behavior. It would be awesome if you could do a quick screen recording verifying that this PR works as intended.

Vinzent03:
I tried a different example app and indeed found the issue of not properly saving the session if it's null. Should be fixed now. Here's the recording as well:
https://github.com/user-attachments/assets/8a515a16-1dc2-4ff0-9328-29a8e5e1c998

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -1128,6 +1143,59 @@ class GoTrueClient {
     _currentUser = null;
   }
 
+  void mayStartBroadcastChannel() {
+    if (const bool.fromEnvironment('dart.library.html') && _broadcastSession) {
```
dshukertjr, line 1147: Not a blocker, but just personally curious. Why are you using this instead of `kIsWeb`?

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -92,11 +103,13 @@ class GoTrueClient {
     Client? httpClient,
     GotrueAsyncStorage? asyncStorage,
     AuthFlowType flowType = AuthFlowType.pkce,
+    bool broadcastSession = true,
```
dshukertjr, line 106: I'm thinking we might not want to expose a parameter for the user to choose to turn this on or off, but is there a situation where the user might want to turn this off? Is this so that the users can keep the same behavior after this version release as well?

dshukertjr (commented):
Thank you so much for the awesome PR! Just had a few small questions, but it looks amazing!

packages/gotrue/lib/src/gotrue_client.dart
bdlukaa: `kIsWeb` is restricted to Flutter. `gotrue_client` runs in Dart apps as well.

dshukertjr: Ah, right!

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -1128,6 +1143,59 @@ class GoTrueClient {
...
   }
 
+  void mayStartBroadcastChannel() {
+    if (const bool.fromEnvironment('dart.library.html') && _broadcastSession) {
+      // Used by the js library as well
+      final broadcastKey =
+          "sb-${Uri.parse(_url).host.split(".").first}-auth-token";
+
+      _broadcastChannel = web.getBroadcastChannel(broadcastKey);
+      _broadcastChannel?.onMessage.listen((messageEvent) {
```
bdlukaa, line 1153: A subscription should be assigned here and closed when disposed or a new channel is broadcasted. 

https://dart.dev/libraries/async/using-streams#listen-method

packages/gotrue/lib/src/gotrue_client.dart
bdlukaa, line 1147: It may be reasonable to add a try/catch block in this function. It can throw some unexpected errors, like `Uri.parse` and `messageRaw['event]`. What do you think?

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -1128,6 +1143,59 @@ class GoTrueClient {
...
+          } else {
+            _removeSession();
+          }
+          notifyAllSubscribers(event, session: session, broadcast: false);
+        }
+      });
+    }
+  }
+
+  void dispose() {
```
bdlukaa, line 1191: I'd add a `mustCallSuper` annotation from the [`meta` package](https://pub.dev/packages/meta)

```suggestion
  @mustCallSuper
  void dispose() {
```

packages/gotrue/lib/src/types/auth_state.dart
```diff
@@ -4,6 +4,12 @@ import 'package:gotrue/src/types/session.dart';
 class AuthState {
   final AuthChangeEvent event;
   final Session? session;
+  final bool fromBroadcast;
 
-  AuthState(this.event, this.session);
+  AuthState(this.event, this.session, {this.fromBroadcast = false});
```
bdlukaa, line 9: Can be const;

```suggestion
  const AuthState(this.event, this.session, {this.fromBroadcast = false});
```

packages/gotrue/lib/src/gotrue_client.dart
Vinzent03: I was unsure whether to add this or not. auth-js has the session persisting built in and not outsourced to supabase-js like we do with supabase_flutter. auth-js only broadcasts the events if session persistence is enabled. https://github.com/supabase/auth-js/blob/8222ee198a0ab10570e8b4c31ffb2aeafef86392/src/GoTrueClient.ts#L244 So I was thinking about doing it more like in js and managing the session persistence in gotrue-dart so that we only broadcast if session gets persisted. But that's to be discussed in another issue so I will remove the option here for now.

dshukertjr (approved):

Issue #1012, opened 2024-08-21
Solving the realtime issues 


maxfornacon:
I would like to address the existing issues with realtime (e.g. #705, #579), particularly the problem where streams suddenly stop receiving data. I'm willing to spend some time investigating the source code and will report back if I find anything noteworthy. Since I'm not yet familiar with the repository, I would like to keep this issue as an open discussion. This will allow us to collaborate, so I can seek help if I encounter difficulties or if you believe my research is heading in the wrong direction.

If you have any additional information or thoughts on why this problem might exist, please leave a comment. Perhaps we can finally find a solution to these issues together as a community.

I'm using this repo for my local testing:
https://github.com/maxfornacon/supabase_test

I'm using latest supabase_flutter (2.6.0) and Flutter 3.24.0.
I'm running this test project on macOS as a desktop application.

**The main problems I'm having with realtime in my projects for some time now is:**
When the application is running for a longer period of time (e.g. desktop applications that are opened (might be in background) the whole day and don't get killed by the system) streams stop receiving new data after some time. There are no errors or exceptions thrown and the only solution I've found is to encourage the users to restart the application from time to time. 

The current state of my research is that there might be a problem related to the accessToken refresh. 
I've set the access token expiry time in supabase dashboard to a small value (30 seconds) and added a breakpoint in this method (realtime_client.dart):

```
  /// Sets the JWT access token used for channel subscription authorization and Realtime RLS.
  ///
  /// `token` A JWT strings.
  void setAuth(String? token) {
    accessToken = token;

    for (final channel in channels) {
      if (token != null) {
        channel.updateJoinPayload({'user_token': token});
      }
      if (channel.joinedOnce && channel.isJoined) {
        channel.push(ChannelEvents.accessToken, {'access_token': token});
      }
    }
  }
```

This gets called every time the accessToken is refreshed and should update existing channels with the new token (that's my understanding).

I watched the `channels` List which contains 3 entries in the beginning for my test project. After some token refreshes this list contains < 3 entries and eventually 0 in the end. This is also when I noticed that the streams don't receive new data anymore.

I don't know yet why the channels get removed from this list but It feels not intended.

That's my current status. I'm curious about your thoughts on this.

maxfornacon:
Today I checked if the _onClose()-method of the RealtimeChannel gets called and therefore the channel gets removed from the channels List. It does by receiving an event `phx_close`. I'm not sure what this means. Maybe you can help me out with that. 
[...]

Issue #705, opened 2023-11-12
Incorrect data after app goes in the background


vkammerer:
When an iOS app goes in the background, it looses network after a while (about 90 seconds according to my tests with the simulator).
If new rows are then created in a Supabase table that the app listens to, the app doesn't receive the data (which is expected).
When the app reconnects after idling, it starts receiving table addition events (which is expected), but it emits a representation of the data without the events that happened while is was offline (which is **not** expected).

I looked a bit at the code and can see that `supabase-flutter` manages its own cache, and that it manages internal representations of a table on atomic events like "INSERT" for example.

Maybe a solution could be to fetch the whole table after a reconnection? Or maybe to pass a reference to the previous message for all atomic events, so that the app can ensure that it hasn't missed one?

See video below for an example: the "third message" is ignored, and an incorrect representation of the messages list is emitted.
https://github.com/supabase/supabase-flutter/assets/1177483/5b44500b-d3eb-476b-8946-f02d4029c8f4

1. Listen to a table
```dart
supabase.from('messages').stream(primaryKey: ['id']).listen((event) {
  setState(() {
    _messages = event.map(Message.fromJson).toList();
});
```
2. Move the app to the background
3. Wait 1:30 minute
4. Add a row in the supabase table
5. Bring the app to the foreground
6. Add a row in the supabase table

**Expected behavior**
The stream emits a correct representation of the data.

**Version:**
I am on MacOS
```shell
dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"
│   ├── supabase_flutter 2.0.0-dev.1
│   │   ├── supabase 2.0.0-dev.1
│   │   │   ├── functions_client 2.0.0-dev.0
│   │   │   ├── gotrue 2.0.0-dev.0
│   │   │   ├── postgrest 2.0.0-dev.0
│   │   │   ├── realtime_client 2.0.0-dev.1
│   │   │   ├── storage_client 2.0.0-dev.0
```

dippa-1:
Thanks for testing this, as I was wondering what the behaviour in this case is.

How have you fixed it?
I assume the solution is, in the case of Flutter, to add a life cycle state listener and when the app is in foreground again, refetch the data of all streams again.

Just to be sure: I assume that the stream behaves the same as an on change listener, right?

vkammerer:
[...]

Issue #674, opened 2023-10-18
Bug: Stream behaviour on connection lost and reconnect


coolusaHD:
**Describe the bug**
With the update to `1.2.3` , losing the connection doesn't trigger any error anymore. Also reconnecting to the internet doesn't change/trigger anything.

My current code (with `v1.2.2`)

``` 
import 'dart:async';
import 'package:flutter/material.dart';
import 'package:supabase_flutter/supabase_flutter.dart';

class RealtimeGameService extends ChangeNotifier {
  RealtimeGameService({
    required this.client,
    required this.primaryKey,
    required this.tableName,
  }) {
    _init();
  }

  final SupabaseClient client;
  final String tableName;
  final List<String> primaryKey;

  List<Map<String, dynamic>> data = [];
  DateTime lastUpdateTime = DateTime.now();
  bool isConnected = false;
  bool isInitialLoading = true;

  StreamSubscription<List<Map<String, dynamic>>>? _streamSubscription;

  void _init() {
    _streamSubscription?.cancel(); // Cancel any existing subscription
    final liveDataStream =
        client.from(tableName).stream(primaryKey: primaryKey);

    _streamSubscription = liveDataStream.listen(
      (List<Map<String, dynamic>> newData) {
        print('works');
        print('Received data: $newData');
        _updateData(newData, true);
        if (isInitialLoading) {
          isInitialLoading = false; // Mark initial loading as complete
          notifyListeners();
        }
      },
      onError: (e) {
        print('Error during realtime subscription: $e');
        _updateData(data, false);
        if (isInitialLoading) {
          isInitialLoading = false; // Mark initial loading as complete
          notifyListeners();
        }
      },
      onDone: () {
        print('Realtime subscription done');
      },
    );
  }

  void _updateData(List<Map<String, dynamic>> newData, bool newIsConnected) {
    data = newData;
    lastUpdateTime = newIsConnected ? DateTime.now() : lastUpdateTime;
    isConnected = newIsConnected;
    notifyListeners();
  }

  void reload() {
    _init();
    isInitialLoading = true;
    notifyListeners();
  }

  @override
  void dispose() {
    _streamSubscription?.cancel();
    super.dispose();
  }
}
```
**Expected behavior**

1.  At least one single error should be thrown that the websocket lost connection
2. It would be nice if the websocket reconnects it triggers the `onData` again

**Version (please complete the following information):**
????????? realtime_client 1.2.2
????????? supabase_flutter 1.10.18
[...]

PR #1019, opened 2024-08-28
fix: Better stream and access token management

Vinzent03:

Bug fix

The behavior of the app in background depends on the platform, but by always stopping the access token refreshing, the realtime connection may keep an invalid access token and therefore the connection get closed by server. When the realtime connection gets interrupted, it gets automatically reconnected as long as it never sent some expired access token.

## What is the new behavior?

~~When a realtime channel is open, the access token keeps refreshing while the app is in background.~~

In addition, I did more management to keep realtiem and postgrest in sync for the stream method. If the postgrest request fails, I unsubscribe the realtime connection and close the stream as I see no benefit in leaving it open, as it would just be a realtime connection by then. The dev using the stream method should handle such cases. One could also argue about doing a few retries as well, though.
When the realtime connection gets a channel error, but gets resubscribed, I reload the postgrest data to ensure the data is up to date and didn't miss any updates due to the interrupted realtime connection. 

## Additional context

#1012 #705 #579

### DIFF
--- a/packages/realtime_client/lib/src/constants.dart
+++ b/packages/realtime_client/lib/src/constants.dart
@@ -18,8 +18,8 @@ enum SocketStates {
   /// Connection is live and connected
   open,
 
-  /// Socket is closing.
-  closing,
+  /// Socket is closing by the user
+  disconnecting,
 
   /// Socket being close not by the user. Realtime should attempt to reconnect.
   closed,
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -70,7 +70,7 @@ class RealtimeChannel {
       socket.remove(this);
     });
 
-    _onError((String? reason) {
+    _onError((reason) {
       if (isLeaving || isClosed) {
         return;
       }
@@ -260,9 +260,9 @@ class RealtimeChannel {
   }
 
   /// Registers a callback that will be executed when the channel encounteres an error.
-  void _onError(void Function(String?) callback) {
+  void _onError(Function callback) {
     onEvents(ChannelEvents.error.eventName(), ChannelFilter(),
-        (reason, [ref]) => callback(reason?.toString()));
+        (reason, [ref]) => callback(reason));
   }
 
   /// Sets up a listener on your Supabase database.
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -646,11 +646,15 @@ class RealtimeChannel {
     joinPush.resend(timeout ?? _timeout);
   }
 
-  /// Usually a rejoin only happens when the channel timeouts or errors out.
+  /// Resends [joinPush] to tell the server we join this channel again and marks
+  /// the channel as [ChannelStates.joining].
+  ///
+  /// Usually [rejoin] only happens when the channel timeouts or errors out.
   /// When manually disconnecting, the channel is still marked as
   /// [ChannelStates.joined]. Calling [RealtimeClient.leaveOpenTopic] will
   /// unsubscribe itself, which causes issues when trying to rejoin. This method
   /// therefore doesn't call [RealtimeClient.leaveOpenTopic].
+  @internal
   void forceRejoin([Duration? timeout]) {
     if (isLeaving) {
       return;
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -646,6 +646,19 @@ class RealtimeChannel {
     joinPush.resend(timeout ?? _timeout);
   }
 
+  /// Usually a rejoin only happens when the channel timeouts or errors out.
+  /// When manually disconnecting, the channel is still marked as
+  /// [ChannelStates.joined]. Calling [RealtimeClient.leaveOpenTopic] will
+  /// unsubscribe itself, which causes issues when trying to rejoin. This method
+  /// therefore doesn't call [RealtimeClient.leaveOpenTopic].
+  void forceRejoin([Duration? timeout]) {
+    if (isLeaving) {
+      return;
+    }
+    _state = ChannelStates.joining;
+    joinPush.resend(timeout ?? _timeout);
+  }
+
   void trigger(String type, [dynamic payload, String? ref]) {
     final typeLower = type.toLowerCase();
 
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -48,7 +48,7 @@ class RealtimeCloseEvent {
 
   @override
   String toString() {
-    return 'RealtimeCloseEvent{code: $code, reason: $reason}';
+    return 'RealtimeCloseEvent(code: $code, reason: $reason)';
   }
 }
 
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -158,8 +158,13 @@ class RealtimeClient {
       connState = SocketStates.connecting;
       conn = transport(endPointURL, headers);
 
-      // handle connection errors
-      conn!.ready.catchError(_onConnError);
+      try {
+        await conn!.ready;
+      } catch (error) {
+        _onConnError(error);
+        reconnectTimer.scheduleTimeout();
+        return;
+      }
 
       connState = SocketStates.open;
 
@@ -187,17 +192,20 @@ class RealtimeClient {
   void disconnect({int? code, String? reason}) {
     final conn = this.conn;
     if (conn != null) {
+      final connectionWasOpen = connState == SocketStates.open;
       connState = SocketStates.disconnected;
-      if (code != null) {
-        conn.sink.close(code, reason ?? '');
-      } else {
-        conn.sink.close();
+      if (connectionWasOpen) {
+        if (code != null) {
+          conn.sink.close(code, reason ?? '');
+        } else {
+          conn.sink.close();
+        }
+        reconnectTimer.reset();
       }
       this.conn = null;
 
       // remove open handles
       if (heartbeatTimer != null) heartbeatTimer?.cancel();
-      reconnectTimer.reset();
     }
   }
 
@@ -404,7 +412,7 @@ class RealtimeClient {
     /// SocketStates.disconnected: by user with socket.disconnect()
     /// SocketStates.closed: NOT by user, should try to reconnect
     if (connState == SocketStates.closed) {
-      _triggerChanError();
+      _triggerChanError(event);
       reconnectTimer.scheduleTimeout();
     }
     if (heartbeatTimer != null) heartbeatTimer!.cancel();
@@ -415,15 +423,15 @@ class RealtimeClient {
 
   void _onConnError(dynamic error) {
     log('transport', error.toString());
-    _triggerChanError();
+    _triggerChanError(error);
     for (final callback in stateChangeCallbacks['error']!) {
       callback(error);
     }
   }
 
-  void _triggerChanError() {
+  void _triggerChanError([dynamic error]) {
     for (final channel in channels) {
-      channel.trigger(ChannelEvents.error.eventName());
+      channel.trigger(ChannelEvents.error.eventName(), error);
     }
   }
 
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -45,6 +45,11 @@ class RealtimeCloseEvent {
     required this.code,
     required this.reason,
   });
+
+  @override
+  String toString() {
+    return 'RealtimeCloseEvent{code: $code, reason: $reason}';
+  }
 }
 
 class RealtimeClient {
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -353,7 +353,7 @@ class RealtimeClient {
 
     for (final channel in channels) {
       if (token != null) {
-        channel.updateJoinPayload({'user_token': token});
+        channel.updateJoinPayload({'access_token': token});
       }
       if (channel.joinedOnce && channel.isJoined) {
         channel.push(ChannelEvents.accessToken, {'access_token': token});
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -139,9 +139,9 @@ class RealtimeClient {
         (String payload, Function(dynamic result) callback) =>
             callback(json.decode(payload));
     reconnectTimer = RetryTimer(
-      () {
-        disconnect();
-        connect();
+      () async {
+        await disconnect();
+        await connect();
       },
       this.reconnectAfterMs,
     );
@@ -149,7 +149,7 @@ class RealtimeClient {
 
   /// Connects the socket.
   @internal
-  void connect() async {
+  Future<void> connect() async {
     if (conn != null) {
       return;
     }
@@ -161,8 +161,15 @@ class RealtimeClient {
       try {
         await conn!.ready;
       } catch (error) {
-        _onConnError(error);
-        reconnectTimer.scheduleTimeout();
+        // Don't schedule a reconnect and emit error if connection has been
+        // closed by the user or [disconnect] waits for the connection to be
+        // ready before closing it.
+        if (connState != SocketStates.disconnected &&
+            connState != SocketStates.disconnecting) {
+          connState = SocketStates.closed;
+          _onConnError(error);
+          reconnectTimer.scheduleTimeout();
+        }
         return;
       }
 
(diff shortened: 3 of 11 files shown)


packages/realtime_client/lib/src/realtime_client.dart
```diff
@@ -353,7 +358,7 @@ class RealtimeClient {
 
     for (final channel in channels) {
       if (token != null) {
-        channel.updateJoinPayload({'user_token': token});
```
Vinzent03, line 356: The realtime-js library uses `access-token` [see here](https://github.com/supabase/realtime-js/blob/95ab09e47ed7732eb21ac31eac743704bacc29aa/src/RealtimeClient.ts#L322) instead of `user_token`. And I guess that's the correct payload. It got changed to access_token [here](https://github.com/supabase/realtime-js/commit/b3eafe820f61afdb4c663fa600c4f209757ba8b4#diff-f5a6b192eb9d6c05f61c33d8629b814ad893fbfb729bfa7af9b226e5c87d2e85)

coolusaHD:
Hey @Vinzent03 

This PR looks nice and maybe could also be the solution of our problem #674 .
If you can reply here or in the issue thread we would love to test the update when you finished your work ✌️ 

Thanks 🚀

dshukertjr (commented):
Thanks for this. This looks awesome. I'm a bit swamped today and maybe tomorrow, but let me take a look whenever I can.

dshukertjr:
@Vinzent03 BTW, have you actually ran this code on iOS and Android? I tried to fix some of the realtime issues, and found that some behaviors differ on the web sockets between those two OSs. I don't remember what exactly those differences were, but wanted to ask if you have tested on actual apps.

Vinzent03:
I have no way to test this on any Apple device. I tested this on my android device and as a linux application. The stopping of refreshing the access token was especially an issue in the linux application, because the realtime connection keeps being open, when the app is in background, because there is no throttling after some time.

maxfornacon:
I changed the supabase_flutter package in my pubspec.yaml to the version of this PR. 
I started the macOS-Application and left it running over night with my OS in sleep mode and most likely disconnected WiFi at some point. 

I got the exception: 

```
Unhandled Exception: Exception: Realtime subscription closed
#0      SupabaseStreamBuilder._addException (package:supabase/src/supabase_stream_builder.dart:296:67)
#1      SupabaseStreamBuilder._getStreamData.<anonymous closure> (package:supabase/src/supabase_stream_builder.dart:202:9)
#2      RealtimeChannel.subscribe.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:130:39)
#3      RealtimeChannel._onClose.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:259:36)
#4      RealtimeChannel.trigger (package:realtime_client/src/realtime_channel.dart:694:22)
#5      RealtimeChannel.unsubscribe.onClose (package:realtime_client/src/realtime_channel.dart:575:7)
#6      RealtimeChannel.unsubscribe.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:586:7)
#7      Push._matchReceive.<anonymous closure> (package:realtime_client/src/push.dart:133:17)
#8      Iterable.forEach (dart:core/iterable.dart:347:35)
#9      Push._matchReceive (package:realtime_client/src/push.dart:132:48)
#10     Push.startTimeout.<anonymous closure> (package:realtime_client/src/push.dart:96:7)
#11     RealtimeChannel.trigger (package:realtime_client/src/realtime_channel.dart:694:22)
#12     Push.trigger (package:realtime_client/src/push.dart:106:16)
#13     RealtimeChannel.unsubscribe (package:realtime_client/src/realtime_channel.dart:605:17)
#14     RealtimeClient.leaveOpenTopic (package:realtime_client/src/realtime_client.dart:372:18)
#15     RealtimeChannel.rejoin (package:realtime_client/src/realtime_channel.dart:633:12)
#16     RealtimeChannel.rejoinUntilConnected (package:realtime_client/src/realtime_channel.dart:103:7)
#17     new RealtimeChannel.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:56:26)
#18     RetryTimer.scheduleTimeout.<anonymous closure> (package:realtime_client/src/retry_timer.dart:46:15)
#19     Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:18:15)
#20     _Timer._runTimers (dart:isolate-patch/timer_impl.dart:398:19)
#21     _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:429:5)
#22     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
```

The stream did not reconnect on its own. But at least there is this exception now.

I'm gonna do some further testing soon.

maxfornacon:
I'd like to test the changes in my projects.

I tried to clone the repo and switched to the `fix/realtime` branch. Then I ran `melos bs` and added the package to my pubspec.yaml like this: 

```
  supabase_flutter:
    path: /Users/maxfornacon/projects/test/supabase-flutter/packages/supabase_flutter
```

But I noticed that e.g. `realtime_client` still gets installed from hosted and not using the local version. 
Can you give me a hint on how to use all packages from the local path?

coolusaHD:
>  Can you give me a hint on how to use all packages from the local path?

Have you tried also add the local import to your `pubspec.yaml` file
Like 

```
realtime_client:
    path: /Users/maxfornacon/projects/test/supabase-flutter/packages/realtime_client
```

maxfornacon:
> Have you tried also add the local import to your pubspec.yaml file

I added all necessary packages like this as `dependency_overrides`. This works I guess.

maxfornacon:
I started the Application on macOS, closed my MacBook and went home. When I opened it at home I saw:

```
AuthException(message: ClientException with SocketException: Failed host lookup: 'adnwhmobivhficannylf.supabase.co' (OS Error: nodename nor servname provided, or not known, errno = 8), uri=https://adnwhmobivhficannylf.supabase.co/auth/v1/token?grant_type=refresh_token, statusCode: null, errorCode: null)
flutter: #0      GotrueFetch._handleRequest (package:gotrue/src/fetch.dart:179:7)
flutter: <asynchronous suspension>
flutter: #1      GotrueFetch.request (package:gotrue/src/fetch.dart:130:12)
flutter: <asynchronous suspension>
flutter: #2      GoTrueClient._refreshAccessToken.<anonymous closure> (package:gotrue/src/gotrue_client.dart:1052:26)
flutter: <asynchronous suspension>
flutter: #3      RetryOptions.retry (package:retry/retry.dart:131:16)
flutter: <asynchronous suspension>
flutter: #4      GoTrueClient._refreshAccessToken (package:gotrue/src/gotrue_client.dart:1044:12)
flutter: <asynchronous suspension>
flutter: #5      GoTrueClient._callRefreshToken (package:gotrue/src/gotrue_client.dart:1150:20)
flutter: <asynchronous suspension>
flutter: #6      GoTrueClient._autoRefreshTokenTick (package:gotrue/src/gotrue_client.dart:1032:9)
```
This indicates a loss of internet connection, as expected.

Also the StreamBuilder printed out `RealtimeSubscribeException(status: RealtimeSubscribeStatus.channelError, details: null)`.

This is an improvement!

If it could auto reconnect would be even better. But that's something I could work with.

maxfornacon:
I started the Application on macOS, closed my MacBook and went home. When I opened it at home I saw:

```
AuthException(message: ClientException with SocketException: Failed host lookup: 'xxx.supabase.co' (OS Error: nodename nor servname provided, or not known, errno = 8), uri=https:/xxx.supabase.co/auth/v1/token?grant_type=refresh_token, statusCode: null, errorCode: null)
flutter: #0      GotrueFetch._handleRequest (package:gotrue/src/fetch.dart:179:7)
flutter: <asynchronous suspension>
flutter: #1      GotrueFetch.request (package:gotrue/src/fetch.dart:130:12)
flutter: <asynchronous suspension>
flutter: #2      GoTrueClient._refreshAccessToken.<anonymous closure> (package:gotrue/src/gotrue_client.dart:1052:26)
flutter: <asynchronous suspension>
flutter: #3      RetryOptions.retry (package:retry/retry.dart:131:16)
flutter: <asynchronous suspension>
flutter: #4      GoTrueClient._refreshAccessToken (package:gotrue/src/gotrue_client.dart:1044:12)
flutter: <asynchronous suspension>
flutter: #5      GoTrueClient._callRefreshToken (package:gotrue/src/gotrue_client.dart:1150:20)
flutter: <asynchronous suspension>
flutter: #6      GoTrueClient._autoRefreshTokenTick (package:gotrue/src/gotrue_client.dart:1032:9)
```
This indicates a loss of internet connection, as expected.

Also the StreamBuilder printed out `RealtimeSubscribeException(status: RealtimeSubscribeStatus.channelError, details: null)`.

This is an improvement!

If it could auto reconnect would be even better. But that's something I could work with.

maxfornacon:
Today I noticed the StreamBuilder not receiving new data after 5min. 

I could see this printed out from listening to the channel directly:

```
PostgresChangePayload(schema: xxx, table: notifications, commitTimestamp: 2024-09-13 12:51:23.119Z, eventType: PostgresChangeEvent.delete, newRow: {}, oldRow: {id: 55311517-01b3-4f8e-b3d5-6b4abbb43e8f}, errors: null)
```

But the StreamBuilder was not showing the change.

packages/supabase_flutter/lib/src/supabase_auth.dart
```diff
@@ -119,7 +119,12 @@ class SupabaseAuth with WidgetsBindingObserver {
       case AppLifecycleState.detached:
       case AppLifecycleState.inactive:
       case AppLifecycleState.paused:
-        Supabase.instance.client.auth.stopAutoRefresh();
+        // Realtime channels are kept alive in the background for some amount
+        // of time after the app is paused. If we stop refreshing the token
+        // here, the channels will be closed.
+        if (Supabase.instance.client.realtime.getChannels().isEmpty) {
+          Supabase.instance.client.auth.stopAutoRefresh();
+        }
```
dshukertjr, line 127: Hmm, the purpose of this piece of code is to lower the risks of creating an invalid JWT on the client side, and we should stop attempting to refresh the token when the app is in the background. 

If we could maybe stop subscribing to realtime through stream when the app is in background, that would be a more ideal implementation of this I would say.

dshukertjr (commented):
The core issue with the realtime currently is the fact that the SDK isn't able to reconnect in the following scenario:
1. app is connected to realtime
2. realtime is disconnected (bad network or app goes to the background) long enough for the JWT to expire
3. app comes back online

At step 3 above, the connection should be restored, and that is the behavior of the js SDK. I have spend some time trying to fix the issue, but have had no success on this SDK, but if we could fix that, all the other issues should be fixed fairly straight forward.

packages/supabase_flutter/lib/src/supabase_auth.dart
Vinzent03: I think it's a bad idea to disconnect from realtime when going to background. On desktop for example everything keeps running, when going to background. On mobile you may often switch for a quick moment to another app. Disconnecting and reconnecting might miss events and causes overhead. I will do some further testing on this part.

packages/realtime_client/lib/src/constants.dart
```diff
@@ -18,8 +18,8 @@ enum SocketStates {
   /// Connection is live and connected
   open,
 
-  /// Socket is closing.
-  closing,
```
Vinzent03, line 22: This enum value was never in use and replaced it therefore with a new one.

Vinzent03:
I've done now quite more changes. 
- I've undone #637 by again emitting the exact channel error, so that one can handle `SocketExceptions` and so on explicitly
- The errors received from `onError` aren't stringified anymore. 
- I had to make `connect` and `disconnect` async, because the previous code was wrong. We have to wait for the `conn.ready` future to be [completed before using the sink](https://pub.dev/packages/web_socket_channel#docs-and-usage).
- Added a new `SocketStates.disconnecting` since disconnecting had to be async as well, and emitting a `SocketStates.disconnected`, while it's not actually disconnected is wrong. 
- Removed `AppLifecycleState.inactive` as state to trigger stopping the refresh token and disconnecting realtime, because from reading the [documentation](https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html) this is called often and does not result in stopping the programm code
- When the app goes to actually background I now disconnect the realtime client. When resuming the work is more complicated. When quickly switching between fore and background and having a bad internet connection, the disconnection may still be running, which needed some extra care, but should be well documented in the code comments.
- I needed the method `RealtimeChannel.forceRejoin()`, because the channel is still marked as joined, which results in unsubscribing itself in `leaveOpenTopic()`, which resulted in only closed channels.

themightychris:
> The core issue with the realtime currently is the fact that the SDK isn't able to reconnect in the following scenario:
> 
> 1. app is connected to realtime
> 2. realtime is disconnected (bad network or app goes to the background) long enough for the JWT to expire
> 3. app comes back online
> 
> At step 3 above, the connection should be restored, and that is the behavior of the js SDK. I have spend some time trying to fix the issue, but have had no success on this SDK, but if we could fix that, all the other issues should be fixed fairly straight forward.

I consistently see the SDK fail to reconnect when the above happens well within the JWT's expiration time, is there another explanation for that addressed here?

maxfornacon:
Thanks @Vinzent03 for your amazing work.

I made a few tests where I had the application open for hours during the day or over night. The stream still received the changes after that, every time. This is huge!

Nevertheless, I got quite a lot of Exceptions in the log.

Mostly:
```
Unhandled Exception: RealtimeSubscribeException(status: RealtimeSubscribeStatus.channelError, details: RealtimeCloseEvent(code: 1002, reason: null))
#0      SupabaseStreamBuilder._addException (package:supabase/src/supabase_stream_builder.dart:331:67)
#1      SupabaseStreamBuilder._getStreamData.<anonymous closure> (package:supabase/src/supabase_stream_builder.dart:229:11)
#2      RealtimeChannel.subscribe.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:127:39)
#3      RealtimeChannel._onError.<anonymous closure> (package:realtime_client/src/realtime_channel.dart:265:36)
#4      RealtimeChannel.trigger (package:realtime_client/src/realtime_channel.dart:718:22)
#5      RealtimeClient._triggerChanError (package:realtime_client/src/realtime_client.dart:451:15)
#6      RealtimeClient._onConnClose (package:realtime_client/src/realtime_client.dart:432:7)
#7      RealtimeClient.connect.<anonymous closure> (package:realtime_client/src/realtime_client.dart:190:11)
```

```
Unhandled Exception: RealtimeSubscribeException(status: RealtimeSubscribeStatus.channelError, details: RealtimeCloseEvent(code: 1005, reason: ))
```


What I haven't tested yet is what happens when changes occur while the application is paused. Whether they are reliably loaded when resuming.

Vinzent03:
@themightychris Are you sure you've used the latest state of this pr? For me and looks like for @maxfornacon it works fine.

@maxfornacon These errors are actually expected, because the websocket connection fails when loosing internet or the app is in background. I think it's better to add them to the `.stream()`  stream, so you can react when the connection is interrupted.

themightychris:
@Vinzent03 I have not tested this PR in a real environment yet, I was just responding to the description of the problem that seemed to suggest JWT expiration being the root of it, which would make #674 a separate issue

maxfornacon:
> For me and looks like for @maxfornacon it works fine.
It works for a long time, but eventually the stream gets closed. 

I receive those errors from `snapshot.error` of a StreamBuilder.
```
RealtimeSubscribeException(status: RealtimeSubscribeStatus.channelError, details: RealtimeCloseEvent(code: 1000, reason: heartbeat timeout))
```
and 
```
RealtimeSubscribeException(status: RealtimeSubscribeStatus.timedOut, details: null)
```

>  I think it's better to add them to the .stream() stream, so you can react when the connection is interrupted.

Yes, but there are a lot of the exceptions, but the stream seams to work fine despite of them. Wouldn't handling them cause more harm than good?

Handling the above mentioned exceptions would be fine, because then the stream is really closed and it makes sense to show it to the user to manually reload or implement some kind of auto refresh.

Vinzent03:
The heartbeat exception is thrown from the client side, because the server doesn't answer or similar. But this shouldn't cause the stream to be closed. Adding exceptions to the stream doesn't mean it's closed. The connection of the realtime client may be interrupted, but the stream shouldn't be closed yet. You can determine when a stream is closed by the `onDone` callback in `theStream.listen()`. 
From the package side, there should be no uncaught exceptions. They should all be caught and then added to the stream. You are then supposed to catch them by adding the `onError` and `onDone` callback when using `theStream.listen()`.
https://api.flutter.dev/flutter/dart-async/Stream/listen.html

packages/realtime_client/lib/src/realtime_channel.dart
```diff
@@ -646,6 +646,19 @@ class RealtimeChannel {
     joinPush.resend(timeout ?? _timeout);
   }
 
+  /// Usually a rejoin only happens when the channel timeouts or errors out.
+  /// When manually disconnecting, the channel is still marked as
+  /// [ChannelStates.joined]. Calling [RealtimeClient.leaveOpenTopic] will
+  /// unsubscribe itself, which causes issues when trying to rejoin. This method
+  /// therefore doesn't call [RealtimeClient.leaveOpenTopic].
+  void forceRejoin([Duration? timeout]) {
```
dshukertjr, line 654: Could we add `@internal` annotation here? Could we update the comments to describe what this method does as well?

dshukertjr (commented):
Just had a little comment, but looks awesome!

dshukertjr (approved):

PR #1023, opened 2024-09-04
feat: Custom-metadata, exists, info methods

### LINKED ISSUE #1021: Support for custom metadata on storage objects missing
In July, a `user_metadata` JSONB field [was added to the Supabase Storage objects](https://github.com/supabase/storage/pull/518). Two weeks ago, support for this field was [added to the Supabase Storage JS SDK](https://github.com/supabase/storage-js/pull/207).

Now it's time for the Flutter SDK to add support for this 😄

Vinzent03:

feature

No support for specifying metadata for files when uploading and missing info and exists methods.

## What is the new behavior?

Support specifying metadata when uploading files.

## Additional context

close #1021
[js pr](https://github.com/supabase/storage-js/pull/207)


Vinzent03:
Currently waiting for an answer on https://github.com/supabase/storage/issues/541 because I'm unsure about the correct types of the `FileObjectV2` class.

packages/storage_client/lib/src/fetch.dart
```diff
@@ -98,31 +108,15 @@ class Fetch {
       filename: file.path,
       contentType: contentType,
     );
-    final request = http.MultipartRequest(method, Uri.parse(url))
```
Vinzent03, line 101: The deleted code here is the same as in `_handleBinaryFileRequest()` so I added additional `_handleFileRequest()` instead to reduce code duplication. These two can now use an actual `_handleMultipartRequest()` method, which just accepts a `MultipartFile` which is independent of a passed `File` or `Uint8List`.

dshukertjr (approved):

Issue #966, opened 2024-07-06
getAuthenticatorAssuranceLevel() throws Unhandled Exception: Bad state: No element


benjaminledet:
**Describe the bug**
The `getAuthenticatorAssuranceLevel() `method throws an `Unhandled Exception: Bad state: No element`

```
E/flutter (11692): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Bad state: No element
E/flutter (11692): #0      ListBase.firstWhere (dart:collection/list.dart:132:5)
E/flutter (11692): #1      new AMREntry.fromJson (package:gotrue/src/types/mfa.dart:280:32)
E/flutter (11692): #2      GoTrueMFAApi.getAuthenticatorAssuranceLevel.<anonymous closure> (package:gotrue/src/gotrue_mfa_api.dart:184:30)
E/flutter (11692): #3      MappedListIterable.elementAt (dart:_internal/iterable.dart:425:31)
E/flutter (11692): #4      ListIterator.moveNext (dart:_internal/iterable.dart:354:26)
E/flutter (11692): #5      new _GrowableList._ofEfficientLengthIterable (dart:core-patch/growable_array.dart:189:27)
E/flutter (11692): #6      new _GrowableList.of (dart:core-patch/growable_array.dart:150:28)
E/flutter (11692): #7      new List.of (dart:core-patch/array_patch.dart:39:18)
E/flutter (11692): #8      ListIterable.toList (dart:_internal/iterable.dart:224:7)
E/flutter (11692): #9      GoTrueMFAApi.getAuthenticatorAssuranceLevel (package:gotrue/src/gotrue_mfa_api.dart:185:10)

```
**To Reproduce**
1. sign in with magic link
2. call the method `supabase.auth.mfa.getAuthenticatorAssuranceLevel()`

**Expected behavior**
The method returns the assurance level.

**Version (please complete the following information):**
```
─ supabase_flutter 2.5.7
│   ├── supabase 2.2.3
│   │   ├── functions_client 2.2.0
│   │   ├── gotrue 2.8.2
│   │   ├── postgrest 2.1.2
│   │   ├── realtime_client 2.1.0
│   │   ├── storage_client 2.0.2
```

**Additional context**
The `toList() `of the piece of code throws the exception:
```
  final amr = (payload['amr'] as List)
        .map((e) => AMREntry.fromJson(Map.from(e)))
        .toList();
```

My amr payload is : 
```
{
[...]

PR #1030, opened 2024-09-10
fix: Support all mfa auth methods

Vinzent03:

Bug fix

Not all mfa auth methods can be parsed.

## What is the new behavior?

Add the rest of methods from [here](https://github.com/supabase/auth/blob/a6c18243b92b74798b6317e1c35c8a73bc3fd6e1/internal/models/factor.go#L54)

## Additional context

Just in case, is the auth team aware that the `MFAPhone` has different string values? :sweat_smile:

https://github.com/supabase/auth/blob/a6c18243b92b74798b6317e1c35c8a73bc3fd6e1/internal/models/factor.go#L81

and

https://github.com/supabase/auth/blob/a6c18243b92b74798b6317e1c35c8a73bc3fd6e1/internal/models/factor.go#L113

I have chosen `mfa/phone` for now.

#966

### DIFF
--- a/packages/gotrue/lib/src/types/mfa.dart
+++ b/packages/gotrue/lib/src/types/mfa.dart
@@ -1,4 +1,3 @@
-import 'package:collection/collection.dart';
 import 'package:gotrue/gotrue.dart';
 
 class AuthMFAEnrollResponse {
--- a/packages/gotrue/lib/src/types/mfa.dart
+++ b/packages/gotrue/lib/src/types/mfa.dart
@@ -1,3 +1,4 @@
+import 'package:collection/collection.dart';
 import 'package:gotrue/gotrue.dart';
 
 class AuthMFAEnrollResponse {
@@ -270,7 +271,8 @@ enum AMRMethod {
   emailChange('email_change'),
   tokenRefresh('token_refresh'),
   anonymous('anonymous'),
-  mfaPhone('mfa/phone');
+  mfaPhone('mfa/phone'),
+  unknown('unknown');
 
   final String code;
   const AMRMethod(this.code);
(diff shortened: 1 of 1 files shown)


packages/gotrue/lib/src/types/mfa.dart
```diff
@@ -257,7 +257,24 @@ class AuthMFAGetAuthenticatorAssuranceLevelResponse {
...
+  magiclink('magiclink'),
+  recovery('recovery'),
+  invite('invite'),
+  ssoSaml('sso/saml'),
+  emailSignUp('email/signup'),
+  emailChange('email_change'),
+  tokenRefresh('token_refresh'),
+  anonymous('anonymous'),
+  mfaPhone('mfa/phone');
+
```
dshukertjr, line 274: Another thing mentioned from the auth team was there might be others added in the future. Could we have a `unknown` or something to handle values that gets added in the future versions?

packages/gotrue/lib/src/types/mfa.dart
```diff
@@ -277,9 +296,10 @@ class AMREntry {
 
   factory AMREntry.fromJson(Map<String, dynamic> json) {
     return AMREntry(
-      method: AMRMethod.values.firstWhere(
-        (e) => e.name == json['method'],
-      ),
+      method: AMRMethod.values.firstWhereOrNull(
+            (e) => e.code == json['method'],
+          ) ??
+          AMRMethod.unknown,
```
bdlukaa, line 302: It is possible to use `orElse` instead of `collection.firstWhereOrNull`.

```suggestion
      method: AMRMethod.values.firstWhere(
            (e) => e.code == json['method'],
            orElse: () => AMRMethod.unknown
          ),
```

dshukertjr (approved):

PR #1042, opened 2024-09-24
feat: Add logging

Vinzent03:

feature

When debugging the auth broadcasting or the realtime/stream fixed I had to manually add many print statements, which was annoying. The js sdk already has some logging.

## What is the new behavior?

Add logging via the [logging](https://pub.dev/packages/logging) package from dart.dev. This primarily only tracks the logs. When `debug` is true in the supabase_flutter `Supabase` instance, records with level `INFO` or higher are printed.
The logger instances are laid in a hierarchy separated by dots, which allows all supabase loggers to be united in a `supabase` logger instance.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -105,23 +105,22 @@ class GoTrueClient {
     GotrueAsyncStorage? asyncStorage,
     AuthFlowType flowType = AuthFlowType.pkce,
   })  : _url = url ?? Constants.defaultGotrueUrl,
-        _headers = headers ?? {},
+        _headers = {
+          ...Constants.defaultHeaders,
+          ...?headers,
+        },
         _httpClient = httpClient,
         _asyncStorage = asyncStorage,
         _flowType = flowType {
     _autoRefreshToken = autoRefreshToken ?? true;
 
+    final gotrueUrl = url ?? Constants.defaultGotrueUrl;
     _log.config(
         'Initialize GoTrueClient v$version with url: $_url, autoRefreshToken: $_autoRefreshToken, flowType: $_flowType, tickDuration: ${Constants.autoRefreshTickDuration}, tickThreshold: ${Constants.autoRefreshTickThreshold}');
-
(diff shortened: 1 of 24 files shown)


Vinzent03:
@dshukertjr For now, I've only added logging to the gotrue package. What do you think of this? If you consent, I would add this to the rest of the packages.

dshukertjr:
This is awesome! I love it so much!

Vinzent03:
I've noticed that we should be careful with what data we log, especially sensitive data. So I've come up with the idea that records with level `FINEST` contain sensitive data like session data, or which postgrest urls are called. So that those records shouldn't be sent to sentry or similar without making such things clear in privacy notes.  Let me know what you think of this approach and the documentation in the README of the `supabase_flutter` package, so I can add that section to the other packages as well.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -87,6 +89,8 @@ class GoTrueClient {
 
   final AuthFlowType _flowType;
 
+  final _log = Logger('supabase.gotrue');
```
Vinzent03, line 92: We could also change this to `supabase.auth`, since I've noticed the term `gotrue` is more and more replaced by `auth` at supabase.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -101,20 +105,22 @@ class GoTrueClient {
     GotrueAsyncStorage? asyncStorage,
     AuthFlowType flowType = AuthFlowType.pkce,
   })  : _url = url ?? Constants.defaultGotrueUrl,
-        _headers = headers ?? {},
+        _headers = {
```
Vinzent03, line 108: I've noticed a bug here, which has nothing to do with logging. The default headers weren't used for the non admin api calls. So the client version header was missing.

Vinzent03:
I see now that there is no documentation in the README of the rest of the packages. So I would leave it for now only in the supabase_flutter package and you? or I? add it to the supabase website documentation.

PR #1053, opened 2024-10-04
feat: Add saveSession option in goture_client signUp method

htoopyaelinn56:

Feature

When the user successfully signup, goture_client will automatically saves the user session.

## What is the new behavior?

Ability to set the saveSession option in case we don't want to save the user's session after signup for some app that the user have to login after successful signup.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -190,6 +190,8 @@ class GoTrueClient {
   /// [data] sets [User.userMetadata] without an extra call to [updateUser]
   ///
   /// [channel] Messaging channel to use (e.g. whatsapp or sms)
+  /// 
+  /// [saveSession] save the user session after signUp
   Future<AuthResponse> signUp({
     String? email,
     String? phone,
@@ -198,6 +200,7 @@ class GoTrueClient {
     Map<String, dynamic>? data,
     String? captchaToken,
     OtpChannel channel = OtpChannel.sms,
+    bool saveSession = true,
   }) async {
     assert((email != null && phone == null) || (email == null && phone != null),
         'You must provide either an email or phone number');
@@ -252,7 +255,7 @@ class GoTrueClient {
     final authResponse = AuthResponse.fromJson(response);
 
     final session = authResponse.session;
-    if (session != null) {
+    if (session != null && saveSession) {
       _saveSession(session);
       notifyAllSubscribers(AuthChangeEvent.signedIn);
     }


dshukertjr:
> for some app that the user have to login after successful signup.

Why would one want to add extra steps for the users? I guess it's not the same, but you can set the `autoRefreshToken` to false if you don't want the session to be refreshed once the token expires. 

```dart
await Supabase.initialize(
  url: supabaseUrl,
  anonKey: supabaseKey,
  debug: false,
  authOptions: FlutterAuthClientOptions(
    autoRefreshToken: false,
  ),
);
```

htoopyaelinn56:
> Why would one want to add extra steps for the users? 

I think it is just about app behavior. Most apps including GitHub asks a user to login again after successful sign up.

```autoRefreshToken``` will not do this behavior I guess.

dshukertjr:
@htoopyaelinn56 Hmm, why do you want the user to sign in again when the user is already signed in? There is no real benefit to having them sign out right after they sign up. If you really want to do this today, you could just call `signOut()` right after you complete the `signUp()` call, but I really don't see why one would want to do that.

htoopyaelinn56:
This is about applications' business logic other than benefits. As I mentioned, some apps including even GitHub asks a user to login again after the registration. Justing calling ```signOut()``` after ```signUp()``` might not be a good practice, it could be better not saving user's session at the first place for those who don't want to save the session of the user and want the user to re-login after signup. I set the newly added parameter ```saveSession``` to true so it still works as expected as before without breaking change.

dshukertjr:
Sorry, my comment [here](https://github.com/supabase/supabase-flutter/pull/1053#issuecomment-2392657563) stands and I'm closing this.

awalias:
hey @htoopyaelinn56 . I can see the use-case for this but I think it's important that it happens on the server and not on the client (since a malicious actor could still access the session token in the API response even if you decide to drop it in the app).

Currently to my knowledge the session token is _not_ returned when awaiting email signup confirmation. Which signup mechanism are you using?

awalias:
@kangmingtay I have a question. If [SECURITY_UPDATE_PASSWORD_REQUIRE_REAUTHENTICATION](https://github.com/supabase/auth?tab=readme-ov-file#reauthentication) is set to true, on sign up - will the user be required to re-authenticate? or does this config only affect password resets?

If it doesn't affect signups (for cases where MAILER_AUTOCONFIRM is true), perhaps it could be useful functionality 🤔. Although it may have to be a new config, to give more granular control.

htoopyaelinn56:
Hi @awalias. Thanks for the response. I'm using email auth and set auto confirmation to true. We have the existing auth service and use supabase auth just to access database. Sorry if my use-case is a bit complex.

PR #1067, opened 2024-10-21
add tiktok provider

dhamanutd:

Relate: https://github.com/supabase/auth/pull/1807
Add tiktok as oauth provider

There is no option for tiktok

## What is the new behavior?

Option for oauth with tiktok

## Additional context

### DIFF
--- a/packages/gotrue/lib/src/types/types.dart
+++ b/packages/gotrue/lib/src/types/types.dart
@@ -26,6 +26,7 @@ enum OAuthProvider {
   notion,
   slack,
   spotify,
+  tiktok,
   twitch,
   twitter,
   workos,


packages/gotrue/lib/src/types/types.dart
```diff
@@ -29,6 +29,7 @@ enum OAuthProvider {
   spotify,
   twitch,
   twitter,
+  twitter,
```
dshukertjr, line 32: This should be tiktok, right?

dshukertjr (changes requested):
Let's hold off on this one until the auth PR is merged.

packages/gotrue/lib/src/types/types.dart
dhamanutd: @dshukertjr sorry, committing wrong changes

Vinzent03:
I would close this pr for now since the auth pr is stale for over a year now.

Issue #1045, opened 2024-09-25
Keycloak provider in signInWithIdToken method in the gotrue package


Ppito:
I'm using Keycloak as my authentication provider and we need to use the `signInWithIdToken` method in our implementation.
The problem is that the gotrue package limits this method to the Google, Apple and Kakao providers : [gotrue_client.dart](https://github.com/supabase/supabase-flutter/blob/main/packages/gotrue/lib/src/gotrue_client.dart#L392)
Looking at the NodeJs implementation of this same package and function, I noticed that there was no particular blockage to using the Keycloak provider : [GoTrueClient.ts](https://github.com/supabase/auth-js/blob/master/src/GoTrueClient.ts#L631).

And indeed, by authorising Keycloak it works without any problem.

So my question is this: can you add this provider to the list of authorised providers?

dshukertjr:
Sure, would you like to open a PR? Note that it seems like Keycloak is deprecated on the JS library, so there might be chances that it won't be supported in the future.

Ppito:
I'll be unavailable for the next 3 weeks, so if no-one in my team has done a PR I'll do it when I get back.

Keycloak is not deprecated, quite the contrary. He has joined the Cloud Native Computing Foundation and is strongly supported by the RedHat team.
They have restructured the various dev/tools they had and the old SDK is effectively deprecated but not the new one.

Ppito:
PR Merged. 

Thanks again.

PR #1068, opened 2024-10-21
feat: Add Keycloak Provider on signInWithIdToken

Ppito:

Relate: https://github.com/supabase/supabase-flutter/issues/1045
Add Keycloak Provider on signInWithIdToken

signInWithIdToken are only available for Google, Apple or Kakao

## What is the new behavior?

Authorize Keycloak to use signInWithIdToken

## Additional context
We need a transparent connection that uses the accessToken and idToken that provides us with keycloak authentication.
As the connection has already been established through our systems, we need to provide the authentication information to gotrue directly using the signInWithIdToken method.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -382,7 +382,7 @@ class GoTrueClient {
 
   /// Allows signing in with an ID token issued by certain supported providers.
   /// The [idToken] is verified for validity and a new session is established.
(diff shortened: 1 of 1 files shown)


dshukertjr (commented):
Thanks for the PR! It looks good to me, but could we put keycloak at the end of each of them instead of at the beginning?

Ppito:
Yes, I'll do that.

Ppito:
done. :D

dshukertjr (approved):
Thanks for this! Looks great!

PR #1070, opened 2024-10-22
fix: Add equality to user attributes classes

bkalil:

Fix both `UserAttributes` and `AdminUserAttributes` classes by adding equality.

```dart
final attributeOne = UserAttributes(password: newPassword);
final attributeTwo = UserAttributes(password: newPassword);

attributeOne == attributeTwo // false
```

## What is the new behavior?

```dart
final attributeOne = UserAttributes(password: newPassword);
final attributeTwo = UserAttributes(password: newPassword);

attributeOne == attributeTwo // true
```

## Additional context

Encountered this behaviour when I tried to verify if my mock was called with the right values.

```dart
verify(() => mockAuthClient.updateUser(supa.UserAttributes(password: newPassword))); // failed
``` 
I then added the equality to the class in my pub cache and it worked.

### DIFF
--- a/packages/gotrue/lib/src/types/user_attributes.dart
+++ b/packages/gotrue/lib/src/types/user_attributes.dart
@@ -1,4 +1,3 @@
-// ignore_for_file: public_member_api_docs, sort_constructors_first
 import 'package:collection/collection.dart';
 
 class UserAttributes {
--- a/packages/gotrue/lib/src/types/user_attributes.dart
+++ b/packages/gotrue/lib/src/types/user_attributes.dart
@@ -1,3 +1,6 @@
+// ignore_for_file: public_member_api_docs, sort_constructors_first
+import 'package:collection/collection.dart';
+
 class UserAttributes {
   /// The user's email.
   String? email;
@@ -34,6 +37,26 @@ class UserAttributes {
       if (password != null) 'password': password,
       if (data != null) 'data': data,
     };
+  }
+
+  @override
+  bool operator ==(covariant UserAttributes other) {
+    if (identical(this, other)) return true;
+
+    return other.email == email &&
+        other.phone == phone &&
+        other.password == password &&
+        other.nonce == nonce &&
+        other.data == data;
+  }
+
+  @override
+  int get hashCode {
+    return email.hashCode ^
+        phone.hashCode ^
+        password.hashCode ^
+        nonce.hashCode ^
+        data.hashCode;
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/gotrue/lib/src/types/user_attributes.dart
```diff
@@ -35,6 +38,26 @@ class UserAttributes {
       if (data != null) 'data': data,
     };
   }
+
+  @override
+  bool operator ==(covariant UserAttributes other) {
```
Vinzent03, line 43: I don't like the covariant keyword here. It should be possible to compare any Object with this object, like when the comparing object is of type `dynamic` at compile time, this shouldn't throw an error at runtime.

packages/gotrue/lib/src/types/user_attributes.dart
```diff
@@ -35,6 +38,26 @@ class UserAttributes {
...
+    if (identical(this, other)) return true;
+
+    return other.email == email &&
+        other.phone == phone &&
+        other.password == password &&
+        other.nonce == nonce &&
+        other.data == data;
```
Vinzent03, line 50: I think `data` is very often a `Map`, which wouldn't be equal to another Map instance. So I would just use `DeepCollectionEquality()` here instead.

packages/gotrue/lib/src/types/user_attributes.dart
```diff
@@ -102,4 +125,25 @@ class AdminUserAttributes extends UserAttributes {
       if (banDuration != null) 'ban_duration': banDuration,
     };
   }
+
+  @override
+  bool operator ==(covariant AdminUserAttributes other) {
+    if (identical(this, other)) return true;
+    final mapEquals = const DeepCollectionEquality().equals;
+
+    return mapEquals(other.userMetadata, userMetadata) &&
```
Vinzent03, line 134: Didn't you miss the types of the `UserAttributes` class here and in `hashCode`?

Vinzent03 (changes requested):

packages/gotrue/lib/src/types/user_attributes.dart
```diff
@@ -1,3 +1,6 @@
+// ignore_for_file: public_member_api_docs, sort_constructors_first
```
Vinzent03, line 1: Sorry for the late review. I think this ignore comment can be removed here.

Vinzent03 (changes requested):

Vinzent03 (approved):

PR #1073, opened 2024-10-30
fix: Support custom access token

### LINKED ISSUE #1072: `Supabase.initialize` throws an `AuthException` when an `accessToken` is provided
**Describe the bug**
Hello! I am trying to provide an Auth0 access token to Supabase, but find myself unable to do so. After a little debugging I discovered that the Supabase initialize code is still initializing an instance of `SupabaseAuth` even with an accessCode, which references `Supabase.instance.client.auth`.
I believe this was introduced in #999 

**To Reproduce**
1. Initialize Supabase with an access token

**Expected behavior**
`Supabase.initialize` should accept a custom access token.

**Screenshots**
Here is the call stack for this bug:
image

Here is the exception being thrown:
image

**Version:**
On Linux/macOS
```
├── supabase_flutter 2.7.0
│   ├── supabase 2.4.0
│   │   ├── functions_client 2.3.3
│   │   ├── gotrue 2.9.0
│   │   ├── postgrest 2.2.0
│   │   ├── realtime_client 2.3.0
│   │   ├── storage_client 2.1.0
```

**Additional context**
For context, here is the code where I am initializing Supabase:
```dart
  final auth0 = Auth0(Environment.auth0Domain, Environment.auth0ClientId);

  await Supabase.initialize(
    url: Environment.supabaseUrl,
    anonKey: Environment.supabaseAnonKey,
    accessToken: () async {
      final credentials = await auth0.credentialsManager.credentials();
      return credentials.accessToken;
    },
  );
```

Here is the exception message:
```
Unhandled Exception: AuthException(message: Supabase Client is configured with the accessToken option, accessing supabase.auth is not possible., statusCode: null, errorCode: null)
```

Vinzent03:

Bug fix

Passing a custom access token to supabase_flutter doesn't work. I noticed that the `SupabaseClient.dispose()` method didn't work with custom access token as well.

## What is the new behavior?

The realtime reconnection logic is moved to the `Supabase` class and `SupabaseAuth` is only initialized when no custom `accessToken` is passed.

## Additional context

close #1072

### DIFF
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -52,7 +52,7 @@ class SupabaseClient {
   final Client? _httpClient;
   late final Client _authHttpClient;
 
-  late final GoTrueClient _authInstance;
+  GoTrueClient? _authInstance;
 
   /// Supabase Functions allows you to deploy and invoke edge functions.
   late final FunctionsClient functions;
@@ -160,7 +160,7 @@ class SupabaseClient {
 
   GoTrueClient get auth {
     if (accessToken == null) {
-      return _authInstance;
+      return _authInstance!;
     } else {
       throw AuthException(
         'Supabase Client is configured with the accessToken option, accessing supabase.auth is not possible.',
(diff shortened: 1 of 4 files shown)


dshukertjr:
Thanks for this. I might not have the time to review it today, but will get to it as soon as possible!

dshukertjr (approved):
Sorry for the massive delay on this one. Thanks for the fix!

gbaladi:
@dshukertjr hi, is there any timetable that a new version of the client will be published with this fix?

PR #1087, opened 2024-11-21
refactor: combine the storage interface for storage and pkce in gotrue_client

### LINKED ISSUE #1085: Version 2.7.0 prevents reading data on web under certain conditions
**Describe the bug**
Under certain conditions a client with a service role key can not circumvent RLS policy restrictions on web.

**To Reproduce**
Add the following code to a flutter project:

```dart
// integration_test/supabase_test.dart

import 'package:flutter_test/flutter_test.dart';
import 'package:integration_test/integration_test.dart';
import 'package:supabase_flutter/supabase_flutter.dart' as supabase;

void main() async {
  IntegrationTestWidgetsFlutterBinding.ensureInitialized();

  testWidgets('Analytics', (WidgetTester tester) async {
    final client1 = supabase.SupabaseClient(
      getSupabaseUrl(),
      getSupabaseKey(role: 'service'),
    );

    final client2 = supabase.SupabaseClient(
      getSupabaseUrl(),
      getSupabaseKey(),
    );

    await tester.pumpWidget(
      const MaterialApp(
        home: Scaffold(
          body: Center(child: Text('Hello')),
        ),
      ),
    );

    // Commenting out this line makes the test pass
    await client2.auth.signInAnonymously();

    // Commenting out this line makes the test pass
    await client1.from('session_events').select();

    final events = await client1.from('session_events').select();

    expect(events, isNotEmpty, reason: 'Session events empty');
  });
}
```

```dart
// test_driver/integration_test.dart

import 'package:flutter_driver/flutter_driver.dart';
import 'package:integration_test/integration_test_driver_extended.dart';

Future<void> main() async {
  return integrationDriver(driver: await FlutterDriver.connect());
}
```

Create a table "session_events" that has an RLS policy preventing everyone from reading it.

Run the test on Linux with `flutter test -d linux integration_test` and on Chrome with `flutter drive --driver=test_driver/integration_test.dart --target=integration_test/supabase_test.dart -d web-server --browser-name chrome --no-headless`. Make sure to first start ChromeDriver on port 4444.

**Observed behavior**

- Run on Chrome -> The test fails
- Run on Linux -> The test passes
- Comment out the anonymous sign-in -> The test passes
- Comment out the first read -> The test passes
- Run on Chrome with version 2.6.0 -> Test passes

```
├── supabase 2.4.0
│   ├── functions_client 2.3.3
│   ├── gotrue 2.9.0
│   ├── postgrest 2.2.0
│   ├── realtime_client 2.3.0
│   ├── storage_client 2.1.0
├── supabase_flutter 2.7.0
│   ├── supabase...
```

Vinzent03:

refactor with breaking changes and fixes

The session gets stored by the supabase_flutter package by listening to auth events. The session is stored by a different class/interface than the pkce token. This adds unnecessary complexity and is a difference to the supabase-js sdk.

## What is the new behavior?

The gotrue package manages the session storage itself (like in supabase-js) and uses the same interface as the pkce token.

Reasons for this pr:
- Have a more similar approach to saving the session like in supabase-js
- Make it less confusing for the user to have to provide two different storage interfaces to customize the storage of the session and the pkce token
- Provide a built-in interface to persist the session if not using supabase_flutter, e.g. when building a cli with sessions or I suppose a [jaspr](https://docs.jaspr.site) website.
- The auth client has now the `persistSession` field, which is now used to explicitly turn on the session broadcasting. fixes #1085

I deprecated both `localStorage` (used for session storage by `supabase_flutter`) and `pkceAsyncStorage` (used by `supabase` to store the pkce token) and replace them with a unified storage interface `asyncStorage`, which uses the `GotrueAsyncStorage` class, which was previously used for pkce only. Though I refactored the arguments of the methods to use positional arguments instead of named arguments. This is much more idiomatic to other storage packages and common for methods with just one or two arguments. To support existing configuration as best as possible I created a new class, which combines both deprecated config options (`localStorage` and `pkceAsyncStorage`) into one and decides by the suffix of the storage key to use the session storage or pkce storage. Note that an implementation of the `GotrueAsyncStorage` class still needs to be refactored, but that only affects the method signature and not the implementation itself. 

Additionally, I noticed that many document comments were outdated, because of the 'new' options classes for each package, so I updated them.

## Breaking Changes

- Only the change of the method signatures of the `GotrueAsyncStorage` class
```diff
-Future<void> setItem({required String key, required String value});
+Future<void> setItem(String key, String value);
```

## Additional context

I think this is a great refactoring that does not affect many people, because I don't think many flutter user customize the storage. But it makes it easier and more streamlined for the ones that do so and opens new easier usage ways outside of flutter with auth session storage now built-in.


dshukertjr:
Thanks for this PR, but honestly, I am not a huge fan of these refactorings. I feel like the library is deviating more and more from supabase-js, which makes it hard to maintain. Unless there is something broken, I'd say we keep it the way it is, and maybe we can revisit it when a major version release comes.

Vinzent03:
I'm quite confused, because the whole goal of this pr is to align with supabase-js .

- This now uses a single storage interface like supabase-js -> session and pkce are no longer handled separately like it's done in supabase-js
- This now uses the same storage interface like supabase-js
- The session is stored by the auth client itself like in auth-js instead of having to listen to the `onAuthStateChange` stream in supabase_flutter.
- The session is only broadcasted if the session is persisted, just like in auth-js

So this pr makes it easier to maintain the similarity to the js sdk.
Example:
Beforehand it was not possible to disable the session broadcasting like in auth-js, which I already [mentioned in that pr](https://github.com/supabase/supabase-flutter/pull/1005#discussion_r1760268910), because the sdk were deviated. But with these changes it's implemented like in supabase-js

dshukertjr:
It's a refactoring that doesn't really solve any issues that users are asking for. We can probably wait for a major version bump for changes like this.

packages/supabase_flutter/lib/src/supabase.dart
```diff
@@ -195,7 +199,6 @@ class Supabase {
   /// Dispose the instance to free up resources.
   Future<void> dispose() async {
     _targetLifecycleState = null;
-    await _restoreSessionCancellableOperation.cancel();
```
Vinzent03: This code was actually not functional before as one cannot really cancel a future. Canceling this future would just call the `onCancel` function, but that is not specified here. So it would have just been fine to call `recoverSession()` directly.

spydon:
@Vinzent03 just fyi, I changed the base of this from the v3 branch to main so that I can remove the v3 branch.

spydon:
@Vinzent03 this was superseded by #1804 and #1805, I'd love your review on those ones!

PR #1097, opened 2024-12-10
fix: Ignore email and phone assertions when token hash is being verified

Dastaan2k:
[Supabase Docs](https://supabase.com/docs/guides/auth/auth-email-passwordless?queryGroups=language&language=dart#with-magic-link
)

Following the above link, verifying token hash from magic link doesn't require email or phone in request body
(only 'token hash' and 'type' is required in the js example)

Pre commit state of the code threw assertion error if neither email nor phone is present in verifyOtp call preventing users to verify token hash without passing email or phone in request body ("auth/verify" endpoint called inside this function throws an error as well if email or phone is passed in request body along with token hash and type)

Hence have added a condition that will ignore the first email and phone assertion if token hash is populate

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -533,7 +533,10 @@ class GoTrueClient {
     String? captchaToken,
     String? tokenHash,
   }) async {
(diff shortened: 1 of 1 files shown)


Vinzent03 (approved):
Thanks for your contribution!

PR #1100, opened 2024-12-13
fix: Handling non success status codes on "/verify" call

Dastaan2k:
Handling non success status codes from the "/verify" call caused when the magic link is invalid or expired.
Current commit will return AuthResponse with null user and session returning to the "not logged in" state.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -538,33 +538,38 @@ class GoTrueClient {
     assert(token != null || tokenHash != null,
         '`token` or `tokenHash` needs to be specified.');
 
-    final body = {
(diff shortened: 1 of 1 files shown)


Vinzent03:
I think the current implementation is better. By just catching every exception, the dev doesn't know what's wrong. If we actually would want to handle that case, we should not just return an empty session, but log the user out as well. 
By catching the exception yourself, you can still log out.

Vinzent03:
A link expiry could also be sent via an exception. But I took a look at the code and the `verifyOtp()` method is not called from the library anywhere. So there is no need to add an event to the stream at all? Since you are already responsible to call that method yourself?

Vinzent03:
I am going to close this for now. If you have any further comment, just leave them here and we might reopen the pr.

PR #1105, opened 2024-12-31
docs: Fix typo for RPC docs

bqubique:

Docs update.

Typo on RPC function docs where the current docs says:
```
/// [params] is an optinal object to pass as arguments to the function call.
```

## What is the new behavior?

Typo on RPC function docs where the current docs says:
```
/// [params] is an optional object to pass as arguments to the function call.
```

## Additional context


Vinzent03 (approved):
Thanks for the quick typo fix!

PR #1110, opened 2025-01-17
chore: Generate coverage reports and upload them.

dshukertjr:

Generate coverage reports for each package and upload them.

A new GitHub actions workflow file, `coverage.yml` is introduced, which runs the test for all the packages, generates coverage reports, combines them, and uploads them. This test is redundant as the tests are also run in each individual CI pipeline, but after struggling a bit trying to combine the coverage reports from multiple different workflows, I settled with this implementation.

This PR also includes minor type fix and test stub fix in the realtime rep.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,52 @@
+name: Combined Coverage Report
+
+on:
+  workflow_run:
+    workflows:
+      - 'functions_client'
+      - 'gotrue'
+      - 'postgrest'
+      - 'realtime_client'
+      - 'storage_client'
+      - 'supabase'
+      - 'supabase_flutter'
+      - 'yet_another_json_isolate'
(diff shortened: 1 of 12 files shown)


grdsdev:
@dshukertjr I see this has been marked as ready for review, but I see no report on coveralls, is it working already?

dshukertjr:
@grdsdev I think so. The upload action has succeeded: https://github.com/supabase/supabase-flutter/actions/runs/13031130088/job/36350498930?pr=1110

grdsdev:
@dshukertjr I think you still need to add a step for uploading to coveralls instead of GH artifacts, I see a 0% coverage on https://coveralls.io/github/supabase/supabase-flutter

dshukertjr:
Should be good now

grdsdev (approved):

PR #1118, opened 2025-02-14
feat: Add debugName parameter to YAJsonIsolate.

misha:

This is a new feature for `YAJsonIsolate`.

In testing environments, it's easy to have multiple clients open. Specifically, I usually have an admin client and a client for the current user, but I can't tell which is which because all isolates have the same name. When an error occurs in the isolate, it takes a lot of effort to figure out which client produced it.

I would like to be able to rename isolates, specifically for use in debugging.

Currently, all isolates spawned have the name `_compute`.

Here's a screenshot of what it looks like in VSCode.

before_change

## What is the new behavior?

There's now a `debugName` constructor parameter on `YAJsonIsolate`. The user can add the library to their package and instantiate an instance with their desired `debugName` if they want to access it. 

Here's a screenshot of what it looks like in VSCode, after setting the name to `my_isolate`.

after_change

## Additional context

Two questions:

1. There's also a web implementation in `_isolates_web.dart`. Does this now need the same (albeit unused) constructor parameter?
2. I didn't write a test. I could store and expose the underlying `Isolate` instance to verify it gets created with the specified `debugName`, but then I have to remember to unassign it so I'd rather just not store it. If a test is required, let me know what it should do.

### DIFF
--- a/packages/yet_another_json_isolate/lib/src/_isolates_io.dart
+++ b/packages/yet_another_json_isolate/lib/src/_isolates_io.dart
@@ -8,6 +8,13 @@ import 'package:async/async.dart';
 
 // One instance manages one isolate
 class YAJsonIsolate {
+  YAJsonIsolate({
+    this.debugName,
+  });
+
+  /// The debug name used for the isolate spawned by this instance.
+  final String? debugName;
+
   final _receivePort = ReceivePort();
   late final SendPort _sendPort;
   final _createdIsolate = Completer<void>();
@@ -26,6 +33,7 @@ class YAJsonIsolate {
       _receivePort.sendPort,
       onExit: _receivePort.sendPort,
       onError: _receivePort.sendPort,
+      debugName: debugName,
     );
     _sendPort = await _events.next;
     _createdIsolate.complete();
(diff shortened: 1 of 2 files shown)


dshukertjr:
Thanks for the PR! A test would be great! I don't think we need to worry about the web implementation!

Vinzent03:
I'm pretty sure we have to change the web part as well. Otherwise the signature is different and the build should fail for web.

misha:
> Thanks for the PR! A test would be great! I don't think we need to worry about the web implementation!

Can you mention how the test should be implemented?

The only way I can think to test this is to store the `Isolate` instance, expose it for testing, and then validate the name on the isolate itself. This would require test-visible access to the isolate, ie. via `@visibleForTesting`. Presently, the implementation doesn't even store the spawned isolate, much less expose it in any way. I would say it's not worth the risk to test a non-critical, debugging feature.

> I'm pretty sure we have to change the web part as well. Otherwise the signature is different and the build should fail for web.

Yep, that sounds right to me. I added a commit to make the constructor signatures identical.

Vinzent03 (approved):
I don't think this requires any further tests. The parameter is only passed to the `spawn` method. So it should be fine.

PR #1123, opened 2025-03-05
chore: Test empty local storage

dshukertjr:

Just a PR to increase coverage. More coverage-increase PR to come.

### DIFF
--- a/packages/postgrest/lib/src/types.dart
+++ b/packages/postgrest/lib/src/types.dart
@@ -88,9 +88,9 @@ enum CountOption {
   estimated,
 }
 
-// Not used anywhere. Will be removed in the next major version.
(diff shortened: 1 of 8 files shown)


packages/postgrest/lib/src/types.dart
```diff
@@ -88,11 +88,14 @@ enum CountOption {
   estimated,
 }
 
+// Not used anywhere. Will be removed in the next major version.
+// coverage:ignore-[start]
 /// Returns count as part of the response when specified.
 enum ReturningOption {
```
Vinzent03, line 94: What do you think about adding a `@Deprecated()` like I've done already with similar cases? Makes it easier in the future to find those again.

Vinzent03 (approved):

PR #1130, opened 2025-03-12
feat: Add standard client headers

grdsdev:

## What is the new behavior?

## Additional context

### DIFF
--- a/packages/supabase/lib/src/constants.dart
+++ b/packages/supabase/lib/src/constants.dart
@@ -1,7 +1,18 @@
 import 'package:supabase/src/version.dart';
+import 'dart:io' show Platform;
 
 class Constants {
-  static const Map<String, String> defaultHeaders = {
+  static String get platform {
+    return Platform.operatingSystem;
+  }
+
+  static String get platformVersion {
+    return Platform.operatingSystemVersion;
+  }
+
+  static final Map<String, String> defaultHeaders = {
     'X-Client-Info': 'supabase-dart/$version',
+    'X-Supabase-Client-Platform': platform,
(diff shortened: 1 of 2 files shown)


packages/supabase/lib/src/constants.dart
```diff
@@ -1,7 +1,18 @@
 import 'package:supabase/src/version.dart';
+import 'dart:io' show Platform;
 
 class Constants {
-  static const Map<String, String> defaultHeaders = {
+  static String get platform {
+    return Platform.operatingSystem;
```
Vinzent03, line 6: I know this is still draft, but still want to mention that this won't work on web.
You will either need a conditional import, or probably better use a constant to guard web with a definition like the `kIsWeb` in flutter:
```dart
const bool kIsWeb = bool.fromEnvironment('dart.library.js_util');
```

grdsdev: Thanks, @Vinzent03 

Do we have any test for asserting that it doesn't break in Web? Didn't see any fail on CI.

Vinzent03: Sadly not yet, but hopefully #1140 solves this

grdsdev: @Vinzent03 I added the guard as you suggested, we could send data in case of web, if that is easy to do, not a requirement since our reports also fetches from the `user-agent`, and `user-agent` in web is better defined then mobile.

Vinzent03: Yeah, I think using the user-agent is the better and correct way for web.

Vinzent03 (approved):

PR #1135, opened 2025-03-18
feat(gotrue, supabase, supabase_flutter, realtime_client): Use web package to access web APIs

dshukertjr:

This PR [migrates gotrue to package:web](https://dart.dev/interop/js-interop/package-web).

Because of version constraints issues, this bumps the minimum Dart version to 3.3.0 and Flutter version to 3.19.0, which has been released in Feb 16, 2024.

Uses the old html API, which causes analysis issue, and is not recommended to keep by the Flutter team.

## What is the new behavior?

The new web package is used.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -11,10 +11,10 @@ import 'package:gotrue/src/types/auth_response.dart';
 import 'package:gotrue/src/types/fetch_options.dart';
 import 'package:http/http.dart';
 import 'package:jwt_decode/jwt_decode.dart';
+import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:retry/retry.dart';
 import 'package:rxdart/subjects.dart';
-import 'package:logging/logging.dart';
 
 import 'broadcast_stub.dart' if (dart.library.html) './broadcast_web.dart'
     as web;
@@ -363,8 +363,8 @@ class GoTrueClient {
       ),
     );
 
-    await _asyncStorage!
-        .removeItem(key: '${Constants.defaultStorageKey}-code-verifier');
+    await _asyncStorage.removeItem(
+        key: '${Constants.defaultStorageKey}-code-verifier');
 
     final authSessionUrlResponse = AuthSessionUrlResponse(
         session: Session.fromJson(response)!, redirectType: redirectType?.name);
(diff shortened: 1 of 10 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -11,10 +11,10 @@ import 'package:gotrue/src/types/auth_response.dart';
 import 'package:gotrue/src/types/fetch_options.dart';
 import 'package:http/http.dart';
 import 'package:jwt_decode/jwt_decode.dart';
+import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:retry/retry.dart';
 import 'package:rxdart/subjects.dart';
-import 'package:logging/logging.dart';
 
 import 'broadcast_stub.dart' if (dart.library.html) './broadcast_web.dart'
```
Vinzent03, line 19: This has to be changed as well, doesn't it?

Vinzent03 (changes requested):

packages/gotrue/lib/src/gotrue_client.dart
dshukertjr: I think this import just got moved to line 14 automatically from my IDE.

Vinzent03: No, I meant the `dart.library.html` import: https://dart.dev/interop/js-interop/package-web#conditional-imports

dshukertjr: Should be good now 👍

Vinzent03 (changes requested):

packages/realtime_client/lib/src/websocket/websocket.dart
```diff
@@ -1,3 +1,3 @@
 export 'websocket_stub.dart'
     if (dart.library.io) 'websocket_io.dart'
-    if (dart.library.html) 'websocket_web.dart';
+    if (bool.fromEnvironment) 'websocket_web.dart';
```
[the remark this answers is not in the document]
dshukertjr: Thanks for the catch!

Vinzent03 (changes requested):
There is still a html import here: https://github.com/supabase/supabase-flutter/blob/fix/gotrue-web/packages/supabase_flutter/lib/src/local_storage_web.dart#L2-L2

and this comment: https://github.com/supabase/supabase-flutter/blob/fix/gotrue-web/packages/gotrue/lib/src/types/auth_state.dart#L8-L8 could probably be improved as well now.

Vinzent03:
From a quick look at the [documentation](https://pub.dev/documentation/web/latest/web/Window/localStorage.html) it seems like a quick swap of the import and method names should be possible. You changed the behavior of the broadcast channel as well, so I don't quite get the differentiation to this migration here...

Since this is a breaking change due to the changed dart sdk requirements, I would at least release this pr as a new minor version.

Vinzent03 (approved):

PR #1143, opened 2025-03-31
fix(gotrue): Handle empty error response

Vinzent03:

Bug fix

When receiving an error response, which doesn't contain any body, the decoding to json throws a `FormatException`. This is very misleading, because we shouldn't try to decode an empty response. This happened to me while doing some testing, but I do not exactly remember the reason for the error anymore.

## What is the new behavior?

Do not decode an empty response.

In addition, I added a `toString` for every subclass of `AuthException`.


grdsdev (approved):

dshukertjr (approved):
Nice! Thanks for this!

PR #1148, opened 2025-04-04
fix(gotrue): AuthException .toString method

TillJohanndeiter:

I was a little bit confused by the logs because the AuthException doesn't contain a variable `errorCode`.

The `toString` output is not aligned with the variable name

## What is the new behavior?

The `toString` output is aligned with the variable name

## Additional context

_None_


dshukertjr (approved):
Thanks for this!

PR #1162, opened 2025-05-02
ci: Run CI on all platforms

grdsdev:

Run CI on all available platforms, and verify examples builds.


dshukertjr (commented):
Do we want to add Android to the list of builds?

grdsdev:
> Do we want to add Android to the list of builds?

Yes, missed that one.

dshukertjr (approved):

PR #1164, opened 2025-05-04
fix: Adds support for SharedPreferencesAsync

romaingyh:

It's both a bug fix and feature.

9 months ago with version 2.3.0, `shared_preferences` package added `SharedPreferencesAsync` and `SharedPreferencesWithCache` to replace the legacy (and deprecated in future) `SharedPreferences`. [See here](https://pub.dev/packages/shared_preferences#sharedpreferences-vs-sharedpreferencesasync-vs-sharedpreferenceswithcache)

Supabase flutter uses by default the legacy one with `SharedPreferencesLocalStorage`.

Problem is that on some platform like Windows, the legacy one and new one are not compatible. If the developer made the migration to `SharedPreferencesAsync` then the supabase's local storage is broken.

_Example :_
I'm using `SharedPreferencesAsync` in my app. After sign in, the `sb-x-auth-token` is saved in `saved-preferences.json` by supabase's `SharedPreferences`. After that, any call to my app's `SharedPreferencesAsync` will overwrite the token and user have to sign in on next app launch. In practice it can lead to sign in every time he opens the app if you use shared preferences frequently in your code.

## What is the new behavior?

I added a boolean flag `useSharedPreferencesAsync` to `SharedPreferencesLocalStorage`. There wasn't very much changes as every methods are already async in LocalStorage interface.

If you prefer two distinct classes `SharedPreferencesLocalStorage` and `SharedPreferencesAsyncLocalStorage` I can do this

### DIFF
--- a/packages/supabase_flutter/lib/src/local_storage.dart
+++ b/packages/supabase_flutter/lib/src/local_storage.dart
@@ -61,66 +61,12 @@ class EmptyLocalStorage extends LocalStorage {
   Future<void> persistSession(persistSessionString) async {}
 }
 
-/// A [LocalStorage] implementation that implements SharedPreferences as the
-/// storage method.
-class SharedPreferencesLocalStorage extends LocalStorage {
-  late final SharedPreferences _prefs;
-
-  SharedPreferencesLocalStorage({required this.persistSessionKey});
-
-  final String persistSessionKey;
-
-  static const _useWebLocalStorage =
-      kIsWeb && bool.fromEnvironment("dart.library.js_interop");
-
-  @override
-  Future<void> initialize() async {
-    if (!_useWebLocalStorage) {
-      WidgetsFlutterBinding.ensureInitialized();
-      _prefs = await SharedPreferences.getInstance();
-    }
-  }
-
-  @override
-  Future<bool> hasAccessToken() async {
-    if (_useWebLocalStorage) {
-      return web.hasAccessToken(persistSessionKey);
-    }
-    return _prefs.containsKey(persistSessionKey);
-  }
-
-  @override
-  Future<String?> accessToken() async {
-    if (_useWebLocalStorage) {
-      return web.accessToken(persistSessionKey);
-    }
-    return _prefs.getString(persistSessionKey);
-  }
-
-  @override
-  Future<void> removePersistedSession() async {
-    if (_useWebLocalStorage) {
-      web.removePersistedSession(persistSessionKey);
-    } else {
-      await _prefs.remove(persistSessionKey);
-    }
-  }
-
-  @override
-  Future<void> persistSession(String persistSessionString) {
-    if (_useWebLocalStorage) {
-      return web.persistSession(persistSessionKey, persistSessionString);
-    }
-    return _prefs.setString(persistSessionKey, persistSessionString);
-  }
-}
-
 /// A [LocalStorage] implementation that implements SharedPreferencesAsync as the
 /// storage method.
-class SharedPreferencesAsyncLocalStorage extends LocalStorage {
+class SharedPreferencesLocalStorage extends LocalStorage {
   late final SharedPreferencesAsync _prefs;
 
-  SharedPreferencesAsyncLocalStorage({required this.persistSessionKey});
+  SharedPreferencesLocalStorage({required this.persistSessionKey});
 
   final String persistSessionKey;
 
--- a/packages/supabase_flutter/lib/src/local_storage.dart
+++ b/packages/supabase_flutter/lib/src/local_storage.dart
@@ -64,18 +64,11 @@ class EmptyLocalStorage extends LocalStorage {
 /// A [LocalStorage] implementation that implements SharedPreferences as the
 /// storage method.
 class SharedPreferencesLocalStorage extends LocalStorage {
-  late final SharedPreferences _syncPrefs;
+  late final SharedPreferences _prefs;
 
-  late final SharedPreferencesAsync _asyncPrefs;
-
-  SharedPreferencesLocalStorage({
-    required this.persistSessionKey,
-    this.useSharedPreferencesAsync = false,
-  });
+  SharedPreferencesLocalStorage({required this.persistSessionKey});
 
   final String persistSessionKey;
-
-  final bool useSharedPreferencesAsync;
 
   static const _useWebLocalStorage =
       kIsWeb && bool.fromEnvironment("dart.library.js_interop");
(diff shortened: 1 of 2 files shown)


Vinzent03:
I can confirm that this issue exists, as I encountered this as well in a recent project, but didn't have the time to investigate more and come up with a pr. But I'm really unsure about the current design of the solution and will have to think about it a bit.

romaingyh:
> I can confirm that this issue exists, as I encountered this as well in a recent project, but didn't have the time to investigate more and come up with a pr. At that time, I solved it by manually providing my own storage implementation using the async shared preferences. But I'm really unsure about the current design of the solution and will have to think about it a bit.

I thought of another solution like this :

```dart
/// A [LocalStorage] implementation that implements SharedPreferences as the
/// storage method.
class SharedPreferencesLocalStorage extends LocalStorage {
  late final SharedPreferences _prefs;

  SharedPreferencesLocalStorage({required this.persistSessionKey});

  final String persistSessionKey;

  static const _useWebLocalStorage =
      kIsWeb && bool.fromEnvironment("dart.library.js_interop");

  @override
  Future<void> initialize() async {
    if (!_useWebLocalStorage) {
      _prefs = await SharedPreferences.getInstance();
    }
  }

  @override
  Future<bool> hasAccessToken() async {
    if (_useWebLocalStorage) {
      return web.hasAccessToken(persistSessionKey);
    }
    return _prefs.containsKey(persistSessionKey);
  }

  @override
  Future<String?> accessToken() async {
    if (_useWebLocalStorage) {
      return web.accessToken(persistSessionKey);
    }
    return _prefs.getString(persistSessionKey);
  }

  @override
  Future<void> removePersistedSession() async {
    if (_useWebLocalStorage) {
      web.removePersistedSession(persistSessionKey);
    } else {
      await _prefs.remove(persistSessionKey);
    }
  }

  @override
  Future<void> persistSession(String persistSessionString) {
    if (_useWebLocalStorage) {
      return web.persistSession(persistSessionKey, persistSessionString);
    }
    return _prefs.setString(persistSessionKey, persistSessionString);
  }
}

/// A [LocalStorage] implementation that implements SharedPreferencesAsync as the
/// storage method.
class SharedPreferencesAsyncLocalStorage extends LocalStorage {
  late final SharedPreferencesAsync _prefs;

  SharedPreferencesAsyncLocalStorage({required this.persistSessionKey});

  final String persistSessionKey;

  static const _useWebLocalStorage =
      kIsWeb && bool.fromEnvironment("dart.library.js_interop");

  @override
  Future<void> initialize() async {
    if (!_useWebLocalStorage) {
      _prefs = SharedPreferencesAsync();
    }
  }

  @override
  Future<bool> hasAccessToken() async {
    if (_useWebLocalStorage) {
      return web.hasAccessToken(persistSessionKey);
    }
    return _prefs.containsKey(persistSessionKey);
  }

  @override
  Future<String?> accessToken() async {
    if (_useWebLocalStorage) {
      return web.accessToken(persistSessionKey);
    }
    return _prefs.getString(persistSessionKey);
  }

  @override
  Future<void> removePersistedSession() async {
    if (_useWebLocalStorage) {
      web.removePersistedSession(persistSessionKey);
    } else {
      await _prefs.remove(persistSessionKey);
    }
  }

  @override
  Future<void> persistSession(String persistSessionString) {
    if (_useWebLocalStorage) {
      return web.persistSession(persistSessionKey, persistSessionString);
    }
    return _prefs.setString(persistSessionKey, persistSessionString);
  }
}
```

where the dev has to opt-in for `SharedPreferencesAsyncLocalStorage` when he initializes supabase.

I didn't choose this implementation in my PR because the two classes are pretty much the same except for shared prefs init.

romaingyh:
I think CI test fails because with Flutter 3.19 the resolvable `shared_preferences` package version is < 2.3.0 but not sure about this

sai-chandra22:
@Vinzent03 , Can you exactly tell me what is the solution you have done for this solution

Vinzent03:
If we want to provide a `LocalStorage` implementation that uses `SharedPreferencesAsync` we need `shared_preferences` v2.3.0, which requires Dart v3.4.0, but we currently test and support until Dart v3.3.0. So we would need to bump Flutter to v3.22.0 for Dart v3.4.0. But this version is already one year old and this is an actual issue, because the session persistence is broken.
The only other way would be to only provide an example implementation for users to copy and add themselves.
But because we can't detect whether they use the new or old api we would have to stick with the old api for the moment and mention it in the docs that if they use the new shared_preferences api they need the new `LocalStorage` implementation anyway.
@dshukertjr What is your opinion?

Either way @romaingyh I prefer the two distinct classes implementation.

dshukertjr:
@Vinzent03 
>But this version is already one year old and this is an actual issue, because the session persistence is broken.

Broken in what way?

dshukertjr:
@romaingyh I know this blows up the scope of this PR, but I'm okay bumping the minimum Flutter/ Dart version to 3.220/ 3.4.0 and to just use `SharedPreferencesAsync` from now on. One thing that we should be doing is upon initialization, we make sure that existing tokens in the legacy `SharedPreferences` are transferred to `SharedPreferencesAsync` upon initialization.

romaingyh:
@dshukertjr I bumped flutter/dart to 3.22/3.4.0 and replaced SharedPreferences with SharedPreferencesAsync. I also added the token migration at init but not tested

Vinzent03:
But using the new async variant by default causes existing usage in the user's code of the old variant to break, doesn't it? This is a heavy breaking change.

romaingyh:
Yes as Async prefs overwrite the legacy prefs (I'm sure on windows but not for other platforms) I think there is a risk of user losing it's preferences if he upgrade supabase but still use legacy prefs in his code

romaingyh:
I rebased

longtimedeveloper:
@romaingyh do you know when your fix will be available in the public release of Supabase?  Thank you.  

I updated my app today and broke this Windows sign in as described here.

romaingyh:
> @romaingyh do you know when your fix will be available in the public release of Supabase?  Thank you.  
> 
> 
> 
> I updated my app today and broke this Windows sign in as described here.

Sorry no I'm not an officiel maintainer of this repo. We're currently waiting for the main maintainer to publish new releases.

Here is what I'm doing for now. In pubspec : 

```yaml
  supabase_flutter:
    git:
      url: https://github.com/romaingyh/supabase-flutter.git
      ref: shared_prefs_async
      path: packages/supabase_flutter
```

and use this in supabase initialization auth options : 

```dart
       localStorage: SharedPreferencesLocalStorage(
          persistSessionKey: "sb-${Uri.parse(url).host.split(".").first}-auth-token",
        ),
```

You can also override supabase to have the last fixes : 

```yaml
  # TODO: Remove when supabase > 2.8.0 published on pub.dev, including https://github.com/supabase/supabase-flutter/pull/1196
  supabase:
    git:
      url: https://github.com/romaingyh/supabase-flutter.git
      path: packages/supabase
```

spydon:
@romaingyh thanks for your work on this, I missed that it was auto closed.
It has now been implemented in #1680 and will be available in the v3 version of the SDK.

PR #1168, opened 2025-05-07
refactor(gotrue): Remove unused _currentUser field and update currentUser getter logic. 

juampiq6:

Refactor

_currentUser variable is just being stored as a copy of _currentSession.user

## What is the new behavior?

The currentUser getter is handling the indirection correctly, no need to duplicate values and assign both each time they change.

## Additional context

I have simplified a comment (it was a double negation).
I have added some bang operators in cases where null-aware operator where not needed

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -43,9 +43,6 @@ class GoTrueClient {
   /// Namespace for the GoTrue MFA API methods.
   late final GoTrueMFAApi mfa;
 
-  /// The currently logged in user or null.
-  User? _currentUser;
-
   /// The session object for the currently logged in user or null.
   Session? _currentSession;
 
@@ -140,10 +137,10 @@ class GoTrueClient {
   /// Returns the current logged in user, if any;
   ///
   /// Use [currentSession] to determine whether the user has an active session,
-  /// because [currentUser] can be non-null without an active session, such as
+  /// because [currentUser] can be null with an active session, such as
   /// when the user signed up using email and password but has not confirmed
   /// their email address.
-  User? get currentUser => _currentUser;
+  User? get currentUser => _currentSession?.user;
 
   /// Returns the current session, if any;
   Session? get currentSession => _currentSession;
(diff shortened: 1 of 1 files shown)


packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -997,7 +992,7 @@ class GoTrueClient {
         }
       } else {
         final shouldEmitEvent = _currentSession == null ||
```
juampiq6, line 994: if current session is null, then the second bang operator in the expression will never reach

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -748,8 +745,7 @@ class GoTrueClient {
         options: options);
     final userResponse = UserResponse.fromJson(response);
 
-    _currentUser = userResponse.user;
-    _currentSession = currentSession?.copyWith(user: userResponse.user);
+    _currentSession = currentSession!.copyWith(user: userResponse.user);
```
juampiq6, line 748: if access token is null, then the code will never reach the bang operator, hence it should be safe to use it here

Vinzent03 (commented):
The way the library currently works, those removal are okay.
But the comment about a non-null `User`, but `null` `Session` on `currentUser` will never happen, because it's derived from `_currentSession`. So we could either remove that comment or don't do the changes of this pr and instead set the `_currentUser` only in `signUp` if a `User` exists, but no `Session`.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -140,10 +137,10 @@ class GoTrueClient {
   /// Returns the current logged in user, if any;
   ///
   /// Use [currentSession] to determine whether the user has an active session,
-  /// because [currentUser] can be non-null without an active session, such as
```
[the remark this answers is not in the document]
juampiq6: my bad, its actually the opposite

juampiq6:
I totally agree, the comment is wrong, the one I modified aswell. I would just leave: 
`/// Returns the current logged in user, asociated to [currentSession] if any;`
About setting the _currentUser:
For that case where the user is not yet confirmed after SignUp, the `AuthResponse` would handle the `User` data needed.
The API of this library, in my opinion, should only expose a session, to not overcomplicate the understanding of this edge case.

packages/gotrue/lib/src/gotrue_client.dart
Vinzent03: Yeah the request itself doesn't change the `currentSession`, but while the request is going a signout or something else could happen that sets the `currentSession` to null

juampiq6:
@Vinzent03 made the required changes, let me know if you find something else.

Vinzent03 (approved):

Issue #1176, opened 2025-05-27
null check operator used on a null value in onResumed()


martinpelli:
**Describe the bug**
an inner function form `supabase.dart` is throwing `Null check operator used on a null value` 

**To Reproduce**
I don't have reproducible steps but I have the crashlytics logs. These happen in both, iOS and Android devices.

**Expected behavior**

**Logs**
```
          Fatal Exception: FlutterError
0  ???                            0x0 Supabase.onResumed + 242 (supabase.dart:242)
1  ???                            0x0 Supabase.didChangeAppLifecycleState + 224 (supabase.dart:224)
2  ???                            0x0 WidgetsBinding.handleAppLifecycleStateChanged + 1030 (binding.dart:1030)
3  ???                            0x0 ServicesBinding._handleLifecycleMessage + 306 (binding.dart:306)
4  ???                            0x0 BasicMessageChannel.setMessageHandler.<fn> + 261 (platform_channel.dart:261)
5  ???                            0x0 _DefaultBinaryMessenger.setMessageHandler.<fn> + 650 (binding.dart:650)
        
```

**Version (please complete the following information):**
On Linux/macOS
```
supabase_flutter 2.9.0
│   ├── supabase 2.7.0
│   │   ├── functions_client 2.4.2
│   │   ├── gotrue 2.12.0
│   │   ├── postgrest 2.4.2
│   │   ├── realtime_client 2.5.0
│   │   ├── storage_client 2.4.0
```

Vinzent03:
@martinpelli Yeah sure if you want to, that would be great!

PR #1178, opened 2025-05-27
fix(supabase_flutter): Safely check if conn is not null to avoid null check operator

martinpelli:

Bug fix

After app changes its life cycle, `onResumed()` method is called to resume any open web socket connection. This code is executed 
```
 if (realtime.channels.isNotEmpty) {
      if (realtime.connState == SocketStates.disconnecting) {
        bool cancel = false;
        final connectFuture = realtime.conn!.sink.done.then(
```
So, if the `SocketStates` is `disconnecting` it will try to use `realtime.conn!` assuming is not `null` when it can be `null`, hence throwing an error **Null check operator used on a null value**

If we look inside `RealtimeClient` which holds `conn` and `connState` we can see the only time `conn` is set to `null` and the only time `connState` is set to `SocketStates.disconnection` is here, in the `disconnect` method

```
Future<void> disconnect({int? code, String? reason}) async {
    final conn = this.conn;
    if (conn != null) {
      final oldState = connState;
      connState = SocketStates.disconnecting;
      
      
      if (oldState == SocketStates.open ||
          oldState == SocketStates.connecting) {
        if (code != null) {
          await conn.sink.close(code, reason ?? '');
        } else {
          await conn.sink.close();
        }
        connState = SocketStates.disconnected;
        reconnectTimer.reset();
        
      }
      this.conn = null;

      if (heartbeatTimer != null) heartbeatTimer?.cancel();
    }
  }
```
so it is clear that `conn` can be `null` and `connState` is `disconnecting` when `oldState` is `closed`, `disconnecting` or `disconnected`

Fixes [#1176 ](https://github.com/supabase/supabase-flutter/issues/1176)

## What is the new behavior?

The new behavior just checks if `conn` is **not** `null`.

### DIFF
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -222,9 +222,14 @@ class RealtimeClient {
     final conn = this.conn;
     if (conn != null) {
       final oldState = connState;
-      connState = SocketStates.disconnecting;
-      log('transport', 'disconnecting', {'code': code, 'reason': reason},
-          Level.FINE);
+      final shouldCloseSink =
+          oldState == SocketStates.open || oldState == SocketStates.connecting;
+      if (shouldCloseSink) {
+        // Don't set the state to `disconnecting` if the connection is already closed.
+        connState = SocketStates.disconnecting;
+        log('transport', 'disconnecting', {'code': code, 'reason': reason},
+            Level.FINE);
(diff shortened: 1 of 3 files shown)


martinpelli:
@Vinzent03  @dshukertjr  can someone check this? This issue is happening so frequently

grdsdev (approved):

PR #1188, opened 2025-06-06
feat(gotrue): Add phone mfa enrollment

dshukertjr:

feature of adding phone support for MFA enrollment

JS PR here: https://github.com/supabase/auth-js/pull/932

Enrolling in MFA via SMS is supported .

## What is the new behavior?

SMS MFA is not supported.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_mfa_api.dart
+++ b/packages/gotrue/lib/src/gotrue_mfa_api.dart
@@ -28,39 +28,57 @@ class GoTrueMFAApi {
 
   /// Starts the enrollment process for a new Multi-Factor Authentication (MFA) factor.
   /// This method creates a new `unverified` factor.
-  /// To verify a factor, present the QR code or secret to the user and ask them to add it to their authenticator app.
+  /// 
+  /// For TOTP: To verify a factor, present the QR code or secret to the user and ask them to add it to their authenticator app.
+  /// For Phone: The user will receive an SMS with a verification code.
   ///
-  /// The user has to enter the code from their authenticator app to verify it.
+  /// The user has to enter the code from their authenticator app or SMS to verify it.
   ///
   /// Upon verifying a factor, all other sessions are logged out and the current session's authenticator level is promoted to `aal2`.
   ///
   /// [factorType] : Type of factor being enrolled.
   ///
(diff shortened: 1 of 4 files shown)


packages/gotrue/lib/src/types/mfa.dart
```diff
@@ -4,23 +4,33 @@ class AuthMFAEnrollResponse {
...
   final String id;
 
-  /// Type of MFA factor. Only `[FactorType.totp] supported for now.
+  /// Type of MFA factor. Supports both `[FactorType.totp]` and `[FactorType.phone]`.
   final FactorType type;
 
-  /// TOTP enrollment information.
-  final TOTPEnrollment totp;
+  /// TOTP enrollment information (only present when type is totp).
+  final TOTPEnrollment? totp;
```
dshukertjr, line 11: I know this isn't great, but I think we can call it that it's a fix. Open to suggestions to avoid this though.

packages/gotrue/lib/src/gotrue_mfa_api.dart
```diff
@@ -28,39 +28,57 @@ class GoTrueMFAApi {
...
   ///
   /// [friendlyName] : Human readable name assigned to the factor.
+  ///
+  /// [phone] : Phone number to enroll for Phone factor type.
   Future<AuthMFAEnrollResponse> enroll({
     FactorType factorType = FactorType.totp,
     String? issuer,
     String? friendlyName,
+    String? phone,
   }) async {
```
grdsdev, line 51: I wonder if it isn't better to deprecate this method and add 2 new ones such as:

```dart
Future<AuthMFAEnrollTOTPResponse> enrollWithTOTP({
    String issuer,
    Stirng? friendlyName,
}) async

Future<AuthMFAEnrollPhoneResponse> enrollWithPhone({
    String phone,
    Stirng? friendlyName,
}) async
```

Notice the method returns different response types, which will have the totp and phone properties non-null respectively.

@dshukertjr what do you think?

packages/gotrue/lib/src/types/mfa.dart
grdsdev: See my comment in https://github.com/supabase/supabase-flutter/pull/1188/files#r2131807418

PR #1189, opened 2025-06-10
chore(release): Publish packages - supabase_flutter@2.9.1

dshukertjr:
- functions_client@2.4.3
 - gotrue@2.13.0
 - realtime_client@2.5.1
 - supabase@2.8.0
 - supabase_flutter@2.9.1

### DIFF
--- a/packages/supabase_flutter/CHANGELOG.md
+++ b/packages/supabase_flutter/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.9.1
 
  - Update a dependency to the latest release.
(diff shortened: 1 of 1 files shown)


Vinzent03:
I don't like the breaking change of #1188 as code using the totp mfa won't compile. I would favor a new method as proposed in that pr and maybe combine them in the next major version. Doing that breaking change in the js package is a different thing, as it is not as sound as Dart. If you still want to release this as is, at least mark in clearly in the changelog.

PR #1204, opened 2025-07-21
refactor: Simplify httpClient conditional logic using ternary operator

felixkamau:

Refactor 

The _handleRequest method uses an if-else statement to conditionally assign the streamedResponse variable based on whether httpClient is null or not. The logic spans multiple lines and uses a commented-out block.

## What is the new behavior?

The same conditional logic is now implemented using a ternary operator, making the code more concise and readable


Vinzent03:
I prefer the current style and it's easier to extend.

PR #1205, opened 2025-07-25
fix(supabase_flutter): use getInitialAppLink from app_links

MikaelDDavidd:
The previous implementation was trying to support multiple versions of the app_links package, but it was causing a NoSuchMethodError when using newer versions.

This commit removes the backward compatibility logic and uses the getInitialAppLink method directly, which is available in app_links ^4.0.0 and later.

## What is the new behavior?

## Additional context


Vinzent03:
I'm a fan of removing support for the old versions, but this pr sets the version to only ^4.0.0 Although there are newer versions which we want to support as well.

PR #1225, opened 2025-09-18
Add User Banning Functionality with bannedUntil Field

ahmtydn:
## Summary
Add user banning functionality with `bannedUntil` field and `isBanned` computed property to the User class.

### Core Implementation
- **Added `bannedUntil` field**: Nullable String property to store ban expiration timestamp
- **Added `isBanned` computed property**: UTC-based validation to check if user is currently banned
- **JSON serialization**: Full support for parsing `banned_until` from API responses
- **Backward compatibility**: All changes are non-breaking, existing code unaffected

### Technical Details
```dart
class User {
  final String? bannedUntil;  // New field
  
  bool get isBanned {         // New computed property
    if (bannedUntil == null) return false;
    try {
      final banExpiration = DateTime.parse(bannedUntil!);
      final now = DateTime.now().toUtc();
      return now.isBefore(banExpiration);
    } catch (e) {
      return false;  // Safe fallback for invalid dates
    }
  }
}
```

## Usage Example
```dart
final user = Supabase.instance.client.auth.currentUser;

if (user?.isBanned == true) {
  print('User is banned until: ${user!.bannedUntil}');
  // Handle banned user logic
} else {
  // Normal user flow
}
```

## Backend Requirements 🔄
This frontend implementation is ready but **requires backend support**:

1. **Database Schema**: Add `banned_until timestamptz` column to `auth.users` table
2. **API Response**: Include `banned_until` field in user object responses  
3. **Server Validation**: Implement ban checking on server endpoints

### Database Migration Needed:
```sql
ALTER TABLE auth.users ADD COLUMN banned_until timestamptz;
```

### Expected API Response Format:
```json
{
  "id": "user-id",
  "email": "user@example.com",
  "banned_until": "2025-12-31T23:59:59.000Z",
  // ... other user fields
}
```

## Current Status
- ✅ Frontend parsing ready
- ✅ UTC timezone handling implemented  
- ✅ Backward compatible
- 🔄 Waiting for backend `banned_until` field support

Once backend adds the `banned_until` field to API responses, this feature will work automatically.

## Testing
The implementation includes robust error handling:
- Safe parsing of ISO 8601 date strings
- UTC timezone normalization for global consistency
- Graceful fallback for invalid date formats
- Null safety throughout

## Compatibility
- **Dart SDK**: Compatible with existing version requirements
- **Breaking Changes**: None - all changes are additive
- **Migration**: No client-side migration needed

## Next Steps
Seeking backend team collaboration to:
- Add database migration for `banned_until` column
- Update API responses to include the field
- Implement server-side ban validation
- Add admin endpoints for user ban management (optional)

The client-side implementation is production-ready! 🚀

## Files Changed
- `packages/gotrue/lib/src/types/user.dart` - Added bannedUntil field and isBanned method

### DIFF
--- a/packages/gotrue/lib/src/types/user.dart
+++ b/packages/gotrue/lib/src/types/user.dart
@@ -27,6 +27,7 @@ class User {
   final List<UserIdentity>? identities;
   final List<Factor>? factors;
   final bool isAnonymous;
+  final String? bannedUntil;
 
   const User({
     required this.id,
@@ -51,7 +52,23 @@ class User {
     this.identities,
     this.factors,
     this.isAnonymous = false,
+    this.bannedUntil,
   });
+
+  /// Returns true if the user is currently banned
+  /// Compares the current UTC time with the bannedUntil timestamp
+  bool get isBanned {
+    if (bannedUntil == null) return false;
+
+    try {
+      final banExpiration = DateTime.parse(bannedUntil!);
+      final now = DateTime.now().toUtc();
+      return now.isBefore(banExpiration);
+    } catch (e) {
+      // If bannedUntil cannot be parsed, consider user not banned
+      return false;
+    }
+  }
 
   /// Returns a `User` object from a map of json
   /// returns `null` if there is no `id` present
@@ -88,6 +105,7 @@ class User {
           ? List<Factor>.from(json['factors']?.map((x) => Factor.fromJson(x)))
           : null,
       isAnonymous: json['is_anonymous'] ?? false,
+      bannedUntil: json['banned_until'],
     );
   }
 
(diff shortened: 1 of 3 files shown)


packages/gotrue/lib/src/types/user.dart
```diff
@@ -51,8 +52,24 @@ class User {
...
     this.isAnonymous = false,
+    this.bannedUntil,
   });
 
+  /// Returns true if the user is currently banned
+  /// Compares the current UTC time with the bannedUntil timestamp
+  bool get isBanned {
+    if (bannedUntil == null) return false;
+
+    try {
```
Vinzent03, line 63: Might use `tryParse()` instead and check for null to not catch every exception.

Vinzent03 (changes requested):

Vinzent03:
`banned_until` already exists as a column in `auth.users`, but seems not to be returned by the api. There is already a banned functionality via the admin method `updateUserById()` and `banDuration`, but I currently don't see a way to know whether a user is banned or not. But this has to be first done by the auth team on the server. I don't think already implementing this here is helpful.

Vinzent03:
I would close this for now, because the required values are not provided by the backend.

Issue #1227, opened 2025-09-22
upsert fails to pass onConflict into the on_conflict in postgREST


sbutlerjr:
**Describe the bug**
When doing an UPSERT and supplying the onConflict vie the SDK, the on_conflict property and value isn't passed through to the REST API endpoint.  That causes the request to fail with a 409 status code because the unique constraint is violated.

**To Reproduce**
```dart
Supabase.instance.client
                .from('party')
                .upsert({company_uuid: nMQXF9BL-D, uuid: PAR_642d9e4f-669e-402d-8ccd-75297f8c63a2, is_active: true, enable_sms: false, roles: [3], first_name: TEST, last_name: TEST, email: test.test@test.com, phone: 123-876-5309, agency_uuid: 08efd21a-d25a-4c7f-a7d9-dced19e477f8, uniq: TEST:TEST}, onConflict: 'uniq');
```

This generates a POST request to: 
```https://nqnopvehcfcqebspvkba.supabase.co/rest/v1/party?columns="company_uuid","uuid","is_active","enable_sms","roles","first_name","last_name","email","phone","agency_uuid","uniq"```

With the correct header: Prefer "resolution=merge-duplicates"
and the correct body:
```
[{"company_uuid":"nMQXF9BL-D","uuid":"PAR_642d9e4f-669e-402d-8ccd-75297f8c63a2","is_active":true,"enable_sms":false,"roles":[3],"first_name":"TEST","last_name":"TEST","email":"test.test@test.com","phone":"123-867-5309","agency_uuid":"08efd21a-d25a-4c7f-a7d9-dced19e477f8","uniq":"TEST:TEST"}]
```
But its missing the value I specified with **onConflict: 'uniq'**

**Expected behavior**
According to [PostgREST docs](https://docs.postgrest.org/en/v13/references/api/tables_views.html#on-conflict)  it should have added: on_conflict=uniq to the URL.  
So the POST REST URL should have been 
```https://nqnopvehcfcqebspvkba.supabase.co/rest/v1/party?columns="company_uuid","uuid","is_active","enable_sms","roles","first_name","last_name","email","phone","agency_uuid","uniq"&on_conflict=uniq```

**Version (please complete the following information):**
supabase_fluter: 2.10.1
Project is  Postgres version:17.6.1.004,  PostgREST version 13.0.5

sbutlerjr:
[...]

PR #1228, opened 2025-09-22
fix(postgrest): URL replacement logic in query builder

sbutlerjr:
Resolves: https://github.com/supabase/supabase-flutter/issues/1227

Prior bug omitted passing onConflict through to on_conflict if the value data was a list.

on_conflict was omitted from the REST URL if the data was a list

## What is the new behavior?

Works correctly with POSTGrest on_conflict docs and passes the columns and on_conflict properties in the URL.


grdsdev:
Hi @sbutlerjr thanks for opening this fix, can you also include a unit test for asserting this behavior?

sbutlerjr:
Hi @grdsdev Tests added

grdsdev (approved):

PR #1242, opened 2025-09-30
feat: add support for broadcast replay configuration

grdsdev:
## Summary
- Added `ReplayOption` class for configuring broadcast replay
- Added `replay` field to `RealtimeChannelConfig`
- Updated `toMap()` method to include replay configuration in broadcast config

## Changes
This PR implements broadcast replay support following the pattern from realtime-js PR #540:

1. **ReplayOption**: New class with `since` (int) and optional `limit` (int?)
2. **RealtimeChannelConfig**: Added optional `replay` field
3. **toMap()**: Updated to serialize replay configuration

## Usage
```dart
// Configure broadcast with replay
final channel = supabase.channel(
  'my-channel',
  RealtimeChannelConfig(
    self: true,
    ack: true,
    replay: ReplayOption(
      since: 1234567890,
      limit: 100,
    ),
  ),
);

// Broadcast callback receives meta field
channel.onBroadcast(
  event: 'position',
  callback: (payload) {
    final meta = payload['meta'] as Map<String, dynamic>?;
    if (meta?['replayed'] == true) {
      print('Replayed message: ${meta?['id']}');
    }
  },
).subscribe();
```

## Test plan
- [ ] Verify type checking passes
- [ ] Test broadcast replay configuration
- [ ] Verify meta field is properly included in callbacks

### DIFF
--- a/packages/realtime_client/lib/src/types.dart
+++ b/packages/realtime_client/lib/src/types.dart
@@ -138,12 +138,38 @@ extension ToType on RealtimeListenTypes {
   }
 }
 
+/// Configuration for broadcast replay feature.
+/// Allows replaying broadcast messages from a specific timestamp.
+class ReplayOption {
+  /// Unix timestamp (in milliseconds) from which to start replaying messages
+  final int since;
+
+  /// Optional limit on the number of messages to replay
+  final int? limit;
+
+  const ReplayOption({
+    required this.since,
+    this.limit,
+  });
+
+  Map<String, dynamic> toMap() {
+    final map = <String, dynamic>{'since': since};
+    if (limit != null) {
+      map['limit'] = limit;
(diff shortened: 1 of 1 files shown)


Vinzent03:
If I'm correct, your usage example is not entirely correct as the js documentation states that it only works for private channels. I think this should be added to the documentation of the `ReplayOption` as well. The limit of 25 and it only works for messages sent from the database should be added as well. Basically this block from the js readme:

> Broadcast Replay enables **private** channels to access messages that were sent earlier. Only messages published via [Broadcast From the Database](https://supabase.com/docs/guides/realtime/broadcast#trigger-broadcast-messages-from-your-database) are available for replay.

Vinzent03 (approved):

PR #1246, opened 2025-10-06
feat(gotrue): introduce getClaims method to verify and extract JWT claims

grdsdev:
## Summary

This PR ports the `getClaims` method from [supabase/auth-js#1030](https://github.com/supabase/auth-js/pull/1030) to the Flutter client. The method supports verifying JWTs (both symmetric and asymmetric) and returns the entire set of claims in the JWT payload.

## Changes

- ✨ **New `getClaims()` method** in `GoTrueClient` for JWT verification and claims extraction
- 📦 **Base64URL utilities** (RFC 4648 compliant) for encoding/decoding JWT parts
- 🔒 **JWT types**: `JwtHeader`, `JwtPayload`, `DecodedJwt`, `GetClaimsResponse`
- 🛠️ **Helper functions**: `decodeJwt()` and `validateExp()`
- 🚨 **New exception**: `AuthInvalidJwtException` for JWT-related errors
- ✅ **Comprehensive tests** for getClaims, JWT helpers, and base64url utilities

## Implementation Details

The method verifies JWTs by calling `getUser()` to validate against the server, supporting both:
- **HS256** (symmetric) algorithms
- **RS256/ES256** (asymmetric) algorithms

This approach differs slightly from the JavaScript implementation, which uses WebCrypto API for asymmetric JWT verification. In the Flutter implementation, we rely on server-side verification through `getUser()` for all JWT types, which provides:
- ✅ Security through server-side validation
- ✅ Consistency across platforms
- ✅ No additional cryptographic dependencies

## Usage Example

```dart
// Get claims from current session
final response = await client.getClaims();
print(response.claims['sub']); // User ID
print(response.claims['email']); // User email
print(response.claims['role']); // User role

// Get claims from explicit JWT
final response = await client.getClaims(accessToken);
```

## Testing

All tests pass:
- ✅ `getClaims()` with valid JWT from current session
- ✅ `getClaims()` with explicit JWT parameter
- ✅ Error handling for missing session, invalid JWT, and expired tokens
- ✅ Server verification through `getUser()`
- ✅ Base64URL encoding/decoding round-trips
- ✅ JWT compatibility tests

## Notes

- This is marked as an **experimental API** and may change in future versions
- The method validates JWT expiration before making the server call
- Claims include all standard JWT fields (sub, aud, exp, iat, iss) plus custom fields

## Related

- Ported from: https://github.com/supabase/auth-js/pull/1030

### DIFF
--- a/packages/gotrue/lib/gotrue.dart
+++ b/packages/gotrue/lib/gotrue.dart
@@ -8,6 +8,7 @@ export 'src/types/auth_exception.dart';
 export 'src/types/auth_response.dart' hide ToSnakeCase;
 export 'src/types/auth_state.dart';
 export 'src/types/gotrue_async_storage.dart';
+export 'src/types/jwt.dart';
 export 'src/types/mfa.dart';
 export 'src/types/types.dart';
 export 'src/types/session.dart';
--- a/packages/gotrue/lib/src/base64url.dart
+++ b/packages/gotrue/lib/src/base64url.dart
@@ -1,0 +1,122 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+/// Base64URL encoding and decoding utilities for JWT operations.
+/// Extracted and adapted from RFC 4648 specification.
+class Base64Url {
+  static const String _chars =
+      'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_';
+  static const int _bits = 6;
+
+  /// Decodes a base64url encoded string to bytes
+  ///
+  /// [input] The base64url encoded string to decode
+  /// [loose] If true, allows lenient parsing that doesn't strictly validate padding
+  static Uint8List decode(String input, {bool loose = false}) {
+    // Remove padding characters
+    String string = input.replaceAll('=', '');
+
+    // Build character lookup table
+    final Map<String, int> codes = {};
+    for (int i = 0; i < _chars.length; i++) {
+      codes[_chars[i]] = i;
+    }
+
+    // For loose mode or when there's actual content, skip strict validation
+    // The validation below will catch actual errors during decoding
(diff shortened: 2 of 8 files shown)


Vinzent03:
Okay seems like pointycastle v4.0 needs Dart 3.4.0, but we only support Dart 3.3.0. Great...
I think it should be possible to allow a range of 3< v <4 for pointycastle. 
But now that I saw the version bump for dart_jsonwebtoken, can't we simply use that for the verifying? https://pub.dev/packages/dart_jsonwebtoken#check-if-the-jwt-made-is-correct
In the latest version, dart_jsonwebtoken uses pointycastle under the hood either way.

grdsdev:
Yes @Vinzent03 I think we can use `dart_jsonwebtoken` directly, let me try that.

packages/gotrue/pubspec.yaml
```diff
@@ -1,26 +1,26 @@
...
+  http: ">=0.13.0 <2.0.0"
   jwt_decode: ^0.3.1
   retry: ^3.1.0
-  rxdart: '>=0.27.7 <0.29.0'
+  rxdart: ">=0.27.7 <0.29.0"
   meta: ^1.7.0
   logging: ^1.2.0
-  web: '>=0.5.0 <2.0.0'
+  web: ">=0.5.0 <2.0.0"
+  dart_jsonwebtoken: ^2.17.0
```
Vinzent03:
```suggestion
  dart_jsonwebtoken: ">=2.17.0 <4.0.0"
```
I know these versions conflicts are really tedious, be I think we should support if possible the latest version. This constraint should use 3.0.0 for people with at least Dart 3.4.0 and fallback to 2.17.0 for people below that.

grdsdev: I don't think that will be possible, as implementation differs significantly, but let me try.

PR #1253, opened 2025-10-09
feat(realtime): add explicit REST API call method for broadcast messages

grdsdev:
## Summary

This PR adds a new `httpSend` method to the `RealtimeChannel` class that explicitly uses the REST API endpoint for sending broadcast messages, regardless of WebSocket connection state.

## Changes

- ✨ Added `httpSend` method with proper error handling and timeout support
- ⚠️ Added deprecation warning to `send` method when falling back to REST
- ✅ Comprehensive test coverage for the new method

## Motivation

Currently, the `send` method automatically falls back to using the REST API when the WebSocket connection is not available. However, this behavior is implicit and users may not be aware they are using REST instead of WebSocket.

This change provides an explicit method for REST delivery, giving users more control over their message delivery mechanism and preparing for future deprecation of the implicit fallback behavior.

## Test Plan

- [x] All existing tests pass
- [x] New tests for `httpSend` method covering:
  - Successful 202 responses
  - Authorization header handling (with and without access token)
  - Private channel support
  - Error handling for non-202 status codes
  - Timeout handling
  - Correct payload formatting

## Related

Ported from: https://github.com/supabase/supabase-js/pull/1751

### DIFF
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -550,37 +550,30 @@ class RealtimeChannel {
       ]
     };
 
-    try {
-      final res = await (socket.httpClient?.post ?? post)(
-        Uri.parse(broadcastEndpointURL),
-        headers: headers,
-        body: json.encode(body),
-      ).timeout(
-        timeout ?? _timeout,
-        onTimeout: () => throw TimeoutException('Request timeout'),
-      );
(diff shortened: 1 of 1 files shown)


Vinzent03:
@grdsdev What's the status here?

grdsdev:
> @grdsdev What's the status here?

@Vinzent03 I think I forgot this PR, applied your suggestions, can you review it again?

PR #1279, opened 2025-11-26
fix: mark lastAccessedAt field as deprecated

Vinzent03:
docs/fix

## What is the new behavior?

The storage `lastAccessAt` field is deprecated. I couldn't find a real reason or alternative, so the message is currently empty. I raised the issue initially in supabase-js that the value is missing despite its type https://github.com/supabase/storage/issues/541

## Additional context

js pr: https://github.com/supabase/supabase-js/pull/1743


Vinzent03:
@grdsdev  The deprecation still needs to be marked in this sdk, so I reopened this pr.
(The tests are failing due to #1366)

grdsdev (approved):

PR #1282, opened 2025-12-10
chore: bump app_links to ` '>=6.2.0 <8.0.0'`

spydon:

This updates `app_links` to ^7.0.0 so that users with this as a direct dependency aren't stuck on an old version.

Users can't use the newest `app_links` version.

## What is the new behavior?

Users can use the newest `app_links` version.


Vinzent03:
This is a breaking change and requires the latest flutter version which is not desired. Since the v7 is compatible with v6 we should just allow both versions to be used instead of forcing v7.

spydon:
> This is a breaking change and requires the latest flutter version which is not desired. Since the v7 is compatible with v6 we should just allow both versions to be used instead of forcing v7.

Fair enough, updated it to include both versions

Vinzent03:
Thanks for the change!

Issue #1286, opened 2025-12-12
Bug: `GoTrueClient.getClaims()` crashes on first use for RS*/ES* JWTs (null `_jwks`)


khensunny:
@grdsdev 

> [!NOTE]
> I used AI for the write up but this is a legit bug 

### Summary
`GoTrueClient.getClaims()` can throw `DartError: Unexpected null value.` when verifying JWTs signed with **asymmetric algorithms** (e.g. `RS256`, `ES256`) because it force-unwraps `_jwks` before `_jwks` is ever initialized.

This is a broken assumption in the JWKS verification path.

---

### Affected code
In `gotrue` (example from gotrue_client.dart), `getClaims()` contains:

```dart
final signingKey =
    (decoded.header.alg.startsWith('HS') || decoded.header.kid == null)
        ? null
        : await _fetchJwk(decoded.header.kid!, _jwks!);
```

`_jwks` is declared as nullable and is initialized lazily:

```dart
JWKSet? _jwks;
DateTime? _jwksCachedAt;
```

`_jwks` is only set inside `_fetchJwk()` **after** fetching `/.well-known/jwks.json`:

```dart
final jwksResponse = await _fetch.request('$_url/.well-known/jwks.json', ...);
final jwks = JWKSet.fromJson(jwksResponse as Map<String, dynamic>);
_jwks = jwks;
_jwksCachedAt = now;
```

So on the **first** call to `getClaims()` for a token requiring JWKS verification, `_jwks` is still `null`, and `_jwks!` crashes immediately.

---

### Expected behavior
- `getClaims()` should not crash due to internal cache state.
- On first use, if `_jwks` is null, it should fetch `/.well-known/jwks.json`, cache it, and verify the JWT (or fall back to server verification if applicable).

---

### Actual behavior
When using an asymmetric JWT with a `kid`:
- `getClaims()` throws a runtime null error:
  - `DartError: Unexpected null value.`

This happens before any network call to retrieve JWKS occurs.

---

### Why Supabase Flutter CI tests may still pass
The CI/local tests can pass without hitting this bug if the test GoTrue instance signs tokens with **HS256** (symmetric signing via `GOTRUE_JWT_SECRET`). In that case:

- `decoded.header.alg.startsWith('HS')` is `true`
- `signingKey` becomes `null`
- `getClaims()` falls back to `await getUser(token)` (server verification path)
- JWKS code path (and `_jwks!`) is never executed

[...]

PR #1288, opened 2025-12-15
fix(gotrue): fix getClaims JWT token decoding

khensunny:

Bug fix (fixes `GoTrueClient.getClaims()` JWT decoding for asymmetric algorithms), plus tests, CI/workflow updates, package/pubspec adjustments, and infra (JWKS docker-compose) additions.

`getClaims()` can throw `DartError: Unexpected null value.` when verifying JWTs signed with asymmetric algorithms (e.g., `RS256`, `ES256`) because `_jwks` is force-unwrapped before it is initialized — see supabase/supabase-flutter#1286. CI/tests may not surface this because the local test auth uses HS* signing.

## What is the new behavior?

`getClaims()` no longer force-unwraps `_jwks`; the JWKS path is handled safely by fetching and caching `/.well-known/jwks.json` on first use (or falling back appropriately), so asymmetric (RS*/ES*) JWTs with `kid` are verified without crashing. Tests updated to cover the JWKS path; CI/workflows, gotrue pubspec, and docker-compose.jwk.yml were updated to support and validate the fix.

## Additional context

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -3,7 +3,6 @@ import 'dart:convert';
 import 'dart:math';
 
 import 'package:collection/collection.dart';
-import 'package:dart_jsonwebtoken/dart_jsonwebtoken.dart';
 import 'package:gotrue/gotrue.dart';
 import 'package:gotrue/src/constants.dart';
 import 'package:gotrue/src/fetch.dart';
@@ -11,6 +10,7 @@ import 'package:gotrue/src/helper.dart';
 import 'package:gotrue/src/types/auth_response.dart';
 import 'package:gotrue/src/types/fetch_options.dart';
 import 'package:http/http.dart';
+import 'package:jose_plus/jose.dart' as jose;
 import 'package:jwt_decode/jwt_decode.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
@@ -1417,7 +1417,10 @@ class GoTrueClient {
     final signingKey =
         (decoded.header.alg.startsWith('HS') || decoded.header.kid == null)
             ? null
-            : await _fetchJwk(decoded.header.kid!, _jwks!);
+            : await _fetchJwk(
+                decoded.header.kid!,
+                _jwks ?? JWKSet(keys: const <JWK>[]),
+              );
 
     // If symmetric algorithm, fallback to getUser()
     if (signingKey == null) {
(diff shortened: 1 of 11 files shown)


grdsdev:
@khensunny thanks for opening this PR, I'll review and provide any feedback if needed.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -1429,11 +1432,20 @@ class GoTrueClient {
     }
 
     try {
-      JWT.verify(token, signingKey.rsaPublicKey);
+      final keyStore = jose.JsonWebKeyStore();
```
Vinzent03: I would prefer we keep using the previous library and add the ES* alg case as well. That was the reason to change this here, right?

packages/gotrue/test/get_claims_test.dart
```diff
@@ -165,6 +165,38 @@ void main() {
       expect(claimsResponse.claims.claims['email'], newEmail);
     });
 
+    test('getClaims() verifies asymmetric JWT via JWKS (ES*/RS*)', () async {
```
Vinzent03: Can't the test here be the same as before and testing RS/ES only depends on the sign key on the server? So this test can be removed here?

packages/gotrue/lib/src/gotrue_client.dart
khensunny: No, with json_webtoken the solution would be more involved. Though it does understand the algorithm it doesn't understand the JWK format(the response from the public key well known endpoint), that is,
this part:

https://github.com/supabase/supabase-js/blob/5c06206cc10c759382d716a6624dd6301bb3ed44/packages/core/auth-js/src/GoTrueClient.ts#L3819

in supabase-js.

To do it manually we'd need this:

```dart
RSAPublicKey _jwkToRSAPublicKey(Map<String, dynamic> jwk) {
  final n = _base64UrlToBigInt(jwk['n']);
  final e = _base64UrlToBigInt(jwk['e']);
  
  return RSAPublicKey(n, e);
}

BigInt _base64UrlToBigInt(String value) {
  final bytes = base64Url.decode(base64Url.normalize(value));
  return BigInt.parse(
    bytes.map((b) => b.toRadixString(16).padLeft(2, '0')).join(),
    radix: 16,
  );
}
```


I strongly suggest moving the package to jose/jose-plus because they have the features built-in and will reduce the maintenance burden especially with Supabase supporting asymmetric keys. Jose has caching builtin(I didn't use because we already have our own implementation)

PR #1315, opened 2026-02-23
getUser is also refresh the user reference in the currentSession

stefanschaller:

Bug fix/feature -> Depending on who you ask 😅

The `getUser` call only gets the user without doing anything with the user 

## What is the new behavior?

The old user object is also updated in the `currentSession` 

## Additional context

We update the user object in an edgefunction and want to refresh the user afterwards.


Vinzent03:
The auth-js package works the same way as it is implemented in Dart currently. So I would keep it the way it is.
If you want a session object with the updated user object, you can just use `refreshSession()`

PR #1323, opened 2026-03-22
fix(storage): use toString() for statusCode to handle non-string types

### LINKED ISSUE #1312: Mismatch status code in StorageException object and in exception message body.
**Describe the bug**
We're getting 2 different status code for an exception hence causing conflict which one to use.

**To Reproduce**
1. Upload an entity in Storage bucket
2. Make in accessible, could be wrong path name or file name
3. Make the download API call

**Expected behavior**
The status code should be 404 both in `StorageException` object and in nested message param.

**Screenshots**

Image

On Linux/macOS
Please run `dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"` in your project directory and paste the output here.
```
hamza@M1-Max lensfolio_mobile_app % dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"
├── supabase_flutter 2.12.0
│   ├── supabase 2.10.2
│   │   ├── functions_client 2.5.0
│   │   ├── gotrue 2.18.0
│   │   ├── postgrest 2.6.0
│   │   ├── realtime_client 2.7.0
│   │   ├── storage_client 2.4.1
```

**Additional context**
The object for me does exists on storage, the policy for storage is to read by any authenticated user yet i'm facing this issue.
- Bucket name: 'resumes'
- File path: 'resumes/uuid.png'

Code snippet:
```dart
  static Future<File> downloadResume(String filePath) async {
    try {
      final file = await AppSupabase.supabase.storage
          .from(SupaBuckets.resumes)
          .download(filePath);
      return File.fromRawPath(file);
    } catch (e, st) {
      if (e is StorageException) {
        throw SupaStorageFault.fromStorageException(e, st);
      }
      throw UnknownFault('Something went wrong!', st);
    }
  }
```

mitchkoko:
Fixes #1312

Bug fix

`StorageException.fromJson` uses `as String?` to cast `json['statusCode']`, which silently returns `null` when the server sends `statusCode` as an integer (e.g., `404` instead of `"404"`). This causes the `statusCode` field to be `null` instead of the actual value from the JSON response body.

## What is the new behavior?

Uses `.toString()` instead of `as String?`, which correctly converts the `statusCode` regardless of whether the server sends it as a string or integer.

### DIFF
--- a/packages/storage_client/lib/src/types.dart
+++ b/packages/storage_client/lib/src/types.dart
@@ -309,7 +309,7 @@ class StorageException implements Exception {
       StorageException(
         json['message'] as String? ?? json.toString(),
         error: json['error'] as String?,
-        statusCode: (json['statusCode'] as String?) ?? statusCode,
+        statusCode: json['statusCode']?.toString() ?? statusCode,
       );
 
   @override
--- a/packages/storage_client/test/client_test.dart
+++ b/packages/storage_client/test/client_test.dart
@@ -415,7 +415,7 @@ void main() {
         await storage.from('bucket2').download(uploadPath);
         fail('File that does not exist was found');
       } on StorageException catch (error) {
-        expect(error.statusCode, '400');
+        expect(error.statusCode, '404');
       }
       await storage
           .from(newBucketName)
@@ -435,7 +435,7 @@ void main() {
         await storage.from('bucket2').download('$uploadPath 3');
         fail('File that does not exist was found');
       } on StorageException catch (error) {
-        expect(error.statusCode, '400');
+        expect(error.statusCode, '404');
       }
       await storage
           .from(newBucketName)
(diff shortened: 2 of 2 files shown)


grdsdev:
Hey @mitchkoko, I understand the change and it makes sense, but tests are failing.

The issue arises because the `StorageException` is initialized through the following code:

https://github.com/supabase/supabase-flutter/blob/main/packages/storage_client/lib/src/fetch.dart#L38-L40

However, it doesn't use `fromJson`, which is where the change was made.

This happens because, even though the JSON response contains `statusCode = 404`, the HTTP response status code remains `400`. I know this isn’t ideal, but it’s how the storage service currently works.

Please revert the test changes, but keep the `toString()` since it does make sense.

Thanks.

grdsdev (approved):

PR #1324, opened 2026-03-22
fix: use private Realtime channel for .stream()

### LINKED ISSUE #1311: Using `stream` convenience function doesn't work with private-only realtime channels setting.
I have a table with proper RLS policies set up and I'm using the `stream` function ([stream docs](https://supabase.com/docs/reference/dart/stream)) to show a list and keep up with changes.

This was all working great. Later, when I began using realtime broadcasts for other features, I decided I didn't want to allow public realtime connections so I went to the realtime settings and changed the "Allow public access to channels" options to False. Changing this connection immediately broke my stream function with the following error:
```
RealtimeSubscribeException(status: RealtimeSubscribeStatus.channelError, details:
Exception: "PrivateOnly: This project only allows private channels")
```

According to the documentation [here](https://supabase.com/docs/guides/realtime/authorization?queryGroups=language&language=python#interaction-with-postgres-changes), the private setting shouldn't affect Postgres Changes (since those use the RLS policies of the table being watched, not realtime.messages). Perhaps I'm misunderstanding, but I thought the `stream` function used Postgres Changes under the hood since I'm not setting up any broadcast triggers or anything like that.

So if I'm not mistaken, I don't see why disabling public realtime channels would break the `stream` function.

My user is definitely signed in and has proper RLS access, since once I enable public access again, I can create, view, and delete records just fine.

Is this a bug, or am I misunderstanding how the `stream` function works?

jeremiahseun:

Bug fix. 

When public channels are disabled in a Supabase project ("Allow public access to channels" is false), attempting to use `.stream()` results in an Exception: "PrivateOnly: This project only allows private channels". This happens because `.stream()` creates a Realtime channel without specifying it as private. By explicitly configuring the `RealtimeChannelConfig` with `private: true`, `.stream()` bypasses this restriction and instead relies on Row Level Security (RLS) policies, effectively allowing users to track Postgres changes on tables with private realtime channel settings.

## What is the new behavior?

Fixes an issue where `supabase.from('table').stream(...)` throws a `RealtimeSubscribeException` ("PrivateOnly: This project only allows private channels") when the "Allow public access to channels" setting is disabled in the Supabase project.

This changes `_realtimeClient.channel(...)` inside `SupabaseStreamBuilder` to use `RealtimeChannelConfig(private: true)`, ensuring the stream function respects table RLS policies and successfully connects even if public channels are disabled.

## Additional context

This resolves #1311

### DIFF
--- a/packages/supabase/lib/src/supabase_stream_builder.dart
+++ b/packages/supabase/lib/src/supabase_stream_builder.dart
@@ -167,7 +167,10 @@ class SupabaseStreamBuilder extends Stream<SupabaseStreamEvent> {
       );
     }
 
-    _channel = _realtimeClient.channel(_realtimeTopic);
+    _channel = _realtimeClient.channel(
+      _realtimeTopic,
+      const RealtimeChannelConfig(private: true),
+    );
 
     _channel!
         .onPostgresChanges(


edgurgel:
I think the proper solution is to allow a call like the one below to receive a channel configuration for the stream so that people can decide what 

```dart
supabase.from('countries')
  .stream(primaryKey: ['id'])
  .listen((List<Map<String, dynamic>> data) {
  // Do something awesome with the data
});
```

grdsdev:
supersed by #1357

Issue #1278, opened 2025-11-26
☂️ v3 tracking issue


Vinzent03:
This is an umbrella issue for planned changes for v3. of all dart client libraries.

The following issues were first collected in #790:
- [ ] #1274
- [ ] #1275

These are planned as well:

- [ ] #1276
- [ ] #1277

spydon:
> [!NOTE]
> At the moment we're not looking for external contributions for the v3 issues.

PR #1349, opened 2026-04-10
fix(gotrue): replace assert with AuthException for PKCE asyncStorage check

### LINKED ISSUE #1319: Null check operator used on a null value
await supabase.auth.getOAuthSignInUrl(provider: OAuthProvider.github,);
 Null check operator used on a null value
 #0      GoTrueClient._getUrlForProvider (package:gotrue/src/gotrue_client.dart:1162:26)
#1      GoTrueClient.getOAuthSignInUrl (package:gotrue/src/gotrue_client.dart:324:12)
#2      main (file:///Users/apple/Desktop/publens_api/test_oauth.dart:19:41)
#3      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:314:19)
#4      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:193:12)

SamirChatwiti:
Fixes #1319

## Problem

When `flowType` is `AuthFlowType.pkce` (the default) and no `asyncStorage`
is provided, calling `getOAuthSignInUrl()` or `getLinkIdentityUrl()` crashes
in **release mode** with a cryptic `Null check operator used on a null value`
error. The existing `assert()` only runs in debug builds and is silently
stripped in release.

## Changes

- `_getUrlForProvider`: replace `assert` + `_asyncStorage!` with an explicit
  null check that throws a descriptive `AuthException` in all build modes.
- `getLinkIdentityUrl`: add null check on `res['url']` before casting,
  throwing a descriptive `AuthException` if the provider returns no URL.

## Before / After

```dart
// Before (crashes silently in release)
assert(_asyncStorage != null, 'You need to provide asyncStorage...');
await _asyncStorage!.setItem(...);

// After (throws clear error in all build modes)
final storage = _asyncStorage;
if (storage == null) {
  throw AuthException('asyncStorage is required for PKCE flow...');
}
await storage.setItem(...);

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -1069,7 +1069,14 @@ class GoTrueClient {
         jwt: _currentSession?.accessToken,
       ),
     );
-    return OAuthResponse(provider: provider, url: res['url']);
+    final url = res['url'] as String?;
+    if (url == null) {
+      throw AuthException(
+        'OAuth provider did not return a URL. '
+        'Ensure the provider is enabled in your Supabase project.',
+      );
+    }
+    return OAuthResponse(provider: provider, url: url);
   }
 
   /// Unlinks an identity from a user by deleting it.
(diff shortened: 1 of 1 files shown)


Vinzent03:
I think a better solution for this is to add an assert statement in the constructor of the gotrue client. But that is a breaking change and therefore add it in v3 #1278 of the library.

spydon:
I'll close this for now and we'll go with the assertiion check in v3 instead.
Thanks for your contribution though @SamirChatwiti!

PR #1356, opened 2026-04-16
fix(storage): handle null signedURL in createSignedUrls response

grdsdev:
## Summary

When a requested path does not exist, the Storage API returns `"signedURL": null` for that item. The Flutter SDK was silently producing broken URLs like `"${storageUrl}null"` with no way for callers to detect the issue. This mirrors the fix applied to the Swift SDK in supabase/supabase-swift#958.

## Changes

- **`types.dart`**: `SignedUrl.signedUrl` is now `String?` (nullable); added optional `error: String?` field populated from the per-item server error
- **`storage_file_api.dart`**: `createSignedUrl` throws `StorageException` when the API returns null (consistent with `createSignedUploadUrl`); `createSignedUrls` sets `signedUrl` to null and populates `error` for missing-path items
- **`test/basic_test.dart`**: Added tests for both null-signedURL scenarios; updated existing `createSignedUrl` test to assert the URL suffix

## Root Cause

- `storage_file_api.dart:402` — `'$url${e['signedURL']}'` with a null value produces the literal string `"${url}null"`
- `storage_file_api.dart:370` — same issue in the single-URL variant
- `types.dart:250` — `signedUrl` was non-nullable, making it impossible for callers to detect a missing path

### Reproduction Tests
- `createSignedUrl throws StorageException when signedURL is null` — previously returned broken URL, now throws
- `createSignedUrls returns null signedUrl and error for missing path` — previously returned broken URL, now returns `signedUrl: null, error: 'not_found'`

### Test Results
All 37 unit tests pass. Integration tests (`client_test.dart`) require a live storage server and are unaffected by this change.

## Breaking Changes

`SignedUrl.signedUrl` changes from `String` to `String?`. Callers that use this field without a null check will need to be updated. This is intentional — the previous non-null return was a false guarantee.

## Acceptance Criteria

- [x] `createSignedUrls` does not return items with `signedUrl == "${storageUrl}null"`
- [x] `createSignedUrl` throws `StorageException` when the API returns null signedURL
- [x] Unit tests added and passing for both null-signedURL scenarios
- [x] No regression in existing tests

## Linear Issue

Closes: [SDK-862](https://linear.app/supabase/issue/SDK-862)
Related: [SDK-851](https://linear.app/supabase/issue/SDK-851) (Swift), [SDK-842](https://linear.app/supabase/issue/SDK-842) (Python)

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) `/take`

### DIFF
--- a/packages/storage_client/lib/src/storage_file_api.dart
+++ b/packages/storage_client/lib/src/storage_file_api.dart
@@ -366,7 +366,11 @@ class StorageFileApi {
       },
       options: options,
     );
-    final signedUrlPath = (response as Map<String, dynamic>)['signedURL'];
+    final signedUrlPath =
+        (response as Map<String, dynamic>)['signedURL'] as String?;
+    if (signedUrlPath == null) {
+      throw StorageException('No signed URL returned by API');
+    }
     final signedUrl = '$url$signedUrlPath';
     return signedUrl;
   }
@@ -395,11 +399,11 @@ class StorageFileApi {
       options: options,
     );
     final List<SignedUrl> urls = (response as List).map((e) {
+      final signedUrlPath = e['signedURL'] as String?;
       return SignedUrl(
-        // Prevents exceptions being thrown when null value is returned
-        // https://github.com/supabase/storage-api/issues/353
         path: e['path'] ?? '',
-        signedUrl: '$url${e['signedURL']}',
+        signedUrl: signedUrlPath != null ? '$url$signedUrlPath' : null,
+        error: e['error'] as String?,
       );
     }).toList();
     return urls;
--- a/packages/storage_client/lib/src/types.dart
+++ b/packages/storage_client/lib/src/types.dart
@@ -246,16 +246,22 @@ class SignedUrl {
   /// The file path, including the current file name. For example `folder/image.png`.
   final String path;
 
-  /// Full signed URL of the files.
-  final String signedUrl;
+  /// Full signed URL of the file. Null when the path does not exist or the
+  /// server returned an error for this item; check [error] for details.
+  final String? signedUrl;
+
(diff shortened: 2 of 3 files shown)


packages/storage_client/lib/src/storage_file_api.dart
```diff
@@ -395,11 +399,11 @@ class StorageFileApi {
       options: options,
     );
     final List<SignedUrl> urls = (response as List).map((e) {
+      final signedUrlPath = e['signedURL'] as String?;
       return SignedUrl(
-        // Prevents exceptions being thrown when null value is returned
-        // https://github.com/supabase/storage-api/issues/353
```
Vinzent03, line 400: I think we should either keep this comment or remove the null catch for the `path` field. Or is that also null if `signedURL` is null?

packages/storage_client/lib/src/storage_file_api.dart
```diff
@@ -395,11 +399,11 @@ class StorageFileApi {
...
         path: e['path'] ?? '',
-        signedUrl: '$url${e['signedURL']}',
+        signedUrl: signedUrlPath != null ? '$url$signedUrlPath' : null,
+        error: e['error'] as String?,
```
Vinzent03, line 406: Does the `error` field only exists for requesting multiple urls or would that also exist for the case above of a single url request?

Vinzent03 (commented):
While yeah the previous behavior was not correct this is really a breaking change for everybody using the method for multiple urls. Their code will not compile as Dart is sound. So I would not release this. Either way I wanted to start working now on v3, so we could delay it for v3.

grdsdev:
Closed in favor of #1385

PR #1357, opened 2026-04-20
feat: optionally configure stream method to use a private channel

### LINKED ISSUE #1311: Using `stream` convenience function doesn't work with private-only realtime channels setting.

Vinzent03:

feature

Realtime channels that are created by the `stream` method are always public, which can be an issue if a project configuration is enabled to only allow private channels.

## What is the new behavior?

One can pass a realtime channel config to the `stream` method to allow making a channel private.

## Additional context

I chatted with a member of the realtime team and discussed something like this to be a reasonable solution. It is not possible to make all channels private by default as it is done in #1324 because even without actively using the broadcast channel, the RLS on the `realtime.messages` table needs to allow the joining for broadcast. So it needs to be decided by the library user whether a channel should be private or not.

close #1311

### DIFF
--- a/packages/supabase/lib/src/supabase_query_builder.dart
+++ b/packages/supabase/lib/src/supabase_query_builder.dart
@@ -29,7 +29,7 @@ class SupabaseQueryBuilder extends PostgrestQueryBuilder {
   ///
   /// Pass the list of primary key column names to [primaryKey], which will be used to update and delete the proper records internally as the stream receives real-time updates.
   ///
-  /// You may pass an optional [channelConfig] to configure the realtime channel to e.g., make it private.
+  /// The underlying [RealtimeChannel] is public by default. Set [private] to `true` to make it private, which requires additional RLS policies to be set up. See https://supabase.com/docs/guides/realtime/authorization for more details.
   ///
   /// It handles the lifecycle of the realtime connection and automatically refetches data from PostgREST when needed.
   ///
@@ -47,7 +47,7 @@ class SupabaseQueryBuilder extends PostgrestQueryBuilder {
   /// ```
   SupabaseStreamFilterBuilder stream({
     required List<String> primaryKey,
-    RealtimeChannelConfig? channelConfig,
+    bool private = false,
   }) {
     assert(primaryKey.isNotEmpty, 'Please specify primary key column(s).');
     return SupabaseStreamFilterBuilder(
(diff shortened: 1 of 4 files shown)


packages/supabase/lib/src/supabase_query_builder.dart
```diff
@@ -43,7 +45,10 @@ class SupabaseQueryBuilder extends PostgrestQueryBuilder {
   /// ```dart
   /// supabase.from('chats').stream(primaryKey: ['id']).eq('room_id','123').order('created_at').limit(20).listen(_onChatsReceived);
   /// ```
-  SupabaseStreamFilterBuilder stream({required List<String> primaryKey}) {
+  SupabaseStreamFilterBuilder stream({
+    required List<String> primaryKey,
+    RealtimeChannelConfig? channelConfig,
```
grdsdev: I think is better to add a new config type specific for configuring stream which contains only the `private` field, since the other ones doesn't apply here, and may confuse users.

Something like `StreamChannelConfig`.

grdsdev (changes requested):

packages/supabase/lib/src/supabase_query_builder.dart
Vinzent03: Yeah sounds valid as well. But since there is just one option for the moment I would propose to use just 'private' as an additional argument to 'primaryKey' and leave out the additional class. What do you think?

grdsdev: that works too

grdsdev (approved):

PR #1360, opened 2026-04-27
feat: rename anon key → publishable key, service role key → secret key

grdsdev:
## Summary

Updates API key terminology across code, tests, docs, and infra. No breaking changes — the old `anonKey` parameter is deprecated with a migration message pointing to the new name.

### `supabase_flutter` — `Supabase.initialize()`

`anonKey` is deprecated in favour of `publishableKey`. Both are accepted; `publishableKey` takes precedence if both are supplied.

```dart
// Before (still compiles, emits deprecation warning)
await Supabase.initialize(url: url, anonKey: key);

// After
await Supabase.initialize(url: url, publishableKey: key);
```

`publishableKey` is the right name for the Flutter package: apps run on the client and should always use the publishable (anon) key. The doc comment now explicitly warns against using the secret key here.

### `supabase` — `SupabaseClient`

The positional `supabaseKey` parameter is kept as-is — the plain Dart package is used in both client and server contexts, so a generic name is appropriate. The doc comment is updated to clarify the distinction:

> Pass the `publishable` (anon) key for client-side usage or the `secret` key for trusted server-side environments.

### Doc comments — `GoTrueAdminApi` / `GoTrueAdminOAuthApi`

References to `` `service_role` key `` updated to `` `secret` key `` to match the new naming.

## What was intentionally left unchanged

- JWT role claims (`'role': 'service_role'`) — internal backend values, unrelated to key naming
- `ANON_KEY` in `infra/storage_client/docker-compose.yml` — the storage binary reads this env var by name; renaming it would silently break the service
- CHANGELOG entries — historical entries are not refactored

### DIFF
--- a/packages/supabase/CHANGELOG.md
+++ b/packages/supabase/CHANGELOG.md
@@ -410,7 +410,7 @@
   ```dart
   // The following will initialize a supabase client that will retry failed uploads up to 25 times,
   // which is about 10 minutes of retrying.
-  final supabase = SupabaseClient('Supabase_URL', 'Anon_key', storageRetryAttempts: 25);
+  final supabase = SupabaseClient('Supabase_URL', 'Publishable_key', storageRetryAttempts: 25);
   ```
 - feat: update storage to v1.1.0
 - feat: update gotrue to v1.2.0
--- a/packages/supabase/CHANGELOG.md
+++ b/packages/supabase/CHANGELOG.md
@@ -410,7 +410,7 @@
   ```dart
   // The following will initialize a supabase client that will retry failed uploads up to 25 times,
   // which is about 10 minutes of retrying.
-  final supabase = SupabaseClient('Supabase_URL', 'Publishable_key', storageRetryAttempts: 25);
+  final supabase = SupabaseClient('Supabase_URL', 'Anon_key', storageRetryAttempts: 25);
   ```
 - feat: update storage to v1.1.0
 - feat: update gotrue to v1.2.0
--- a/packages/supabase_flutter/lib/src/supabase.dart
+++ b/packages/supabase_flutter/lib/src/supabase.dart
@@ -203,7 +203,7 @@ class Supabase {
 
   void _init(
     String supabaseUrl,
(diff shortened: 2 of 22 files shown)


Vinzent03:
The `SupabaseClient` in the `supabase` package calls the argument `supabaseKey`. It is positional so renaming it is no breaking change, but I think it can/should also be renamed to align it with the rest and update the documentation of it.

mandarini (approved):

grdsdev:
@Vinzent03 I decided to go with `supabaseKey` in all public facing APIs, since keys can be either publishable or secret, please review it again.

packages/supabase/CHANGELOG.md
```diff
@@ -410,7 +410,7 @@
   ```dart
   // The following will initialize a supabase client that will retry failed uploads up to 25 times,
   // which is about 10 minutes of retrying.
-  final supabase = SupabaseClient('Supabase_URL', 'Anon_key', storageRetryAttempts: 25);
```
Vinzent03, line 413: I don't think we should refactor old changelog entries.

packages/supabase_flutter/lib/src/supabase.dart
```diff
@@ -52,7 +52,7 @@ class Supabase {
   /// This must be called only once. If called more than once, an
   /// [AssertionError] is thrown
   ///
-  /// [url] and [anonKey] can be found on your Supabase dashboard.
+  /// [url] and [supabaseKey] can be found on your Supabase dashboard.
```
Vinzent03: The dashboard does not use the phrase `supabaseKey`, right? So stating that the user probably wants the publishable key might be helpful. Especially to prevent them from accidentally using the secret key on the client.
The `SupabaseClient` class documentation has that same line, so I would update its documentation as well. Though the `SupabaseClient` class is much more likely to be used with the secret key than the flutter package.

Vinzent03 (changes requested):

Vinzent03 (approved):

PR #1365, opened 2026-05-10
fix(realtime_client): suppress InvalidJWTToken from setAuth in joinPush 'ok' handler

### LINKED ISSUE #1363: realtime_client: setAuth call inside joinPush 'ok' callback can throw uncaught FormatException for expired token
## Summary

`RealtimeChannel`'s `joinPush.receive('ok', ...)` callback in [`realtime_channel.dart` (line ~156 in v2.4.2, ~179 on `main`)](https://github.com/supabase/supabase-flutter/blob/main/packages/realtime_client/lib/src/realtime_channel.dart#L179) calls:

```dart
if (socket.accessToken != null) {
  await socket.setAuth(socket.accessToken);
}
```

`setAuth` throws `FormatException('InvalidJWTToken: Invalid value for JWT claim "exp" with value …')` when the cached `socket.accessToken` has already expired. The callback is `async` and has no `try/catch`, so the exception escapes through the `Push.trigger` async chain into `runZonedGuarded`. With Sentry / Crashlytics installed it is logged as a fatal unhandled error, even though the app keeps running and the next `tokenRefreshed` event recovers normally.

The supabase wrapper [already handles this case](https://github.com/supabase/supabase-flutter/blob/main/packages/supabase/lib/src/supabase_client.dart#L375-L395) for the `_handleTokenChanged` path:

```dart
try {
  await realtime.setAuth(token);
} on FormatException catch (e) {
  if (e.message.contains('InvalidJWTToken')) {
    // ignored
  } else {
    rethrow;
  }
}
```

…but the `joinPush.receive('ok', ...)` re-auth path bypasses that wrapper.

## Reproduction

1. Subscribe to a Realtime channel (e.g. via `supabase.channel('x').onPostgresChanges(...).subscribe()`).
2. Let the device sleep / page go to background long enough for the access token's `exp` to lapse.
3. Resume — the channel reconnects, `joinPush` receives `'ok'`, and the callback calls `setAuth` with the stale token before the SDK gets a chance to refresh it.
4. The `FormatException` propagates out of the async callback to whatever zone error handler is installed.

Observed in our Flutter Web app on `realtime_client: 2.4.2` / `supabase: 2.6.3`. Code path is unchanged on `main`, so it should still reproduce on `^2.10`.

Stack (de-minified, from a Sentry event):

```
realtime_client/src/realtime_channel.dart   joinPush.receive 'ok' callback
realtime_client/src/realtime_client.dart    setAuth → throw FormatException('InvalidJWTToken …')
```

## Suggested fix

Wrap the `setAuth` call in the `'ok'` callback with the same `FormatException` / `InvalidJWTToken` filter the supabase wrapper uses, so that an expired cached token at re-subscribe time is silently swallowed (the auth refresh listener will re-call `setAuth` with a fresh token shortly after):

```dart
joinPush.receive('ok', (response) async {
  final serverPostgresFilters = response['postgres_changes'];
  if (socket.accessToken != null) {
    try {
      await socket.setAuth(socket.accessToken);
    } on FormatException catch (e) {
      if (!e.message.contains('InvalidJWTToken')) rethrow;
      // Stale cached token — auth refresh will re-call setAuth shortly.
    }
  }
  // …rest of handler
});
```

Happy to send a PR if useful — just want to confirm the desired behaviour first (silent swallow, vs. emitting on the channel's status callback).

## Versions

- `supabase: 2.6.3`
- `realtime_client: 2.4.2`
- `gotrue: 2.x`
- Flutter 3.x, web (dart2js, release).

theprantadutta:
## Summary

Fixes #1363.

`RealtimeChannel.subscribe()` registers an `async` `'ok'` callback on `joinPush` that calls `socket.setAuth(socket.accessToken)` without a `try`/`catch`. When the cached access token has expired by the time the server replies `'ok'` (e.g. the device woke from a long background sleep and the channel rejoined before the auth refresh fired), `setAuth` throws:

```
FormatException: InvalidJWTToken: Invalid value for JWT claim "exp" with value …
```

Because `Push.trigger` invokes the callback as a `void`-returning function, the resulting `Future`'s error is discarded and surfaces in the zone error handler. Sentry / Crashlytics report it as a fatal even though the app keeps running and the next `tokenRefreshed` event recovers normally.

`SupabaseClient._handleTokenChanged` already filters this exact exception on the auth-state-change path:

https://github.com/supabase/supabase-flutter/blob/main/packages/supabase/lib/src/supabase_client.dart#L379-L389

This PR mirrors that filter on the rejoin path so the two re-auth code paths behave consistently.

## Change

`packages/realtime_client/lib/src/realtime_channel.dart` — wrap the rejoin `setAuth` call in the same `FormatException` / `InvalidJWTToken` filter that `SupabaseClient._handleTokenChanged` uses. Other `FormatException`s still propagate.

```dart
if (socket.accessToken != null) {
  try {
    await socket.setAuth(socket.accessToken);
  } on FormatException catch (e) {
    if (!e.message.contains('InvalidJWTToken')) {
      rethrow;
    }
  }
}
```

## Tests

Added two unit tests in `packages/realtime_client/test/channel_test.dart`:

1. **Positive case** — when `setAuth` throws `FormatException('InvalidJWTToken: …')` on rejoin, the `'ok'` handler still runs to completion and emits `RealtimeSubscribeStatus.subscribed`. Without the fix this test fails (the callback aborts at `setAuth` and the subscribed status is never emitted), confirming it genuinely validates the fix.
2. **Rethrow case** — non-`InvalidJWTToken` `FormatException`s still abort the rejoin handler (subscribed is not emitted).

Both tests use a `RealtimeClient` subclass that overrides `setAuth` to throw a controlled exception and overrides `connect` to a no-op so transport-level async errors do not pollute the test runner zone.

## Test plan

- [x] `dart test packages/realtime_client/test/channel_test.dart` — both new tests pass; pre-existing tests unchanged
- [x] `dart analyze` clean for `realtime_client`, `supabase`, and `supabase_flutter`
- [x] Verified the positive test fails when the source change is reverted (confirming it tests the right thing)

> Note: `httpSend throws error on non-202 status` is flaky on `main` independently of this change (~40% fail rate locally on a clean checkout). Out of scope for this PR.

### DIFF
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -176,7 +176,19 @@ class RealtimeChannel {
         (response) async {
           final serverPostgresFilters = response['postgres_changes'];
           if (socket.accessToken != null) {
-            await socket.setAuth(socket.accessToken);
+            try {
+              await socket.setAuth(socket.accessToken);
+            } on FormatException catch (e) {
+              // The cached access token may have expired by the time the
+              // channel rejoins (e.g. after the device wakes from a long
+              // sleep). Auth state listeners will re-call setAuth with a
+              // fresh token shortly after, so swallow this specific error
+              // to avoid surfacing it as an uncaught exception. The same
+              // filter is applied in `SupabaseClient._handleTokenChanged`.
+              if (!e.message.contains('InvalidJWTToken')) {
+                rethrow;
+              }
+            }
           }
 
           if (serverPostgresFilters == null) {
--- a/packages/realtime_client/test/channel_test.dart
+++ b/packages/realtime_client/test/channel_test.dart
@@ -99,6 +99,76 @@ void main() {
     });
   });
 
+  group("joinPush 'ok' setAuth error handling", () {
+    // Re-authorizing the channel on rejoin can throw `FormatException`
+    // ('InvalidJWTToken: ...') when the cached access token has expired
+    // by the time the server responds with 'ok'. The handler must swallow
+    // that specific message so it does not escape as an unhandled async
+    // error (which Sentry/Crashlytics surface as a fatal). See
+    // https://github.com/supabase/supabase-flutter/issues/1363.
+    test(
+        "swallows FormatException with 'InvalidJWTToken' from setAuth and "
+        "still emits 'subscribed' status", () async {
+      final socket = _SetAuthThrowingSocket(
+        '/socket',
+        thrown: const FormatException(
+          'InvalidJWTToken: Invalid value for JWT claim "exp" with value 0',
+        ),
+      );
+      socket.accessToken = 'expired-token';
+
+      final channel = socket.channel('topic');
+
+      RealtimeSubscribeStatus? status;
+      channel.subscribe((s, _) => status = s);
+      channel.joinPush.trigger('ok', {});
+
+      // Drain the microtask queue so the async 'ok' callback completes.
+      await Future<void>.value();
+      await Future<void>.value();
+
+      expect(socket.setAuthCalls, 1);
+      expect(
+        status,
+        RealtimeSubscribeStatus.subscribed,
+        reason:
+            "If the catch is missing the 'ok' callback aborts at setAuth "
+            "and the subscribed status is never emitted.",
+      );
+    });
+
+    test(
+        "non-InvalidJWTToken FormatExceptions from setAuth still abort the "
+        "rejoin handler", () async {
+      final socket = _SetAuthThrowingSocket(
+        '/socket',
+        thrown: const FormatException('some other parsing failure'),
+      );
+      socket.accessToken = 'some-token';
+
+      final channel = socket.channel('topic');
+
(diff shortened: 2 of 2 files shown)


Vinzent03:
If you could fix the formatting issue, I think this looks great. Thanks!

theprantadutta:
Thanks for the review @Vinzent03! Formatting fixed in `34c2f35` — re-ran `dart format lib test -l 80 --set-exit-if-changed` locally to match CI, clean now. Tests still pass.

Vinzent03 (approved):

theprantadutta:
Hi @Vinzent03 — friendly nudge on this one. It's been about a week since the approval and I want to make sure nothing's blocking on my side. Happy to rebase or address anything new if needed.

Vinzent03:
I'm an external contributor and currently do not have the permission to merge without approval of a codeowner. The responsible Supabase member is currently OOF, but will surely be back soon. So there is nothing blocking from your side.

spydon (approved):

Issue #898, opened 2024-04-23
Add support for aborting/canceling requests


Dev-Salem:
**Is your feature request related to a problem? Please describe.**
It's a common use case to cancel/abort a query when it's no longer needed (e.g user scrolling fast in a paginated list, or user leaving the page/ closing a dialog before the query returns).
**Describe the solution you'd like**
Add support to cancel requests just like the [JS Library](https://github.com/orgs/supabase/discussions/5992)

Vedsaga:
Assuming `http` quest need to be cancelled, which unfortunately is not yet supported officially in `http` package to abort/cancel the `http` request [ Design an API to support aborting requests.](https://github.com/dart-lang/http/issues/424)

devphilou:
Is there any update on this feature request? Since last month the official `http` package supports to abort requests:
- https://github.com/dart-lang/http/issues/424#issuecomment-3049897637
- https://pub.dev/packages/http/versions/1.5.0-beta#aborting-requests

Vinzent03:
Requires a version bump of the `http` package to `^1.5.0` so we can do it in v3.

spydon:
Most of it was shipped in #1368 and #1593, opening #1829 for aboring storage requests.

PR #1368, opened 2026-05-29
feat(postgrest): abort requests

Vinzent03:

feature

One cannot abort a started http request. Even by adding a `.timeout` to the feature only ignores the result of the future but does not cancel its execution. 

## What is the new behavior?

Starting with `http` v1.5.0` it is now possible to abort requests.

```dart
final abortCompleter = Completer<void>();
Timer(Duration(seconds: 5), () => abortCompleter.complete());
try {
  final response = await client
  .from('table')
  .select()
  .abortCompleter(abortCompleter);
} on RequestAbortedException catch (e) {
 print('Request was aborted: $e');
}
```

Supabase-js uses `.abortSignal`, but since the dart implementation uses futures/completer I switched to `.abortCompleter`. But I'm not entirely sure and open for discussion about choosing a different name.

related: #898 
The issue covers the general feature to abort requests, so we can cover aborting the requests in functions in another pr.


Vinzent03:
For now waiting for the Dart sdk bump across all packages.

Vinzent03:
Now that we've already bumped the minimum versions in v2 this has not to wait for v3.

spydon (approved):
Lgtm, I'll make sure that we add an example in the docs for this.

spydon (commented):
LLM review:

Two things I'd like to sort out before merging:

1. **API shape.** The builder only ever reads `abortCompleter.future`, it never completes it. Taking a `Completer` forces callers to allocate one even when they already have a future. Accepting the trigger directly is strictly more flexible and matches how `http`'s `Abortable` models it:

   ```dart
   final response = await client.from('table').select()
       .abortSignal(Future.delayed(Duration(seconds: 5)));
   ```

   On naming: you raised this yourself, and I'd lean towards `abortSignal(Future<void> signal)`. It mirrors supabase-js's `.abortSignal`, and a `Future` is the natural Dart analog of an `AbortSignal`. If we change the name, let's update the `sdk-compliance.yaml` symbol to match.

2. **The exception isn't reachable by consumers.** `RequestAbortedException` comes from `package:http` and isn't re-exported by `postgrest` (or `supabase`), so the `on RequestAbortedException` catch in your example only compiles if the user adds `http` as a direct dependency. Could we re-export it from `postgrest.dart`, or wrap it in a postgrest-native exception?

A few minor things, none blocking:

- The empty `if`/`else` branches in the new `send` closure are a bit awkward. Setting `request.body` only for post/put/patch and validating unknown methods separately would read cleaner.
- Bumping the shared `_buildClient` retry delay to 500ms plus the mock's 200ms delay slows the whole retry suite noticeably. Could the delay be scoped to just the abort test?
- The `basic_test` abort test catches the exception and only prints it, relying on the elapsed-time bounds. An explicit `throwsA(isA<RequestAbortedException>())` would fail louder if abort silently regressed.

Deferring functions/storage to a follow-up is fine by me.

More info:

The PR's Completer is one indirection above the http primitive. It only unwraps to .future internally anyway, so wrapping it adds nothing. Two supporting points from http's own docs:

- It explicitly blesses both patterns: Completer for event-driven abort, Future.delayed for timeouts. Accepting a Future<void> supports both; accepting a Completer only covers the event case cleanly (a timeout caller would have to build a Completer + Timer instead of a one-liner Future.delayed).
- "This future must not complete with an error." — a contract worth mirroring in our doc comment.

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -74,6 +75,7 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
     PostgrestConverter<S, R>? converter,
     bool retryEnabled = true,
     @visibleForTesting Duration Function(int attempt)? retryDelay,
+    Future? abortSignal,
```
spydon:
```suggestion
    Future<void>? abortSignal,
```

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -56,6 +56,7 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
   final CountOption? _count;
   final bool _retryEnabled;
   final Duration Function(int attempt) _retryDelay;
+  final Future? _abortSignal;
```
spydon:
```suggestion
  final Future<void>? _abortSignal;
```

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -100,6 +103,7 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
     PostgrestConverter<S, R>? converter,
     bool? retryEnabled,
     Duration Function(int attempt)? retryDelay,
+    Future? abortSignal,
```
spydon:
```suggestion
    Future<void>? abortSignal,
```

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -126,6 +131,50 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
...
+  /// try {
+  ///   final response = await client
+  ///   .from('table')
+  ///   .select()
+  ///   .abortSignal(Future.delayed(Duration(seconds: 5)));
+  /// } on RequestAbortedException catch (e) {
+  ///  print('Request was aborted: $e');
+  /// }
+  /// ```
+  PostgrestBuilder<T, S, R> abortSignal(Future abortSignal) {
```
spydon:
```suggestion
  PostgrestBuilder<T, S, R> abortSignal(Future<void> abortSignal) {
```

packages/postgrest/lib/src/postgrest_builder.dart
Vinzent03: I changed that by design to be more flexible, but forgot you can just assign every future to a `Future<void>`. So thanks!

spydon: Yeah, it's a bit unintuitive that you can assign any Future to Future<void>, I forget that sometimes too

spydon (approved):

PR #1385, opened 2026-06-08
fix(storage): handle null signedURL in createSignedUrls response

grdsdev:
## Summary

When a requested path does not exist, the Storage API returns `"signedURL": null` for that item. The Flutter SDK was silently producing broken URLs like `"${storageUrl}null"` with no way for callers to detect the issue.

This is a backward-compatible reimplementation of #1356, using [supabase/supabase-swift#958](https://github.com/supabase/supabase-swift/pull/958) as the reference. Because Dart does not support method overloading (unlike Swift's `@_disfavoredOverload`), the new preferred API gets the name `createSignedUrlsResult` and the old one is deprecated — this will be cleaned up in v3 ([SDK-1002](https://linear.app/supabase/issue/SDK-1002)).

## Changes

- **`types.dart`**: Add `SignedUrlResult` sealed class with `SignedUrlSuccess` and `SignedUrlFailure` subtypes — mirrors Swift's `SignedURLResult` enum
- **`storage_file_api.dart`**:
  - `createSignedUrl` now throws `StorageException` when the API returns a null `signedURL` instead of producing a malformed string
  - `createSignedUrls` is deprecated; it delegates to `createSignedUrlsResult` and silently drops failures (matching Swift's `compactMap` on the deprecated overload)
  - New `createSignedUrlsResult` returns `List<SignedUrlResult>` — one entry per requested path with a clear success/failure distinction
- **`test/basic_test.dart`**: Tests for the null `signedURL` throw, mixed success/failure results, and deprecated drop behaviour

## Backward Compatibility

No breaking changes:
- `SignedUrl.signedUrl` stays `String` (non-nullable)
- `createSignedUrls` still compiles — callers get a deprecation hint pointing to `createSignedUrlsResult`
- `createSignedUrl` throwing on null is a bug fix (previously it returned a broken string)

## Testing

All 38 unit tests pass.

## Linear

Closes: [SDK-862](https://linear.app/supabase/issue/SDK-862)
v3 cleanup: [SDK-1002](https://linear.app/supabase/issue/SDK-1002)

### DIFF
--- a/packages/storage_client/lib/src/storage_file_api.dart
+++ b/packages/storage_client/lib/src/storage_file_api.dart
@@ -366,22 +366,53 @@ class StorageFileApi {
       },
       options: options,
     );
-    final signedUrlPath = (response as Map<String, dynamic>)['signedURL'];
+    final signedUrlPath =
+        (response as Map<String, dynamic>)['signedURL'] as String?;
+    if (signedUrlPath == null) {
+      throw StorageException('No signed URL returned by API');
+    }
     final signedUrl = '$url$signedUrlPath';
     return signedUrl;
+  }
+
+  // TODO(v3): Remove this deprecated overload and rename createSignedUrlsResult
+  // to createSignedUrls. Dart lacks overloading so the preferred API had to be
+  // given a temporary name. https://linear.app/supabase/issue/SDK-1002
+  /// Create signed URLs to download files without requiring permissions.
+  ///
+  /// Items for paths that do not exist are silently omitted. Use
+  /// [createSignedUrlsResult] to distinguish missing paths from successful ones.
+  ///
+  /// [paths] is the file paths to be downloaded, including the current file
+  /// names. For example: `createSignedUrls(['folder/image.png', 'folder2/image2.png'])`.
+  ///
+  /// [expiresIn] is the number of seconds until the signed URLs expire. For
+  /// example, `60` for URLs which are valid for one minute.
+  @Deprecated('Use createSignedUrlsResult to handle missing paths correctly.')
+  Future<List<SignedUrl>> createSignedUrls(
+    List<String> paths,
+    int expiresIn,
+  ) async {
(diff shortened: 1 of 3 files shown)


packages/storage_client/lib/src/storage_file_api.dart
```diff
@@ -394,15 +425,18 @@ class StorageFileApi {
...
+    return (response as List).map<SignedUrlResult>((e) {
+      final signedUrlPath = e['signedURL'] as String?;
+      final path = e['path'] as String? ?? '';
+      if (signedUrlPath != null) {
+        return SignedUrlSuccess(path: path, signedUrl: '$url$signedUrlPath');
+      } else {
+        return SignedUrlFailure(
+          path: path,
+          error: e['error'] as String? ?? 'Unknown error',
+        );
```
spydon, line 437: I just realized that since this method is returning a future we should throw an exception here instead of `SignedUrlFailure`, because otherwise the future will come back as a success but with an error body, which is kind of strange. So the sealed class isn't needed after all.

grdsdev: The batch case is exactly why a sealed type is needed here — if we throw on the first  signedURL, we discard all the successfully-signed URLs that came back in the same response.

For example, if a caller requests 10 paths and 2 don't exist, the API returns 10 items (8 with a real signed URL, 2 with `signedURL: null`). Throwing would lose the 8 good results.

This is the same reasoning behind the Swift reference implementation (supabase/supabase-swift#958), which uses a `SignedURLResult` enum for the same reason: each item in the response independently either succeeded or failed.

The `createSignedUrl` (single path) does throw, because there's nothing to partially succeed on.

Happy to discuss further if you think there's a cleaner way to model this in Dart!

spydon: Aaah right, that makes sense then!

spydon (approved):

PR #1387, opened 2026-06-09
fix(postgrest): include caller frames in PostgrestException stack traces

### LINKED ISSUE #899: Postgrest giving stacktrace of only postgrest code
**Describe the bug**
Whenever I use database function on the supabase client (`supabase.rpc` or `supabase.from`) and these throw an error then the stack trace is limited to postgrest code only and non of my app lines (other supabase functions do give a full stacktrace like calling `supabase.signInWithPassword` with an invalid email so I don't think it's my logging setup.

For example:
```
I/flutter ( 4414): PostgrestException(message: Could not find the function public.nearby_venues1(current_lat, current_long, limit_param, utc_offset_minutes_param) in the schema cache, code: PGRST202, details: Searched for the function public.nearby_venues1 with parameters current_lat, current_long, limit_param, utc_offset_minutes_param or with a single unnamed json/jsonb parameter, but no matches were found in the schema cache., hint: Perhaps you meant to call the function public.nearby_venues), #0 PostgrestBuilder._parseResponse (package:postgrest/src/postgrest_builder.dart:283:7)
I/flutter ( 4414): #1 PostgrestBuilder._execute (package:postgrest/src/postgrest_builder.dart:164:14)
I/flutter ( 4414): <asynchronous suspension>
I/flutter ( 4414): #2 PostgrestBuilder.then (package:postgrest/src/postgrest_builder.dart:372:24)
I/flutter ( 4414): <asynchronous suspension>
```

**To Reproduce**
Call a supabase function which will give an error like `await supabase.rpc('invalid_function')`

**Expected behavior**
The stacktrace should should lines in my app code which led up to the error.

spydon:
## Summary

- Captures `StackTrace.current` synchronously at `PostgrestBuilder` construction time (inside the constructor initializer list), which always runs in user code context when chaining filter methods like `.select()`, `.eq()`, etc.
- Rewrites `then()` to be non-`async`, using an intermediate native `Future.then()` to enable Dart's Zone causal-chain machinery.
- Appends the construction-time trace to every error via `StackTrace.fromString('$internalStack\n$_callerTrace')` and re-throws with `Error.throwWithStackTrace`, so both `await builder` (plain usage) and explicit `.then(onError:)` usage show user frames.

Fixes #899.

## Test plan

- [x] `dart test test/stack_trace_test.dart` passes both cases: plain `await builder` and explicit `.then()` with `onError`
- [x] `dart test test/retry_test.dart` still passes (no regression in retry behavior)
- [x] Integration tests pass against a live PostgREST instance

### DIFF
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -442,49 +442,64 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
   Future<U> then<U>(
     FutureOr<U> Function(T value) onValue, {
     Function? onError,
-  }) async {
+  }) {
     if (onError != null &&
         onError is! Function(Object, StackTrace) &&
         onError is! Function(Object)) {
-      throw ArgumentError.value(
+      return Future.error(ArgumentError.value(
         onError,
         "onError",
-        "Error handler must accept one Object or one Object and a StackTrace"
-            " as arguments, and return a value of the returned future's type",
-      );
+        "Error handler must accept one Object or one Object and a StackTrace "
+            "as arguments, and return a value of the returned future's type",
+      ));
     }
 
-    try {
-      final response = await _execute();
-      return onValue(response);
-    } catch (error, stack) {
-      final FutureOr<U> result;
-      if (onError != null) {
-        if (onError is Function(Object, StackTrace)) {
-          result = onError(error, stack);
-        } else if (onError is Function(Object)) {
-          result = onError(error);
-        } else {
-          throw ArgumentError.value(
-            onError,
-            "onError",
-            "Error handler must accept one Object or one Object and a StackTrace"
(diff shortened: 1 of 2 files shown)


grdsdev (approved):

packages/postgrest/lib/src/postgrest_builder.dart
```diff
@@ -442,49 +442,64 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
...
+            Error.throwWithStackTrace(rethrown, enrichedStack);
+          }
+          rethrow;
         }
+      } else {
+        throw ArgumentError.value(
+          onError,
+          "onError",
+          "Error handler must accept one Object or one Object and a StackTrace"
+              " as arguments, and return a value of the returned future's type",
```
spydon:
```suggestion
          "Error handler must accept one Object or one Object and a StackTrace "
              "as arguments, and return a value of the returned future's type",
```

PR #1390, opened 2026-06-09
fix(ci): use package tag ref in publish workflow instead of inputs

spydon:
## Summary

The publish workflow (`release-publish.yml`) required explicit `package-name` and `package-version` inputs, which forced the tag workflow to wrap the `gh workflow run` call in `bash -c` with `--field` params and a `sleep 30` between dispatches. This complexity broke `--no-published` filtering in melos, causing publish workflows to not be triggered after release commits.

Instead, drop the inputs and let the workflow dispatch `--ref` (the package tag) drive the checkout. The melos-action with `publish: true` uses `--no-published` at that ref to determine what to publish, and `create-release: true` handles GitHub releases — the same approach used in the Flame monorepo.

## Changes

- `release-publish.yml`: removed `workflow_dispatch` inputs and explicit `ref:` override in checkout
- `release-tag.yml`: restored `--no-published`, removed `bash -c` wrapper, `--field` params, and `sleep 30`

## Test plan

- [ ] Verify the next release commit triggers publish workflows for all newly tagged packages
- [ ] Verify each triggered `release-publish.yml` run checks out at the correct package tag and publishes only the unpublished package(s)

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -2,18 +2,9 @@ name: Publish Packages
 
 on:
   workflow_dispatch:
-    inputs:
-      package-name:
-        description: 'Name of the package to publish'
-        required: true
-        type: string
-      package-version:
-        description: 'Version to publish (without v prefix, e.g. 2.18.0)'
-        required: true
-        type: string
 
 concurrency:
-  group: release-publish-${{ inputs.package-name }}
+  group: release-publish-${{ github.ref }}
   cancel-in-progress: false
 
 permissions:
@@ -36,7 +27,6 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
-          ref: ${{ inputs.package-name }}-v${{ inputs.package-version }}
 
       - name: Setup Flutter
         uses: subosito/flutter-action@1a449444c387b1966244ae4d4f8c696479add0b2 # v2.23.0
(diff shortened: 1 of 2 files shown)


grdsdev (approved):

.github/workflows/release-publish.yml
```diff
@@ -2,18 +2,9 @@ name: Publish Packages
...
-        required: true
-        type: string
-      package-version:
-        description: 'Version to publish (without v prefix, e.g. 2.18.0)'
-        required: true
-        type: string
 
 concurrency:
-  group: release-publish-${{ inputs.package-name }}
+  group: release-publish-${{ github.ref }}
```
QuintinWillison, line 7: I'm not sure what this is protecting at this point. Locking workflow-run concurrency around package name made sense as you would want to prevent two releases for the same package happening at the same time. The angle that `github.ref` is coming from is _very_ different, representing the nature of the specific release itself. Perhaps you could consider picking something static here like the repo-name?

spydon: Honestly the concurrency lock is not really needed here since this job only is triggered when another job creates the version tags.
If we for some weird reason would want to trigger several releases of the same package but different versions at the same time the `github.ref` (the tag) is correct though, we should only block if we try to publish the same version twice (which wouldn't really matter because pub would block it anyway).
So technically it is correct like this, but also not needed.

QuintinWillison: ok, maybe I was looking at this from the wrong angle. I was assuming it would be in place to enforce "only one release taking place at a time", regardless of version of that release.

PR #1392, opened 2026-06-11
feat(gotrue): add passkey (WebAuthn) support

### LINKED ISSUE #1369: Feature Request: Passkeys / WebAuthn support in Flutter SDK
### Description

With Passkeys (WebAuthn) now available in the Supabase Dashboard under Authentication > Configuration > Passkeys (BETA), it would be great to have support for this in `supabase-flutter`.

Are there plans or a roadmap to add support for passkey registration and sign-in to the Flutter SDK? 

### Use Case
We would like to allow our users to sign up and log in using platform passkeys (biometrics/FaceID/TouchID/security keys) natively in our Flutter app, synced with Supabase Auth.

Any details, timelines, or suggested workarounds for integrating passkeys with `supabase-flutter` currently would be extremely helpful!

spydon:

Feature. Closes #1369.

Adds support for the Supabase Auth Passkeys (BETA) API (`/passkeys/*` endpoints), mirroring the two-step `auth.passkey.*` API surface of auth-js.

## What is the new behavior?

New `supabase.auth.passkey` namespace (`GoTruePasskeyApi`):

- `startRegistration()` / `verifyRegistration(challengeId:, credential:)` to register a passkey for the signed-in user
- `startAuthentication(captchaToken:)` / `verifyAuthentication(challengeId:, credential:)` to sign in with a passkey; persists the session and fires `AuthChangeEvent.signedIn`
- `list()`, `update(passkeyId:, friendlyName:)`, `delete(passkeyId:)` for management

New `supabase.auth.admin.passkey` namespace (`GoTrueAdminPasskeyApi`) with `listPasskeys(userId:)` and `deletePasskey(userId:, passkeyId:)`.

Supporting changes:

- New types: `Passkey`, `PasskeyRegistrationOptionsResponse`, `PasskeyAuthenticationOptionsResponse`
- PATCH support in `GotrueFetch` (used by the passkey rename endpoint)
- `FactorType.webauthn`, a `webauthn` bucket in `mfa.listFactors()`, `AMRMethod.passkey` and `AMRMethod.mfaWebauthn`
- New server error codes in `ErrorCode` (`passkey_disabled`, `webauthn_*`, `mfa_webauthn_*`, `too_many_passkeys`)

## Design notes

- The native WebAuthn ceremony (FaceID/TouchID/security key prompt) is left to the app via a platform plugin (e.g. `credential_manager`, or `navigator.credentials` on web). Options and credentials are exchanged as `Map<String, dynamic>` in the W3C WebAuthn L3 JSON format, which is what those plugins accept and produce. This matches the two-step `auth.passkey.start*/verify*` API in auth-js; its one-call `signInWithPasskey()` performs the ceremony in-browser, which a pure Dart package cannot do cross-platform. A high-level helper on top of a pluggable authenticator interface is a possible follow-up, as is the separate MFA WebAuthn factor flow.
- auth-js gates this API behind `experimental: { passkey: true }` with a call-time throw. Instead of plumbing a runtime flag through three packages, the API is annotated with `@experimental` from `package:meta` so consumers get analyzer warnings. The server independently rejects calls with `passkey_disabled` when the dashboard toggle is off.

## Tests

31 unit tests against a mock server client covering all endpoints and wire format (snake_case bodies, JWT headers, 204 deletes, top-level array responses), session/event side effects, model parsing edge cases, and error mapping. End-to-end tests against the dockerized GoTrue server are not feasible since the server cryptographically verifies attestations/assertions, which would require a software authenticator; auth-js has the same limitation.

## Additional context

Passkeys must be enabled per project in the Dashboard under Authentication > Configuration > Passkeys (BETA). Docs (Dart reference spec and the passkeys guide in supabase/supabase) will follow after release.


grdsdev:
Looks good!

I'd like to see an e2e app implementing this so we can really make sure it works.

As a follow up, write a complete guide on how to do the full native ceremony using an well known platform plugin, would be even better if we could offer that in the SDK so users don't need to implement it themselves.

For Swift, I'm adding the `signInWithPasskey` method under a `#if canImport(AuthenticationService)` check, can't we do something similar for Flutter?

spydon:
> I'd like to see an e2e app implementing this so we can really make sure it works.

I have that locally, I can post it in a temporary repo on Monday.

> As a follow up, write a complete guide on how to do the full native ceremony using an well known platform plugin

That's the plan!

> For Swift, I'm adding the signInWithPasskey method under a #if canImport(AuthenticationService) check, can't we do something similar for Flutter?

Unfortunately there aren't conditional dependencies in Dart, but we do have full tree shaking, so might be worth adding a dependency anyways. The only potential problem is future dependency version resolution problems, but we should be able to handle those.

grdsdev (approved):
Approved!

PR #1411, opened 2026-06-15
test: run integration tests against the Supabase CLI stack

spydon:
> **Stacked on #1410.** Base is `fix/gotrue-getclaims-asymmetric-jwt`; the migrated gotrue tests need that fix to pass under the CLI's asymmetric tokens. Retarget to `main` once #1410 merges.

Resolves SDK-1072

## What

Replaces the four per-package `docker-compose` setups (`infra/`) with a single Supabase CLI stack under `supabase/`. CI now installs a pinned Supabase CLI and runs `supabase start`/`stop` instead of `docker compose`; the package tests point at the unified gateway (`/auth/v1`, `/rest/v1`, `/realtime/v1`, `/storage/v1`) and the database it exposes.

## Why

The hand-written compose files pinned service versions by hand and drifted from real Supabase. Using the CLI keeps the local/CI stack aligned with what users actually run, matching the approach in `supabase-js`.

## Key points

- **Single config at the repo root** (`supabase/config.toml`, `migrations/`, `seed.sql`) serving all four backend packages; ports are offset by +100 (gateway `54421`, db `54422`) so the stack can run alongside another local Supabase project.
- **Committed RS256 signing key** (`supabase/signing_keys.json`): a dedicated, local-only test key. It makes the issued tokens deterministic and uses RS256 (which the Dart client verifies) instead of the CLI default ES256. The `anon`/`service_role` test tokens hardcoded in the suites are signed by it, so it must not be regenerated.
- Migrations/seed recreate the schemas, functions (including the `reset_and_init_auth_data` RPC), `todos` publication and dummy data the tests rely on.
- A few assertions were updated for genuine newer-server behavior (`PGRST205` for missing tables, the storage `emptyBucket` "queued" message).
- `infra/` removed; `README.md`, `AGENTS.md` and the realtime README updated to use the CLI.

## Validation

Locally against the CLI stack: postgrest (151), gotrue (422), storage (79) and realtime (139, including the v1.0.0 and v2.0.0 Dockerized integration tests) all pass.


.github/workflows/test.yml
```diff
@@ -13,6 +13,10 @@ concurrency:
 permissions:
   contents: read
 
+env:
+  # Pin the Supabase CLI so test behaviour (service versions, JWT signing) stays reproducible.
+  SUPABASE_CLI_VERSION: '2.106.0'
```
grdsdev: I prefer not to pin the version and use `latest` so we can find any issues with the CLI earlier, but I understand it makes CI run unreproducible, so it's your call.

grdsdev (approved):

.github/workflows/test.yml
spydon: Yeah I was 50/50 on that, so if you prefer latest, let's go with latest. We can log the CLI version it's running with so that we can try again with that version if we need to later, also I think it'll be quite rare that we'll need this.

PR #1414, opened 2026-06-16
ci: add DCM analysis job to the pipeline

spydon:
Adds the `dcm` job to the Test workflow. Stacks on top of #1416, which adds the `supabase_lints` shared config and the clean DCM baseline, so this job runs against a baseline that already passes.

## What

- New `dcm` job in `.github/workflows/test.yml`: bootstraps the melos workspace, installs DCM (`CQLabs/setup-dcm` pinned to `8697ae0…` / `v2.0.1`), runs `dcm analyze packages` using the `DCM_CI_KEY` and `DCM_EMAIL` repository secrets, and is wired into the `test-result` gate.

## Base

This PR targets `dcm-baseline` (#1416), not `main`. Merge #1416 into `main` first; GitHub will then retarget this PR to `main` automatically.


.github/workflows/test.yml
```diff
@@ -330,6 +330,35 @@ jobs:
...
+          dart pub global activate melos
+          melos bootstrap
+
+      - name: Install DCM
+        uses: CQLabs/setup-dcm@8697ae0790c0852e964a6ef1d768d62a6675481a # v2.0.1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run DCM
+        run: dcm analyze --ci-key="${{ secrets.DCM_CI_KEY }}" --email="${{ secrets.DCM_EMAIL }}" packages
```
grdsdev, line 360: where does it output results?

spydon: Directly in the job output

grdsdev (approved):

PR #1444, opened 2026-06-22
fix(supabase_flutter): accept PasskeyAuthenticatorInterface instead of depending on passkeys

spydon:
## What

Removes the direct dependency on the [`passkeys`](https://pub.dev/packages/passkeys) plugin. `supabase_flutter` now depends only on [`passkeys_platform_interface`](https://pub.dev/packages/passkeys_platform_interface) and the experimental passkey methods accept a `PasskeyAuthenticatorInterface`:

```dart
import 'package:passkeys/authenticator.dart';

final authenticator = PasskeyAuthenticator();

await supabase.auth.signInWithPasskey(authenticator);
await supabase.auth.registerPasskey(authenticator);
```

## Why

`supabase_flutter` only needs the WebAuthn request/response types and the two ceremony methods. Bundling the full `passkeys` plugin forced **every** app, even ones that never touch passkeys, to raise minimum platform versions to iOS 16.0 / macOS 13.5 and pull in transitive deps (`passkeys_doctor` → `device_info_plus`). Depending on the platform interface and accepting an authenticator instance removes that.

## How

- `pubspec.yaml`: `passkeys` → `passkeys_platform_interface: ^2.7.0`.
- `registerPasskey` / `signInWithPasskey` take a `PasskeyAuthenticatorInterface`. Both are `@experimental`, so no major version bump.
- `passkey_options_mapper.dart`: imports types from `passkeys_platform_interface`.
- Removed the now-unnecessary `device_info_plus` `dependency_overrides` in the example (verified the example no longer pulls `passkeys`/`device_info_plus` when the local package is linked).
- Updated the README passkeys section, including the `passkeys >=2.21.0` floor consumers need for `PasskeyAuthenticator` to implement the interface.

## Upstream dependency (now published)

This builds on **corbado/flutter-passkeys#247**, which adds `PasskeyAuthenticatorInterface` to `passkeys_platform_interface` and makes `passkeys`'s `PasskeyAuthenticator` implement it. Both changes are non-breaking for those packages (the method signatures already match exactly, so `implements` only adds a supertype).

That PR is merged and the packages are published:
- `passkeys_platform_interface` **2.7.0** carries the interface (pinned here).
- `passkeys` **2.21.0** is the first release whose `PasskeyAuthenticator` implements it; that is the documented floor for consumers.

## Verification

- `flutter pub get` resolves `passkeys_platform_interface` to the published `2.7.0`.
- `flutter analyze` clean on the changed sources.
- Mapper and package tests pass.


grdsdev (approved):

Vinzent03:
Would it be possible to just pass the `authenticator` without the specific method to the `signInWithPasskey` method? Such that the method in the flutter sdk calls the `authenticate` or `register` method on the passed `authenticator`?

spydon:
> Would it be possible to just pass the `authenticator` without the specific method to the `signInWithPasskey` method? Such that the method in the flutter sdk calls the `authenticate` or `register` method on the passed `authenticator`?

That was my first idea too, but unfortunately not, since `PasskeyAuthenticator` lives in the `passkeys` package and not in the interface package. The interface package only exposes the `PasskeysPlatform` class and if the user has to pass that one we'd lose a lot of error handling etc.

packages/supabase_flutter/README.md
```diff
@@ -293,30 +293,36 @@ supabase.auth.onAuthStateChange.listen((data) {
...
+await supabase.auth.registerPasskey(authenticator);
 
 // Manage the signed in user's passkeys with the server side API.
 final passkeys = await supabase.auth.passkey.list();
 await supabase.auth.passkey.update(passkeyId: passkeys.first.id, friendlyName: 'My phone');
 await supabase.auth.passkey.delete(passkeyId: passkeys.first.id);
 ```
 
-The ceremony is handled by the [`passkeys`](https://pub.dev/packages/passkeys) plugin. Because the plugin is bundled with `supabase_flutter`, every app raises its minimum platform versions to **iOS 16.0** and **macOS 13.5** (the OS versions where the passkey APIs became available), regardless of whether passkeys are used. Set your deployment targets accordingly.
+The platform ceremony is handled by whichever plugin you add. The [`passkeys`](https://pub.dev/packages/passkeys) plugin raises an app's minimum platform versions to **iOS 16.0** and **macOS 13.5** (the OS versions where the passkey APIs became available), so set your deployment targets accordingly when you depend on it.
```
Vinzent03: I think the exact details of minimum platform versions rather belong to the respective package and not in this readme.

packages/supabase_flutter/README.md
```diff
@@ -293,30 +293,36 @@ supabase.auth.onAuthStateChange.listen((data) {
...
 
 It also requires per platform setup that the library cannot do for you:
 
 - **iOS / macOS**: add the Associated Domains capability with `webcredentials:<your-domain>` and host an `apple-app-site-association` file at `https://<your-domain>/.well-known/apple-app-site-association`.
 - **Android**: host an `assetlinks.json` (Digital Asset Links) at `https://<your-domain>/.well-known/assetlinks.json` containing your package name and signing certificate SHA-256 fingerprint.
 - **Web**: include the `passkeys` web SDK `bundle.js` in your `index.html` (download it from the [`flutter-passkeys` releases](https://github.com/corbado/flutter-passkeys/releases)).
```
Vinzent03: But I see the previous pr already added quite some information that is specific to the passkey package. I would prefer to keep our README simpler and as you've notes here multiple times the supabase package works with any passkey package that implements the interface so I would rather just link to the package and its instructions. But I also cannot find the android instructions in the passkey's README so what about creating a pr to that repo?

Issue #1202, opened 2025-07-11
Session expiresAt always corrupted after hot restart – session always expired even after login


Ranojitdas:
Steps to Reproduce:
Use the latest supabase_flutter and supabase packages:
Apply to home_screen....
Log in with email and password.
Hot restart or relaunch the app.
Observe that:
Supabase.instance.client.auth.currentUser is present.
Supabase.instance.client.auth.currentSession is present, but expiresAt is always a value from 1970.
The app always navigates to the login screen because the session is considered expired.
debug output : 
[SupabaseService] Supabase initialized. Current user: User(id: ...)
[SplashScreen] Current user at splash: ...
[SplashScreen] Session: {..., expires_at: 181411429, ...}
[SplashScreen] Session expires at: 1970-01-21T12:14:11.429, now: 2025-07-11T21:00:51.211994
[SplashScreen] Session expired, attempting to refresh session...
supabase.auth: INFO: Refresh session 
[SplashScreen] Session refresh attempted.
[SplashScreen] User after possible refresh: ...
[SplashScreen] Auth status: AuthStatus.sessionExpired
[SplashScreen] Auth status not logged in/session expired, navigating to /login

Upgraded to the latest supabase_flutter and supabase packages.
Cleared all SharedPreferences and secure storage before login.
Confirmed that after login, session tokens are saved and everything works until restart.
No custom session persistence or manual session handling in my code.
Expected Behavior:
After hot restart or relaunch, the session should be restored with a valid expiresAt in the future, and the user should remain logged in.
Actual Behavior:
Session is restored, but expiresAt is always a timestamp from 1970, so the session is always considered expired.

PR #1447, opened 2026-06-22
docs(gotrue): clarify Session.expiresAt is in seconds

spydon:
## Description

Clarifies the doc comment for `Session.expiresAt` to state that the value is a Unix timestamp in **seconds**, and shows the correct conversion to `DateTime`.

## Context

While investigating #1202 ("Session expiresAt always corrupted after hot restart"), I verified the reported symptom is **not a library bug**:

- `Session.expiresAt` is derived from the access token's JWT `exp` claim (unix seconds), not from the persisted `expires_at` JSON value. `Session.fromJson` doesn't even read `expires_at`, so persistence/deserialization cannot corrupt it.
- Every internal consumer (`isExpired`, the auto-refresh tick, `supabase_client.dart`) correctly multiplies by 1000.
- The reporter's logged `1970-...` date decodes to a millisecond value that is exactly a valid 2026 **seconds** timestamp, i.e. their own code passed `expires_at` (seconds) into `DateTime.fromMillisecondsSinceEpoch` without `* 1000`.

The real underlying problem is an ergonomics footgun: the previous doc comment didn't state the unit. This PR fixes that.

No behavior change; existing `session_test.dart` tests pass.

Refs #1202

### DIFF
--- a/packages/gotrue/lib/src/types/session.dart
+++ b/packages/gotrue/lib/src/types/session.dart
@@ -67,8 +67,12 @@ class Session {
     };
   }
 
-  /// A timestamp of when the token will expire. Returned when a login is
-  /// confirmed.
+  /// The Unix timestamp, in **seconds**, of when the token will expire.
+  /// Returned when a login is confirmed.
+  ///
+  /// To convert this to a [DateTime], multiply by 1000 since
+  /// [DateTime.fromMillisecondsSinceEpoch] expects milliseconds:
+  /// `DateTime.fromMillisecondsSinceEpoch(expiresAt * 1000)`.
   late int? expiresAt = _expiresAt;
 
   int? get _expiresAt {


packages/gotrue/lib/src/types/session.dart
```diff
@@ -67,8 +67,12 @@ class Session {
...
 
-  /// A timestamp of when the token will expire. Returned when a login is
-  /// confirmed.
+  /// The Unix timestamp, in **seconds**, of when the token will expire.
+  /// Returned when a login is confirmed.
+  ///
+  /// To convert this to a [DateTime], multiply by 1000 since
+  /// [DateTime.fromMillisecondsSinceEpoch] expects milliseconds:
+  /// `DateTime.fromMillisecondsSinceEpoch(expiresAt * 1000)`.
   late int? expiresAt = _expiresAt;
```
grdsdev, line 76: maybe have this as a `DateTime` in v3?

grdsdev (approved):

packages/gotrue/lib/src/types/session.dart
spydon: Yeah, sounds like a good idea, I'll open an issue about that

PR #1450, opened 2026-06-22
fix(gotrue): avoid duplicate auth error on recoverSession with invalid refresh token

### LINKED ISSUE #930: Sentry reports AuthException
**Describe the bug**
Sentry reports a couple of AuthException's
All of the reports are on iOS(But those are like 99% of my users)

**To Reproduce**
I have not been able to reproduce this with one of my own devices.

**Version:**
supabase_flutter 2.5.2
supabase 2.1.2
functions_client 2.1.0
gotrue 2.6.1
postgrest 2.1.1
realtime_client 2.0.4
storage_client 2.0.1

**Additional context**
```
AuthException(message: Invalid Refresh Token: Refresh Token Not Found, statusCode: 400)
  App                 0x10bb1abd4  GoTrueClient.notifyException (gotrue_client.dart:1206)
  App                 0x10bd5c604  GoTrueClient.recoverSession (gotrue_client.dart:995)
  File "<asynchronous suspension>"
  App                 0x10bd5c32c  SupabaseAuth.recoverSession (supabase_auth.dart:90)
  File "<asynchronous suspension>"
  App                 0x10bb1aa3c  CancelableCompleter.complete.<T> (cancelable_operation.dart:427)
 ```

### PR DESCRIPTION
**Pull request by spydon**
## What

Fixes the long-standing reports of `AuthException(message: Invalid Refresh Token: Refresh Token Not Found, statusCode: 400)` surfacing as an uncaught error (often seen in Sentry).

Closes #930

## Why

When `recoverSession` restores a persisted session whose refresh token is invalid or expired, the refresh goes through `_callRefreshToken` -> `_executeRefresh`. For a non-retryable `AuthException` (like "Refresh Token Not Found"), `_executeRefresh` already does the right thing: it removes the session and emits a `signedOut` event, deliberately **not** calling `notifyException`.

The problem was that `recoverSession`'s blanket `catch` then called `notifyException(error, stackTrace)` a **second time**, pushing the `AuthException` onto the `onAuthStateChange` stream. For an expected condition (a rotated, expired, or revoked token at app startup), consumers listening to `onAuthStateChange` without an `onError` handler get an uncaught async error. That is what those Sentry reports were.

## How

Notification in `recoverSession` is now explicit per path instead of a catch-all re-notify:

- The refresh-token path defers entirely to `_callRefreshToken` / `_executeRefresh`, the single owner of that outcome (`signedOut` for an invalid token, an exception for a retryable failure). No more duplicate stream error.
- Genuinely unexpected parse failures (corrupt JSON) are still notified.
- The explicit "missing data" and "Session expired" (`autoRefreshToken: false`) notifications are preserved unchanged.

## Tests

- Updated `Sign out on wrong refresh token` to assert the invalid-token path surfaces **only** a `signedOut` event and never a stream error, while still throwing to the direct caller.
- Verified the unaffected `autoRefreshToken: false` path still emits its exception (`supabase_flutter` "emits exception when no auto refresh" still passes).

### CODE REVIEW

**Comment by Vinzent03** [comment]
Am I right that on a failed recovery there is only a signed out event without an exception on the stream? This removes the ability to have different handling or notification to the user why they got signed out. Also for future refreshes it still emits an exception to the stream on an issue, right? I don't think the solution to people not having a proper stream listener is to remove the errors.

**Comment by spydon** [comment]
> Am I right that on a failed recovery there is only a signed out event without an exception on the stream? 
> This removes the ability to have different handling or notification to the user why they got signed out. 

So there are two cases right:
- **Invalid/expired refresh token** (non-retryable) → signedOut event, no stream error. This sign-out (and the signedOut event) is already emitted by `_executeRefresh` today. This PR just removes the duplicate error that `recoverSession`'s catch was adding on top of it.
- Network / retryable / unexpected errors → still emitted as an exception on the stream, and the user is not signed out. So this is the same behavior as today.

> Also for future refreshes it still emits an exception to the stream on an issue, right? I don't think the solution to people not having a proper stream listener is to remove the errors.

Removing errors to paper over a missing `onError` would be the wrong fix indeed. The case here though is that we no longer emit an error in a successful, expected state transition: the refresh token is definitively invalid/expired, so the user gets signed out and a `signedOut` event fires. I.e. a normal lifecycle outcome, not an error condition. `_executeRefresh` already treats it that way (it signs out and intentionally doesn't `notifyException`). `recoverSession` was the odd one out, re-emitting that same outcome a second time as an error.

We should probably add a reason to the `signedOut` `AuthState` though, so that the user can differentiate on the sign out reason. But I think that should be a follow-up. What do you think?

**Comment by Vinzent03** [comment]
Removing the additional redundant exception on the stream is great. I took now a deeper look in the existing code and saw that we no longer emit an exception on refresh_token_already_used on timer based refreshes as well. I had the impression that we emit an exception there as well. This got changed in a fix by you:

https://github.com/supabase/supabase-flutter/pull/1351/changes#diff-d471e660d61f380c883733e18e96a1098d984fc6de28771f762b558b6e922872L1430-L1446

I didn't catch this in that pr. So yeah this aligns the behavior of the initial refresh. I think adding a reason to the sign out event that it was not initiated by the user itself would be great.

File: packages/gotrue/lib/src/gotrue_client.dart [hunk 1]
```diff
@@ -1146,38 +1147,46 @@ class GoTrueClient {
...
     } catch (error, stackTrace) {
       notifyException(error, stackTrace);
       rethrow;
     }
+
+    // Called outside the try/catch above on purpose: `_callRefreshToken`
+    // already notifies subscribers about the outcome (a `signedOut` event when
+    // the refresh token is invalid, or an exception for a retryable failure),
+    // so re-notifying the error here would push a duplicate onto
+    // `onAuthStateChange`, surfacing as an uncaught stream error for listeners
```
Vinzent03: This states that the duplicate exception surfaces as an uncaught stream error. But isn't this the case for the first one as well if there is no onError handler? So this just generally prevents a duplicate exception to the stream, but is not really relevant for an uncaught error.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -1146,38 +1147,46 @@ class GoTrueClient {
...
+    // without an `onError` handler.
+    return _callRefreshToken(refreshToken);
```
Vinzent03: IIRC futures need to be awaited here such that `recoverSession` properly appears in a stacktrace?

spydon:
> I think adding a reason to the sign out event that it was not initiated by the user itself would be great.

@Vinzent03 I opened a stacked PR implementing this here: https://github.com/supabase/supabase-flutter/pull/1453
I was considering just adding it as an enum, but it's probably easier if we just expose the exception.

Vinzent03:
I am currently thinking whether these changes and the one before is actually breaking? There is no analysis breaking change, but an application that depends on the auth exception being emitted would not work quite as before. Really unsure here.

packages/gotrue/lib/src/gotrue_client.dart
spydon: True! I simplified the comment in https://github.com/supabase/supabase-flutter/pull/1450/commits/4d08f78a5a95eb7bad3f2a1ff275399f812ff519, maybe I should just remove the comment altogether.

packages/gotrue/lib/src/gotrue_client.dart
spydon: True, good catch! Not the first time I fall into that trap. 😅 
Fixed here: https://github.com/supabase/supabase-flutter/pull/1450/commits/510262d4ab69055cfa26ea30a6babf6189241d0a

spydon: Ah, still fails the test, another fix is coming up.

spydon:
> I am currently thinking whether these changes and the one before is actually breaking? There is no analysis breaking change, but an application that depends on the auth exception being emitted would not work quite as before. Really unsure here.

With the one before, do you mean #1351? That is already released.

Yeah I guess this is technically breaking in the sense that it is a behavior change, but I don't think it deserves a major bump due to how strange/unintentional it was behaving before. We should surely make a note of it in the changelog though.

packages/gotrue/lib/src/gotrue_client.dart
spydon: Fixed in https://github.com/supabase/supabase-flutter/pull/1450/commits/cd68201cd522b971faf384f59c38fd1eeddfb6d6

Vinzent03: I wonder if there is a DCM rule for this?

spydon: Yeah, and normal linter rules too, I just haven't gotten to enabling them yet since it needed quite many changes. But I'll do that now. :)

spydon: Set up in https://github.com/supabase/supabase-flutter/pull/1454 and https://github.com/supabase/supabase-flutter/pull/1455

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -1146,38 +1147,50 @@ class GoTrueClient {
...
     }
+
+    // Run the refresh outside the try/catch above so its error is not
+    // re-notified: `_callRefreshToken` already reports the outcome (a
+    // `signedOut` event for an invalid token, or a stream exception for a
+    // retryable failure). The refresh runs fire-and-forget internally, so its
+    // error would otherwise be rooted at `_executeRefresh`; rethrowing with the
+    // current stack keeps `recoverSession` and the caller in the stack trace.
+    try {
+      return await _callRefreshToken(refreshToken);
```
Vinzent03, line 1175: Primarily to learn something for me here, just adding the `await` without the explicit error with stacktrace, does not work here?
And can't this just be `throw error` or even `rethrow` as the documentation of it states:
> Behaves like throw error would if the [StackTrace.current](https://api.flutter.dev/flutter/dart-core/StackTrace/current.html) was stackTrace at the time of the throw.
https://api.flutter.dev/flutter/dart-core/Error/throwWithStackTrace.html

packages/gotrue/lib/src/gotrue_client.dart
Vinzent03: This already emits an exception, but also throws a new one that is caught in line 1179 and re-emitted to the stream, doesn't it? The same issue appears in line 1175. So there are two errors in the stream.

packages/gotrue/lib/src/gotrue_client.dart
spydon: > Primarily to learn something for me here, just adding the await without the explicit error with stacktrace, does not work here?

Nope, then we'd lose these frames from the stacktrace unfortunately.

>  And can't this just be throw error or even rethrow as the documentation of it states:

Yup we could, but then we would have to remove `use_rethrow_when_possible` because `throw error` would trip that one, so with this long one we are quite clear with the intent, instead of removing the lint rule (that we most often want to apply) or adding an ignore that someone later maybe tries to fix by removing it and replacing with `rethrow` and than unconsciously removes the stacktrace from the user.

It's not that pretty, but I think it's the best we can do for now.

packages/gotrue/lib/src/gotrue_client.dart
spydon: True, good catch! Fixed in https://github.com/supabase/supabase-flutter/pull/1450/changes/d9eb033328014dc4beecc120a5f0189a15e996f8

packages/gotrue/lib/src/gotrue_client.dart
spydon: I thought of another way to do it, what do you think about this: https://github.com/supabase/supabase-flutter/pull/1457 ?

Vinzent03: (To your previous answer) Okay thanks for the explanation. Sorry to reiterate on this 😅: But this means the stacktrace only contains the `recoverSession` method and not where in `_callRefreshToken` the exception occurred, right?

Will take a look at the new pr.

spydon: Indeed, but the most important thing is that it'll contain the stacktrace frames that are from the user's code, which would otherwise be lost (and which is what most people care about).

Vinzent03 (approved):

Tr00d (approved):

PR #1451, opened 2026-06-22
fix(realtime_client): detect dropped connections on iOS via WebSocket ping

### LINKED ISSUE #1071: Inconsistent Stream Behavior on Internet Disconnect Between Android and iOS
**Describe the bug**
I am using the `supabase_flutter` and listening for user changes from the `users` table. When the internet connection is lost on Android, the stream throws the following error:

```dart
RealtimeSubscribeException(status: RealtimeSubscribeStatus.channelError, details: RealtimeCloseEvent(code: 1002, reason: ))
```

However, when the internet connection is lost on iOS devices, no error is thrown, and no event is emitted. This results in inconsistent behavior between platforms, making it difficult to maintain accurate and synchronized results.

**To Reproduce**
1. Listen any supabase table changes in the flutter app
2. Disconnect the internet on an Android device and observe the error.
3. Disconnect the internet on an iOS device and observe the lack of any response or error.

**Expected behavior**
On both Android and iOS, the stream should behave the same way when the internet connection is lost to ensure consistent behavior across platforms.

```shell
dart pub deps | grep -E "supabase|gotrue|postgrest|storage_client|realtime_client|functions_client"
│   ├── supabase_flutter...
│   ├── supabase_flutter...
│   ├── supabase_flutter...
│   ├── supabase_flutter...
│   ├── supabase_flutter...
│   ├── supabase_flutter...
│   ├── supabase_flutter...
│   └── supabase_flutter...
│   ├── supabase_flutter...
│   ├── supabase_flutter...
├── supabase_flutter 2.8.0
│   ├── supabase 2.5.0
│   │   ├── functions_client 2.4.0
│   │   ├── gotrue 2.10.0
│   │   ├── postgrest 2.3.0
│   │   ├── realtime_client 2.4.0
│   │   ├── storage_client 2.2.0
```

**Additional context**
Flutter version: 3.24.4
iOS device on testing: iPhone 14 Pro Max - iOS version: 18.0.1
Android device on testing: Xiaomi 13 Lite (Android 14)

spydon:
## What

Enable `pingInterval` on the native (`dart:io`) WebSocket so dropped connections are detected consistently across platforms.

## Why

Fixes #1071. When the internet drops, Android emits a `RealtimeSubscribeException` (close code 1002) but iOS emits nothing, leaving the stream silently stale.

**Root cause:** `createWebSocketClient` created the native WebSocket without a `pingInterval`, so a *silently* dropped connection was never detected at the socket layer:

- On **Android**, the OS surfaces the dead TCP connection promptly (RST → close code 1002), so `onDone`/`onError` fires and `_triggerChanError` propagates a `channelError`.
- On **iOS**, the OS buffers writes instead of surfacing a reset. The only fallback was the app-level Phoenix heartbeat, whose recovery path calls `conn.sink.close()`, which does not reliably fire `onDone` on a dead iOS socket. So nothing propagated to the channel.

## How

Set `pingInterval` (new `Constants.defaultWebSocketPingInterval`, 25s, aligned with the existing heartbeat cadence) on the native WebSocket. `dart:io` now actively probes the peer and closes the connection (`goingAway`) when a pong is not received, which flows through the same `onDone → _onConnClose → _triggerChanError → reconnect` path Android already uses.

## Scope / safety

- Only the default native transport changes. Web (`websocket_web.dart`) is untouched, since browsers manage ping/pong internally.
- Custom transports injected via `RealtimeClientOptions.transport` are unaffected.
- No public API change: the `WebSocketTransport` typedef is unchanged.
- A 25s pong window is generous for mobile round-trips.

## Testing

- `dart analyze` clean on the changed files.
- Realtime unit suite passes (`socket_test`, `mock_test`, `channel_test`, including the existing CHANNEL_ERROR-on-heartbeat-timeout test).
- Files formatted.

No unit test added: `pingInterval` is a property of the underlying `dart:io` socket and is not observable through `IOWebSocketChannel` without a live connection, and the codebase does not test the default transport (tests inject mocks).

### DIFF
--- a/packages/realtime_client/lib/src/constants.dart
+++ b/packages/realtime_client/lib/src/constants.dart
@@ -3,15 +3,6 @@ import 'package:realtime_client/src/version.dart';
 class Constants {
   static const Duration defaultTimeout = Duration(milliseconds: 10000);
   static const int defaultHeartbeatIntervalMs = 25000;
-
-  /// Interval for the native WebSocket to send protocol-level ping frames.
-  ///
-  /// When a pong is not received within this interval the connection is
-  /// assumed dead and closed, which makes silently dropped connections
-  /// (common on iOS, where the OS buffers writes instead of surfacing a TCP
-  /// reset) detectable and keeps disconnect behavior consistent across
-  /// platforms.
-  static const Duration defaultWebSocketPingInterval = Duration(seconds: 25);
   static const int wsCloseNormal = 1000;
   static const Map<String, String> defaultHeaders = {
     'X-Client-Info': 'realtime-dart/$version',
--- a/packages/realtime_client/lib/src/constants.dart
+++ b/packages/realtime_client/lib/src/constants.dart
@@ -3,6 +3,15 @@ import 'package:realtime_client/src/version.dart';
 class Constants {
   static const Duration defaultTimeout = Duration(milliseconds: 10000);
   static const int defaultHeartbeatIntervalMs = 25000;
+
+  /// Interval for the native WebSocket to send protocol-level ping frames.
+  ///
+  /// When a pong is not received within this interval the connection is
+  /// assumed dead and closed, which makes silently dropped connections
+  /// (common on iOS, where the OS buffers writes instead of surfacing a TCP
+  /// reset) detectable and keeps disconnect behavior consistent across
+  /// platforms.
+  static const Duration defaultWebSocketPingInterval = Duration(seconds: 25);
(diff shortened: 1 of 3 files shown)


Vinzent03:
Just as a sidenote, I detected some similar issues on web with reconnecting and especially channels rejoining behavior being inefficient. Changes are still local, but should be able to create pr this week.

grdsdev (approved):

PR #1453, opened 2026-06-22
feat(gotrue): surface the sign-out reason on the signedOut event

spydon:
## What

Adds `AuthState.signOutReason`, a typed `SignOutReason` populated on the `signedOut` event so listeners can tell **why** the user was signed out:

- `userInitiated` for an explicit `signOut()`.
- `sessionExpired` when the refresh token was rejected (invalid or expired) and the session was removed.
- `sessionMissing` when a stored session could not be recovered because it was missing required data.

It is `null` for every event other than `signedOut`, and also `null` for `signedOut` events received from another tab via `web.BroadcastChannel`.

## Why

Follow-up to #1450, which stops `recoverSession` from double-reporting an expected sign-out as an uncaught stream error. The remaining concern was that a pure `onAuthStateChange` listener then sees `signedOut` without knowing why the user was signed out, removing the ability to handle it differently or notify the user.

This restores that capability through the event itself, rather than through an error that only reaches listeners with an `onError` handler:

```dart
supabase.auth.onAuthStateChange.listen((state) {
  if (state.event == AuthChangeEvent.signedOut &&
      state.signOutReason != null &&
      state.signOutReason != SignOutReason.userInitiated) {
    // involuntary sign out (e.g. expired session) -> show a message
  }
});
```

## How

- New `SignOutReason` enum (`userInitiated`, `sessionExpired`, `sessionMissing`), exported from `gotrue.dart`.
- `AuthState` gains a nullable `signOutReason` field, threaded through `notifyAllSubscribers`.
- `signOut()` now delegates to a private `_signOut({scope, reason})` so the public API is unchanged while internal callers can pass a reason.
- `recoverSession` and `setInitialSession` sign out with the appropriate reason and now throw `AuthException`s tagged with `ErrorCode.sessionMissing` / `ErrorCode.sessionExpired` (new error codes).
- The non-retryable refresh failure path signs the user out with `sessionExpired`.

The reason is intentionally **not** propagated across tabs via the broadcast channel, so it is always `null` when `fromBroadcast` is `true`.

## Tests

- `signedOut` event carries `SignOutReason.sessionExpired` on an invalid refresh token.
- `signedOut` event reports `SignOutReason.userInitiated` on an explicit `signOut`.


Vinzent03:
Adding an `exception` field to the class feels a bit strange, if we simultaneously use errors on the stream itself. But I cannot come up with a better solution right now. I thought about an enum at first as well like you did here: https://github.com/supabase/supabase-flutter/pull/1450#issuecomment-4768962365, but that is probably hard to properly maintain and this solution is more general.
However, I think you've missed the sign out reasons in the recover method, right? They only have a human message and no special code to handle. Might be interesting to create a code that we use across similar issues in the recover method?

spydon:
@Vinzent03 I went with an enum anyways, I think it should be enough and it's much cleaner. What do you think?

packages/gotrue/lib/src/types/auth_state.dart
```diff
@@ -1,18 +1,35 @@
...
 class AuthState {
   final AuthChangeEvent event;
   final Session? session;
 
+  /// Why the user was signed out, when [event] is
+  /// [AuthChangeEvent.signedOut].
+  ///
+  /// Lets listeners tell an explicit [GoTrueClient.signOut] apart from an
+  /// involuntary sign out, such as an invalid or expired refresh token, without
+  /// having to attach an `onError` handler. It is `null` for every event other
```
Vinzent03: I would rephrase the part about the onError handler, because the user still needs to provide a handler for the other exceptions, which would otherwise be uncaught. So they don't need the onError handler for this use case specifically, but needs to be added nevertheless. I feel like this motivates to leave it out.

packages/gotrue/lib/src/types/sign_out_reason.dart
```diff
@@ -0,0 +1,17 @@
+/// The reason why an [AuthChangeEvent.signedOut] event was emitted.
+///
+/// Available on [AuthState.signOutReason] and lets listeners distinguish an
+/// explicit sign out from an involuntary one, such as an expired session,
+/// without inspecting error messages.
+enum SignOutReason {
```
Vinzent03, line 6: Maybe link to the respective error codes as well in the document comment of the error values, but unsure.

Vinzent03:
I like it! Should be much easier to handle for the user and is not too specific to properly maintain for us.

packages/gotrue/lib/src/types/sign_out_reason.dart
spydon: You mean just a dartdoc link like [this](https://github.com/supabase/supabase-flutter/pull/1453/changes/5f342756007acbb19a3f510eba59d692d1ba85c1#diff-38fedb0d42674c23e9278bf3d98f6998f466473ed34f6adefd172000d1fb1313R12-R19)?

packages/gotrue/lib/src/types/auth_state.dart
spydon: True, updated

grdsdev (approved):

packages/gotrue/lib/src/types/sign_out_reason.dart
Vinzent03: Yeah just like that. Just to link all connected parts together.

Vinzent03 (approved):

PR #1457, opened 2026-06-22
refactor(gotrue): simplify the token refresh flow

spydon:
> Draft, stacked on top of #1450 so the diff is isolated to the refactor. Base branch is `fix/recover-session-duplicate-auth-error`.

## Goal

Reduce the complexity of the `recoverSession` + `_callRefreshToken` + `_executeRefresh` + completer + error-emission combo, without any behavior change.

## What changed

- Split the old `_executeRefresh(refreshToken, completer)` into a clean `async` worker `_doRefresh(refreshToken)` that simply **returns** the `AuthResponse` or **throws**. All the scattered `if (!completer.isCompleted) completer.complete/completeError(...)` calls (6 of them, across every branch) are gone; the body now reads top-to-bottom.
- `_doRefresh` is the single place that emits refresh outcomes: `tokenRefreshed` on success, `signedOut` for an invalid token, or a stream error for a retryable/unexpected failure.
- `_callRefreshToken` just wires `_doRefresh` into the completer and cleans up the pending-map entry on completion.

## What deliberately stayed (and why)

**The `Completer` is kept on purpose.** I initially wanted to drop it and use the worker's future directly as the de-dup token, but `dispose()` must be able to cancel an **in-flight** refresh whose network request is hung (the gate never resolves). A plain `async` future can't be completed from the outside, so only an external completion handle can do this. This is covered by the existing `dispose completes active refresh with error` test, where the network gate is never completed and the future resolves solely because `dispose()` errors the completer. So the completer is load-bearing, not accidental.

The other essential parts are unchanged: the `_sessionVersion` guard, the dispose guard, and the dual notification model (events + stream errors).

`recoverSession` is untouched here, it still rethrows with the current stack to stay in the async stack trace (that mechanism is independent of how the refresh resolves).

## Behavior / non-breaking

No public API change. All refresh, race, dedup, dispose, stack-trace, and single-error tests pass (`token_refresh_race_test.dart` + `refresh_token_race_test.dart`), analyze + DCM clean.


Vinzent03:
Yeah this looks much cleaner. Really like it!

Vinzent03 (approved):

PR #1467, opened 2026-06-25
feat(realtime): update httpSend to per-event broadcast URL with binary support

spydon:
## What

Brings `RealtimeChannel.httpSend()` to parity with supabase-js ([supabase/supabase-js#2400](https://github.com/supabase/supabase-js/pull/2400)).

- Uses the new per-event URL format `POST /api/broadcast/{topic}/events/{event}` instead of the old `POST /api/broadcast` with a `messages` array wrapper.
- Adds `?private=true` query parameter for private channels.
- URL-encodes the topic and event names.
- Sends the payload directly (no `messages` wrapper).
- `TypedData` (e.g. `Uint8List`) and `ByteBuffer` payloads are sent as `application/octet-stream`; everything else is JSON-encoded.
- Returns a clearer error when the server responds with 404 (server too old).

The implicit REST fallback in `send()` is intentionally left on the old format, matching supabase-js, which only changed `httpSend()`.

## Compatibility

- The `payload` parameter type widens from `Map<String, dynamic>` to `Object` to allow binary payloads. This is backward compatible: existing calls passing a map still compile.
- The new endpoint requires Realtime server **v2.97.0+** ([supabase/realtime#1864](https://github.com/supabase/realtime/pull/1864)). This matches supabase-js, which shipped the same change as a minor `feat`.

## Test plan

- Updated `httpSend` tests for the new URL format and direct payload body.
- Added tests for the `private` query parameter (present for private, absent for public), URL-encoding of special characters, and `Uint8List`/`ByteBuffer` binary payloads.

Closes SDK-1018

### DIFF
--- a/packages/realtime_client/lib/src/realtime_channel.dart
+++ b/packages/realtime_client/lib/src/realtime_channel.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:typed_data';
 
 import 'package:http/http.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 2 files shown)


packages/realtime_client/lib/src/realtime_channel.dart
```diff
@@ -544,14 +550,29 @@ class RealtimeChannel {
   /// ```
   Future<void> httpSend({
     required String event,
-    required Map<String, dynamic> payload,
+    required Object payload,
```
Vinzent03, line 578: Is data like just a boolean or a number supported as well? They are valid json values, but not a typical key:value pair object. If they are not supported, I would propose adding an assertion here and adding to the documentation that it can either be a Map<String,dynamic> or binary.

spydon: Yeah indeed it can only be Map<String,dynamic> or binary, added assertion and clarified doc comment here: https://github.com/supabase/supabase-flutter/pull/1467/commits/aca55ca5a468ff2caf222a56a7a78c9f39824da3

Vinzent03 (approved):

grdsdev (approved):

PR #1476, opened 2026-06-26
fix(gotrue): copy request header/query maps before mutating them

AndroidPoet:
## What

`GotrueFetch.request()` now copies the request header and query maps before mutating them, so per-request values (`Authorization`, the API version header, `Content-Type`, `redirect_to`) no longer leak into the client's shared `_headers`.

## Why

`GoTrueClient` keeps a single `_headers` map and passes it **by reference** to almost every request (`headers: _headers`). `request()` aliased that map and wrote into it directly:

- `headers['Authorization'] = 'Bearer ${options.jwt}'` for any jwt-bearing call
- the `x-supabase-api-version` header
- `headers['Content-Type']` for non-GET requests (in `_handleRequest`)

So after any authenticated call — `getUser(jwt)`, `updateUser`, MFA, passkeys — the user's access token stayed in the client's default headers and was sent on every later request, including unauthenticated endpoints and requests made *after* `signOut`. Because requests interleave on the event loop, a no-jwt request could also pick up another in-flight request's token. The query map was aliased the same way through `redirect_to`.

## Not a breaking change

The copies are local to `request()`; the headers actually sent on each request are identical to before. The only thing that changes is that the shared map is no longer polluted — which is purely a fix.

## Tests

Added `packages/gotrue/test/header_isolation_test.dart`: after `getUser(jwt)` the token is present on the wire but absent from `client.headers`, and the header map is unchanged across repeated calls. Without the fix both assertions fail.

### DIFF
--- a/packages/gotrue/lib/src/fetch.dart
+++ b/packages/gotrue/lib/src/fetch.dart
@@ -120,7 +120,10 @@ class GotrueFetch {
     RequestMethodType method, {
     GotrueRequestOptions? options,
   }) async {
-    final headers = options?.headers ?? {};
+    // Copy the maps before mutating them. Callers pass the client's shared
+    // header/query maps by reference, so writing `Authorization`, the API
+    // version or `redirect_to` directly would leak into every later request.
+    final headers = {...?options?.headers};
 
     // Set the API version header if not already set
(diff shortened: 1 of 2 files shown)


spydon (approved):

PR #1477, opened 2026-06-26
fix(realtime): cancel pending reconnect on disconnect after a dropped socket

AndroidPoet:
## What

Fixes three related problems in `RealtimeClient` around a socket that dropped unexpectedly:

1. `disconnect()` now always cancels a pending reconnect, so disconnecting after the socket has already dropped no longer silently reopens the connection.
2. Automatic reconnects keep their exponential backoff instead of resetting to the shortest delay on every attempt.
3. A manual `connect()` after an unexpected drop now actually reconnects instead of being a no-op.

## Why

When the socket drops unexpectedly, `_onConnClose` sets `connState = closed` and schedules a reconnect through `reconnectTimer.scheduleTimeout()`. `conn` is not nulled on a drop, so a later `disconnect()` still enters the `conn != null` branch, but `oldState == closed` makes `shouldCloseSink` false, so the close block that cancelled the timer was skipped. `conn` was nulled and the method returned, yet the armed backoff timer still fired, ran `_reconnect()` and reopened a connection the user explicitly closed.

Cancelling the timer on every `disconnect()` fixes that, but it cannot use `reconnectTimer.reset()`: the reconnect path is `_reconnect -> disconnect -> connect`, so `reset()` would zero the retry counter on every attempt and defeat exponential backoff. During an outage the client would hammer an unreachable server at the shortest interval instead of backing off. A new `RetryTimer.cancel()` cancels the scheduled timer without touching the retry counter, so a user disconnect still stops a pending reconnect while automatic reconnects keep backing off. The counter is still zeroed on a successful open through `reset()` in `_onConnOpen`.

Finally, because `conn` is intentionally kept non-null after a drop so `conn.closeCode` and `conn.closeReason` stay readable, the `if (conn != null) return` guard in `connect()` treated the dead channel as a live connection and turned a manual `connect()` into a no-op. `connect()` now only bails out when the connection is live or in progress. When `conn` is set but `connState == closed`, it tears the stale connection down through `disconnect()` first, which cancels the old stream subscription, and then opens a fresh one.

## Not a breaking change

The public `conn` field is still left non-null after an unexpected drop, so consumers reading `conn.closeCode` or `conn.closeReason` are unaffected. `connState` transitions are unchanged: a socket that was already `closed` still ends up `closed`, and an open socket still ends up `disconnected` after `disconnect()`. `RetryTimer.reset()` keeps its existing behavior and now delegates its cancel step to the new `cancel()` method.

## Tests

Added three tests to the `disconnect` group:

- `cancels a pending reconnect after an unexpected drop`: drop the socket so a reconnect is scheduled, then `disconnect()` and assert the transport is never invoked again.
- `reconnects on a manual connect() after an unexpected drop`: drop the socket, call `connect()` manually, and assert a fresh connection is opened and `connState` becomes `open`.
- `grows the reconnect backoff across failed attempts`: fail the first few connection attempts and assert the retry counter passed to `reconnectAfterMs` grows as `1, 2, 3` instead of being reset to `1` on every reconnect.

Each test fails without its corresponding fix.


spydon (approved):

Vinzent03:
Seems like the format is not entirely correct at the moment.

What about also setting the `conn` to null at the end of `_onConnClose`?

spydon:
@AndroidPoet did you have a look at Vinzent's comment?

AndroidPoet:
> @AndroidPoet did you have a look at Vinzent's comment?
Thanks! Formatting's fixed 
 that is good point but right now disconnect() only cleans up when conn isn't null, so if I null conn on a drop it'd skip the reconnect-cancel and the bug comes back. Fix is easy: null conn in _onConnClose and move reconnectTimer.reset() so it always runs (plus a small test tweak). Want me to add it here or in a follow-up?

spydon:
@AndroidPoet I fixed it up, didn't know if you were available and I'm aiming on a release later this afternoon :)

AndroidPoet:
@spydon Awesome

spydon:
I just realized that the timer reset breaks the back-off flow, so fixing that up too.

AndroidPoet:
> I just realized that the timer reset breaks the back-off flow, so fixing that up too.

@spydon Yes, I always try to keep my PRs small. This one needed some extra verification, so thanks for handling that edge case.

PR #1478, opened 2026-06-26
fix(gotrue): handle access token without amr claim in getAuthenticatorAssuranceLevel

AndroidPoet:
## What

`GoTrueClient.mfa.getAuthenticatorAssuranceLevel()` no longer throws when the current session's access token has no `amr` claim — it returns an empty `currentAuthenticationMethods` list instead.

## Why

The method cast the claim unconditionally:

```dart
final amr = (payload['amr'] as List)
    .map((e) => AMREntry.fromJson(Map.from(e)))
    .toList();
```

When the access token carries no `amr` claim, `payload['amr']` is `null`, so `null as List` throws `type 'Null' is not a subtype of type 'List<dynamic>'`. This is a synchronous getter that callers don't expect to throw, and tokens without `amr` are easy to hit (custom/minimal JWTs, third-party signers). The fix defaults to an empty list:

```dart
final amr = (payload['amr'] as List? ?? [])
```

## Not a breaking change

When `amr` is present the behavior is identical. The only change is that an absent claim yields an empty list rather than a crash.

## Tests

Added `packages/gotrue/test/mfa_aal_test.dart`: recovers a session whose access token has `aal` but no `amr`, and asserts `getAuthenticatorAssuranceLevel()` returns `aal1` with empty methods. Without the fix the test crashes with the `Null is not a subtype of List` cast error.


packages/gotrue/test/mfa_aal_test.dart
Vinzent03, line 1: I would put this test into the existing `gotrue_mfa_api_test.dart` file as this is about mfa.

Vinzent03 (commented):
Thanks for the contribution. The fix makes sense as it is listed as an optional claim: https://supabase.com/docs/guides/auth/jwt-fields#optional-claims

packages/gotrue/test/mfa_aal_test.dart
AndroidPoet: Thanks for the review! that file is an integration test (loads .env, resets auth against a live server in setUp), while mine is a server-free unit test dropping it in as-is would make it require a running GoTrue.

Want me to move it there inside its own group(...) without the server setup, or keep it as a separate unit-test file?

spydon: Creating a new group in there sounds sensible

spydon: Fixed

spydon (approved):

PR #1481, opened 2026-06-26
fix(postgrest): escape quotes and backslashes in list filter values

AndroidPoet:
## What

List filter values are now backslash-escaped before being quoted, so values containing `"` or `\` produce well-formed PostgREST filters instead of malformed ones.

## Why

`_cleanFilterArray` quoted each non-numeric element without escaping:

```dart
return filter.map((s) => '"$s"').join(',');
```

So `inFilter('name', ['a"b'])` generated `in.("a"b")`. PostgREST treats the inner `"` as the end of the quoted element, so the filter is mis-parsed and the query silently returns wrong results (or errors). The same helper backs list `eq`/`neq`, `contains`/`containedBy`/`overlaps`, `not`, and rpc array arguments, so all of them are affected.

The fix escapes `\` and `"` within each element, matching PostgREST/PostgreSQL array-literal quoting:

```dart
final escaped = '$s'.replaceAll(r'\', r'\\').replaceAll('"', r'\"');
return '"$escaped"';
```

## Not a breaking change

Values without `"` or `\` are unaffected — the escaping is a no-op for them, so existing valid filters generate identical query strings. Only values that are currently broken change.

## Tests

Added `packages/postgrest/test/filter_escape_test.dart`: `inFilter('name', [r'a"b\c'])` with a capturing mock client, asserting the generated query is `in.("a\"b\\c")`. Without the fix it's the malformed `in.("a"b\c")`.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -410,7 +410,13 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
     if (filter.every((element) => element is num)) {
       return filter.map((s) => '$s').join(',');
     }
-    return filter.map((s) => '"$s"').join(',');
+    // Escape `\` and `"` inside each element before quoting, otherwise a value
+    // containing a double quote (e.g. `a"b`) produces a malformed PostgREST
+    // filter like `in.("a"b")`. This matches PostgREST/PostgreSQL array quoting.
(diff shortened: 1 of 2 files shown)


Vinzent03 (approved):
LGTM. Thanks for your contribution! Crazy this hasn't been caught earlier.

spydon (approved):
LGTM!

Issue #1475, opened 2026-06-26
Add OAuth Server 2.1 authorization methods to SDK


Pieter-JanRobrechtCronos:
### Description

As a developer using the Supabase SDK, I want to use the following methods to authorize an OAuth client.

> `supabase.auth.oauth.getAuthorizationDetails(authorization_id)` - Retrieve client and authorization details
> `supabase.auth.oauth.approveAuthorization(authorization_id)` - Approve the authorization request
> `supabase.auth.oauth.denyAuthorization(authorization_id)` - Deny the authorization request

These methods are available in the Javascript Library as described by [the Getting Started with OAuth 2.1 Server documentation](https://supabase.com/docs/guides/auth/oauth-server/getting-started?queryGroups=oauth-setup&oauth-setup=cli#configure-your-authorization-path).

### Suggested solution

In the GoTrue library we could provide a `GoTrueOAuthApi` that has similar attributes to the `GoTrueAdminMFAApi`. The new API class could then implement the following methods which each use `_fetch.request` to perform the necessary API calls.

```dart
Future<OAuthAuthorizationDetailsResponse> getAuthorizationDetails(String authorizationId);
Future<OAuthServerApprovalResponse> approveAuthorization(String authorizationId);
Future<OAuthServerDeniedResponse> denyAuthorization(String authorizationId);
```

For `getAuthorizationDetails`, this would be a GET call to `/auth/v1/oauth/authorizations/${authorization_id}`
For `approveAuthorization`, it would be a POST call to `/auth/v1/oauth/authorizations/${authorization_id}/consent` with a body `{ "action": "approve" }`.
For `denyAuthorization`, it would be a POST call to `/auth/v1/oauth/authorizations/${authorization_id}/consent` with a body `{ "action": "deny" }`.

The server responses would convert to a proper class with a `fromJson` method.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/supabase/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://supabase.com/docs/reference/dart/introduction).
- [x] Check that there isn't already an issue that requests the same feature to avoid creating a duplicate.

spydon:
Looks great @Pieter-JanRobrechtCronos! 
These two related ones are also missing:
```
- auth.oauth_server.list_grants — "List all active OAuth grants the current user has approved."
- auth.oauth_server.revoke_grant — "Revoke a specific OAuth grant that the current user previously approved."
```
[...]

PR #1483, opened 2026-06-27
feat(gotrue): add user-facing OAuth 2.1 server authorization API

gourabsingha1:
## Description

This PR implements the user-facing OAuth 2.1 server authorization consent flow methods on the `GoTrueClient` under the namespace `oauthServer` (via `client.oauthServer`). 

These methods are useful when your Supabase project acts as an OAuth 2.1 server and you are building a custom consent/authorization screen.

The following user-facing consent endpoints are exposed:
- `client.oauthServer.getAuthorizationDetails(authorizationId)`: Retrieve client details and requested scopes for a pending authorization request.
- `client.oauthServer.approveAuthorization(authorizationId, {skipBrowserRedirect})`: Approve the authorization request.
- `client.oauthServer.denyAuthorization(authorizationId, {skipBrowserRedirect})`: Deny the authorization request.

## Suggested Solution Context
Addresses: #1475

## Test Plan
Added a comprehensive suite of unit tests in `packages/gotrue/test/src/gotrue_oauth_server_api_test.dart` that validates request paths, HTTP methods, body payloads, query parameters, and JSON response deserialization for all three methods. All tests pass successfully.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -22,6 +22,7 @@ import 'broadcast_stub.dart' if (dart.library.js_interop) './broadcast_web.dart'
 import 'version.dart';
 
 part 'gotrue_mfa_api.dart';
+part 'gotrue_oauth_server_api.dart';
 part 'gotrue_passkey_api.dart';
 
 class _SessionState {
@@ -55,6 +56,15 @@ class GoTrueClient {
 
   /// Namespace for the GoTrue MFA API methods.
   late final GoTrueMFAApi mfa;
+
+  /// Namespace for the OAuth 2.1 authorization server consent flow.
+  ///
+  /// Use these methods when your Supabase project acts as an OAuth 2.1 server
+  /// and you are building a custom consent screen. Requires the OAuth 2.1
+  /// server feature to be enabled in your Supabase Auth configuration.
+  ///
+  /// {@macro gotrue_oauth_server_api}
+  late final GoTrueOAuthServerApi oauthServer;
 
   /// Namespace for the passkey (WebAuthn) API methods.
   ///
@@ -169,6 +179,7 @@ class GoTrueClient {
       httpClient: httpClient,
     );
     mfa = GoTrueMFAApi(client: this, fetch: _fetch);
+    oauthServer = GoTrueOAuthServerApi(client: this, fetch: _fetch);
     passkey = GoTruePasskeyApi(client: this, fetch: _fetch);
     if (_autoRefreshToken) {
       startAutoRefresh();
--- a/packages/gotrue/lib/src/gotrue_oauth_server_api.dart
+++ b/packages/gotrue/lib/src/gotrue_oauth_server_api.dart
@@ -1,0 +1,198 @@
+part of 'gotrue_client.dart';
+
+/// Response type representing the details of a pending OAuth authorization
+/// request.
+///
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthAuthorizationDetailsResponse {
+  /// The unique identifier for this authorization request.
+  final String authorizationId;
+
+  /// The OAuth client requesting authorization.
+  final OAuthClient client;
+
+  /// The scopes requested by the OAuth client, if any.
+  final String? scope;
+
+  /// The state parameter echoed from the authorization request, if present.
+  final String? state;
+
+  /// The redirect URI to be used after the authorization decision.
+  final String redirectUri;
+
+  const OAuthAuthorizationDetailsResponse({
+    required this.authorizationId,
+    required this.client,
+    this.scope,
+    this.state,
+    required this.redirectUri,
+  });
+
+  factory OAuthAuthorizationDetailsResponse.fromJson(
+      Map<String, dynamic> json) {
+    return OAuthAuthorizationDetailsResponse(
+      authorizationId: json['authorization_id'] as String,
+      client: OAuthClient.fromJson(json['client'] as Map<String, dynamic>),
+      scope: json['scope'] as String?,
+      state: json['state'] as String?,
+      redirectUri: json['redirect_uri'] as String,
+    );
+  }
+}
+
+/// Response type for an OAuth authorization consent decision (approve or deny).
+///
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthConsentResponse {
+  /// The URL to redirect the user to after the authorization decision.
+  ///
+  /// On approval this will contain the authorization code; on denial it will
+  /// carry an `access_denied` error — both in the form the OAuth client
+  /// registered to receive.
+  ///
+  /// This field is `null` if [skipBrowserRedirect] was `false` (the default)
+  /// and the SDK performed the redirect automatically.
+  final String? redirectUrl;
+
+  const OAuthConsentResponse({this.redirectUrl});
+
+  factory OAuthConsentResponse.fromJson(Map<String, dynamic> json) {
+    return OAuthConsentResponse(
+      redirectUrl: json['redirect_url'] as String?,
+    );
+  }
+}
+
+/// {@template gotrue_oauth_server_api}
+/// API namespace for the OAuth 2.1 authorization server consent flow.
+///
+/// Use these methods when your Supabase project is acting as an OAuth 2.1
+/// server **and** you are building a custom authorization / consent screen.
+/// The typical flow is:
+///
+/// 1. The third-party OAuth client redirects the user to your app's
+///    authorization path, which extracts the `authorization_id` from the
+///    query parameters.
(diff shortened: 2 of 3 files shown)


packages/gotrue/lib/src/gotrue_oauth_server_api.dart
```diff
@@ -0,0 +1,198 @@
...
+    this.scope,
+    this.state,
+    required this.redirectUri,
+  });
+
+  factory OAuthAuthorizationDetailsResponse.fromJson(
+      Map<String, dynamic> json) {
+    return OAuthAuthorizationDetailsResponse(
+      authorizationId: json['authorization_id'] as String,
+      client: OAuthClient.fromJson(json['client'] as Map<String, dynamic>),
```
spydon, line 35: This parses the client with `OAuthClient.fromJson`, which is the admin client type and expects `client_id`, `client_name`, `client_type`, `token_endpoint_auth_method`, `registration_type`, and so on as required fields.

The consent endpoint returns a different, smaller shape. From `internal/api/oauthserver/authorize.go` in `supabase/auth`:

```go
type ClientDetailsResponse struct {
    ID      string `json:"id"`
    Name    string `json:"name,omitempty"`
    URI     string `json:"uri,omitempty"`
    LogoURI string `json:"logo_uri,omitempty"`
}
```

So the JSON is `{id, name, uri, logo_uri}`, and `OAuthClient.fromJson` would throw on `json['client_id'] as String` (null) the first time this runs against a real server. supabase-js models this with a dedicated `OAuthAuthorizationClient` type holding exactly those four fields. Could we add a small response-specific client type here rather than reusing the admin one?

spydon (commented):
Hi @gourabsingha1, thank you so much for contributing to Supabase! :green_heart:

This is a genuinely useful addition, and we have an internal ticket for this too. The structure is clean, the endpoints and HTTP verbs line up with the auth server.

I cross-checked the response contracts against the `oauthserver` package in `supabase/auth` and the supabase-js reference, and there are a few things worth sorting out before we merge. I have left inline comments on the specific lines, but the headline items are:

1. The `client` field and the overall response shapes do not match what the server actually returns, so `getAuthorizationDetails` would throw against a real backend. The tests currently pass because the mock uses a response shape the server never sends.
2. `getAuthorizationDetails` needs to handle the already-consented case, where the server returns only a `redirect_url`.
3. `skipBrowserRedirect` does not have anything to act on in a non-browser package, and its doc comment describes behavior that does not happen here.
4. A small naming question about `oauthServer` versus `oauth` for parity with supabase-js.

None of this is a big lift, and the foundation you have laid is solid. Happy to help work through any of the points, so please do not hesitate to ask questions on the inline threads.

Thank you again for taking the time to build this. It is contributions like yours that make these SDKs what they are.

packages/gotrue/lib/src/gotrue_oauth_server_api.dart
```diff
@@ -0,0 +1,198 @@
...
+    required this.client,
+    this.scope,
+    this.state,
+    required this.redirectUri,
+  });
+
+  factory OAuthAuthorizationDetailsResponse.fromJson(
+      Map<String, dynamic> json) {
+    return OAuthAuthorizationDetailsResponse(
+      authorizationId: json['authorization_id'] as String,
```
spydon, line 34: `GET /oauth/authorizations/{id}` returns one of two shapes, not one. When the user has already consented (auto-approve), the server returns only `{ "redirect_url": "..." }`; when consent is needed it returns the full `{authorization_id, redirect_uri, client, user, scope}`.

Since `authorizationId` and `redirectUri` are parsed as required here, the already-consented response would throw. supabase-js returns a union (`OAuthAuthorizationDetails | OAuthRedirect`) for exactly this reason. Could `getAuthorizationDetails` return a type that models both variants? This is also the root of a few reports we have seen where the call fails for returning users.

packages/gotrue/lib/src/gotrue_oauth_server_api.dart
```diff
@@ -0,0 +1,198 @@
...
+  final String authorizationId;
+
+  /// The OAuth client requesting authorization.
+  final OAuthClient client;
+
+  /// The scopes requested by the OAuth client, if any.
+  final String? scope;
+
+  /// The state parameter echoed from the authorization request, if present.
+  final String? state;
```
spydon, line 18: The server's authorization details response does not include a `state` field (see `AuthorizationDetailsResponse` in `authorize.go`), so this will always be null. Safe to drop unless it is intentional future-proofing.

packages/gotrue/lib/src/gotrue_oauth_server_api.dart
```diff
@@ -0,0 +1,198 @@
...
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthConsentResponse {
+  /// The URL to redirect the user to after the authorization decision.
+  ///
+  /// On approval this will contain the authorization code; on denial it will
+  /// carry an `access_denied` error — both in the form the OAuth client
+  /// registered to receive.
+  ///
+  /// This field is `null` if [skipBrowserRedirect] was `false` (the default)
+  /// and the SDK performed the redirect automatically.
```
spydon, line 54: This doc describes the browser redirect behavior from supabase-js (`window.location.assign`), but this package never performs a redirect, so `redirectUrl` is always populated here. Could we update the comment so it does not promise a null that never happens?

packages/gotrue/lib/src/gotrue_oauth_server_api.dart
```diff
@@ -0,0 +1,198 @@
...
+
+    final data = await _fetch.request(
+      '${_client._url}/oauth/authorizations/$authorizationId/consent',
+      RequestMethodType.post,
+      options: GotrueRequestOptions(
+        headers: _client._headers,
+        jwt: session?.accessToken,
+        body: {
+          'action': 'approve',
+          if (skipBrowserRedirect) 'skip_browser_redirect': true,
```
spydon, line 158: Worth knowing: the consent endpoint's request body only reads `action` (the `ConsentRequest` struct in `authorize.go` has no `skip_browser_redirect` field), so the server ignores this. Combined with there being no browser to redirect in this package, `skipBrowserRedirect` is effectively a no-op today. It might be cleanest to drop the parameter for now, or keep it but document that it currently has no effect.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -56,6 +57,15 @@ class GoTrueClient {
...
   late final GoTrueMFAApi mfa;
 
+  /// Namespace for the OAuth 2.1 authorization server consent flow.
+  ///
+  /// Use these methods when your Supabase project acts as an OAuth 2.1 server
+  /// and you are building a custom consent screen. Requires the OAuth 2.1
+  /// server feature to be enabled in your Supabase Auth configuration.
+  ///
+  /// {@macro gotrue_oauth_server_api}
+  late final GoTrueOAuthServerApi oauthServer;
```
spydon, line 67: Small parity question: supabase-js exposes these as `auth.oauth` (and #1475 asks for `auth.oauth.getAuthorizationDetails`), whereas this is `auth.oauthServer`. Since the admin surface is already `auth.admin.oauth`, naming this `auth.oauth` would match the JS SDK and the issue. Was there a reason you preferred `oauthServer`?

packages/gotrue/test/src/gotrue_oauth_server_api_test.dart
```diff
@@ -0,0 +1,142 @@
...
+
+void main() {
+  const gotrueUrl = 'http://127.0.0.1:54421/auth/v1';
+
+  group('OAuth 2.1 Server User API', () {
+    test('getAuthorizationDetails parses response correctly', () async {
+      final mockResponse = {
+        'authorization_id': 'auth-id-123',
+        'client': {
+          'client_id': 'client-id-abc',
```
spydon, line 17: This mock returns the admin client shape (`client_id`, `client_name`, ...), but the server returns `{id, name, uri, logo_uri}` for this endpoint. That mismatch is why the test passes even though the parser would throw against the real server. Once the response types are corrected, it would be great to mock the actual server shape here, and to add a case for the already-consented `redirect_url` response.

packages/gotrue/lib/src/gotrue_oauth_server_api.dart
spydon: Following up on my own comment here: the consent-needed variant also includes a `user` object (`{id, email}`) that this PR does not model yet. The server returns it in `AuthorizationDetailsResponse`, and supabase-js exposes it on `OAuthAuthorizationDetails`, so the consent screen can show which account is about to grant access. Worth folding in when you restructure the response type.

spydon:
Since @Pieter-JanRobrechtCronos is assigned to this issue (and since this seemingly is a bit of AI slop...) I'll close this one. Please comment on larger issues that you want to take on before opening PRs.

PR #1486, opened 2026-06-28
fix(functions_client): use case-insensitive check for Content-Type header in invoke()

### LINKED ISSUE #1487: functions_client: custom Content-Type header in lower or mixed case is overwritten by default Content-Type
### Description

In `FunctionsClient.invoke()`, the SDK checks whether a `Content-Type` header is present before assigning a default content type (e.g. `application/json`). 

However, the check currently uses exact string matching:
```dart
if (body != null && (headers == null || !headers.containsKey("Content-Type")))
```

Per RFC 7230, HTTP header names are case-insensitive. When callers pass custom content-type headers with lowercase or mixed casing (such as `headers: {'content-type': 'application/custom+json'}`), `headers.containsKey("Content-Type")` evaluates to `false`.

As a result, `finalHeaders['Content-Type']` is set to the default content type, and when headers are merged onto the underlying HTTP request, the default value overwrites the user's intended custom header.

### Suggested Solution

Perform a case-insensitive check against `finalHeaders.keys` to verify if any case variant of `content-type` exists before applying default body content types.

gourabsingha1:
## Description

Closes #1487

Per RFC 7230, HTTP header names are case-insensitive. Previously, `FunctionsClient.invoke()` checked whether a `Content-Type` header was provided using exact string match: `!headers.containsKey("Content-Type")`.

If a caller passed a custom content-type header with lowercase or mixed casing (such as `headers: {'content-type': 'application/custom+json'}`), the check failed to detect it. Consequently, the default content-type for the body type was assigned to `finalHeaders['Content-Type']`, overwriting the caller's custom header when headers were set on the request.

## Fix

Switched the check to a case-insensitive lookup over `finalHeaders.keys` before setting default body content types.

## Test Plan

Added a unit test in `packages/functions_client/test/functions_dart_test.dart` verifying that passing `headers: {'content-type': 'application/custom+json'}` correctly overrides default headers. All 37 tests in the package pass.

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -117,8 +117,9 @@ class FunctionsClient {
         'x-region': effectiveRegion,
     };
 
-    if (body != null &&
-        (headers == null || !headers.containsKey("Content-Type"))) {
+    final hasContentType =
+        finalHeaders.keys.any((k) => k.toLowerCase() == 'content-type');
+    if (body != null && !hasContentType) {
       finalHeaders['Content-Type'] = switch (body) {
         Uint8List() => 'application/octet-stream',
         String() => 'text/plain',
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -117,9 +117,8 @@ class FunctionsClient {
         'x-region': effectiveRegion,
     };
 
-    final hasContentType =
-        finalHeaders.keys.any((k) => k.toLowerCase() == 'content-type');
-    if (body != null && !hasContentType) {
+    if (body != null &&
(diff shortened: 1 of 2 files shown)


packages/functions_client/lib/src/functions_client.dart
```diff
@@ -117,8 +117,9 @@ class FunctionsClient {
         'x-region': effectiveRegion,
     };
 
-    if (body != null &&
-        (headers == null || !headers.containsKey("Content-Type"))) {
+    final hasContentType =
+        finalHeaders.keys.any((k) => k.toLowerCase() == 'content-type');
```
spydon: This will now run even when the body is null, so it will be slightly less performant.

gourabsingha1: thanks for the comment. i've updated the implementation to inline the header checks back into the if block.

spydon (approved):

PR #1489, opened 2026-06-28
fix(gotrue)!: assert asyncStorage is provided for PKCE flow in the constructor

### LINKED ISSUE #1446: Assert asyncStorage is provided for PKCE flow in the GoTrue client constructor
Follow-up from #1349 (originally fixing #1319).

When `flowType` is `AuthFlowType.pkce` (the default) and no `asyncStorage` is provided, calling `getOAuthSignInUrl()` or `getLinkIdentityUrl()` fails late at call time. PR #1349 mitigated this by throwing a descriptive `AuthException` instead of crashing with a cryptic null check error in release mode.

A better long-term solution is to validate this requirement up front by adding an assert in the constructor of the GoTrue client, so the misconfiguration surfaces immediately rather than only when an affected method is called.

This is a breaking change and is therefore planned for v3.

Originally proposed in https://github.com/supabase/supabase-flutter/pull/1349#issuecomment-4263536161 by @Vinzent03.

gourabsingha1:
**Root cause**
When `flowType` is `AuthFlowType.pkce` (the default) and no `asyncStorage` is provided, calling `getOAuthSignInUrl()` or `getLinkIdentityUrl()` fails late at call time. To provide a better developer experience and fail fast, the misconfiguration should be validated up front in the constructor.

**Fix**
Added an `assert` in the `GoTrueClient` constructor to validate that `asyncStorage != null` if `flowType == AuthFlowType.pkce`.

**Tests**
- Added a unit test in `gotrue/test/client_test.dart` to verify that GoTrueClient throws an `AssertionError` if `asyncStorage` is missing when using PKCE flow.
- Updated all test suites in `gotrue` and `supabase` packages to pass a mock storage or specify `implicit` flow type where no storage was configured.

Fixes #1446

## Summary by CodeRabbit

* **New Features**
  * Added in-memory storage for PKCE code verifiers.
  * Supabase clients now provide in-memory PKCE storage by default for same-process authentication flows.
  * Added guidance for configuring persistent storage when PKCE values must survive app restarts.

* **Bug Fixes**
  * PKCE authentication now validates that required storage is configured.
  * OAuth sign-in URLs correctly include a PKCE code challenge.

* **Tests**
  * Expanded coverage for PKCE setup and in-memory storage behavior.

### DIFF
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -46,9 +46,11 @@ import 'trace_http_client.dart';
 ///
 /// The pkce flow is used by default and keeps its code verifiers in the
 /// `AuthAsyncStorage` passed to the `pkceAsyncStorage` field of [authOptions].
-/// Pass a `MemoryAuthAsyncStorage` when the flow starts and completes within
-/// the same process, or a persistent implementation when the code is exchanged
-/// after a restart.
+/// Pass a persistent implementation whenever the flow can leave the process
+/// before the code comes back, which covers every email link and every
+/// redirect to an OAuth provider. `MemoryAuthAsyncStorage` only suits flows
+/// that start and complete in the same process, such as tests and command line
+/// tools that keep a redirect listener open.
 /// {@endtemplate}
 class SupabaseClient {
   final String _supabaseKey;
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -44,9 +44,9 @@ import 'trace_http_client.dart';
 /// Pass an instance of `YAJsonIsolate` to [isolate] to use your own persisted
 /// isolate instance. A new instance will be created if [isolate] is omitted.
 ///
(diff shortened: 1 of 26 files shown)


Vinzent03:
As stated in your linked issue this is considered as a breaking change so won't plan this for main yet but later for v3 when we start actively working on it.

spydon (approved):

Vinzent03:
I'm not entirely sure we want to set a memory only storage option by default for the supabase package. I think this may hide issues developers may face of missing pkce values after a program restart or creating a new client. I would favor having no default, but still include the memory storage for the user to specify manually. This ensures they are more aware of that restriction. It also makes the new assert in the constructor basically redundant as it would only occur for users of the gotrue package only. I'm curious what you @spydon think of this. 

If we keep it as the default it think it would make more sense to set the memory storage as the storage option in the gotrue package itself rather than in the supabase package.

spydon:
@Vinzent03 yeah true, it's a bit more ergonomic for tests etc, but the footgun that it comes with is a bit too risky.
I'll refactor it so that there isn't a default.

Vinzent03 (approved):

PR #1490, opened 2026-06-29
fix(storage): preserve sortBy defaults when list receives a partial sortBy

spydon:
## What

`storage.from(bucket).list()` rejected a partially-specified `SortBy` with a 400.

`SortBy`'s constructor had no default for `column`/`order`, so `SortBy(column: 'updated_at')` left `order` as `null`. Since `SearchOptions.toMap()` always emits both keys and the storage client serializes with `json.encode` (which keeps nulls), the request body sent:

```json
"sortBy": { "column": "updated_at", "order": null }
```

The Storage server requires both `column` and `order`, so it returned a 400.

This mirrors supabase-js [#2454](https://github.com/supabase/supabase-js/pull/2454) (commit `42b7bbc`), which deep-merges the partial `sortBy` onto the defaults.

## Fix

Give `SortBy` constructor defaults so any partially-specified `SortBy` fills in the missing field, which is the idiomatic Dart equivalent of the JS deep-merge and works wherever `SortBy` is constructed:

```dart
const SortBy({this.column = 'name', this.order = 'asc'});
```

Behavior:

- `SortBy(column: 'updated_at')` → `{column: 'updated_at', order: 'asc'}`
- `SortBy(order: 'desc')` → `{column: 'name', order: 'desc'}`
- `list()` with no options → `{column: 'name', order: 'asc'}` (unchanged)

No breaking change for callers passing a complete `SortBy`.

## Tests

Added a `list sortBy defaults` group in `client_test.dart` (using the in-memory `CustomHttpClient` to inspect the request body) covering partial column, partial order, no options, and complete sortBy.

Closes SDK-1089

### DIFF
--- a/packages/storage_client/lib/src/types.dart
+++ b/packages/storage_client/lib/src/types.dart
@@ -236,7 +236,7 @@ class SortBy {
   final String? column;
   final String? order;
 
-  const SortBy({this.column, this.order});
+  const SortBy({this.column = 'name', this.order = 'asc'});
 
   Map<String, dynamic> toMap() {
     return {
--- a/packages/storage_client/lib/src/types.dart
+++ b/packages/storage_client/lib/src/types.dart
@@ -215,10 +215,7 @@ class SearchOptions {
   const SearchOptions({
     this.limit = 100,
     this.offset = 0,
-    this.sortBy = const SortBy(
-      column: 'name',
-      order: 'asc',
-    ),
+    this.sortBy = const SortBy(),
     this.search,
   });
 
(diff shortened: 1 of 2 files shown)


grdsdev (approved):

packages/storage_client/lib/src/types.dart
```diff
@@ -236,7 +236,7 @@ class SortBy {
   final String? column;
   final String? order;
 
-  const SortBy({this.column, this.order});
+  const SortBy({this.column = 'name', this.order = 'asc'});
```
spydon, line 236: In V3 we should explore if we can avoid to have order as a string, and convert it to an enum instead.

grdsdev: Yes, I did this in Swift v3 along with a few other types, check at https://github.com/supabase/supabase-swift/pull/987

spydon: Nice!

PR #1499, opened 2026-06-30
feat(gotrue): add user-facing OAuth 2.1 server authorization API

### LINKED ISSUE #1475: Add OAuth Server 2.1 authorization methods to SDK

Pieter-JanRobrechtCronos:

Adds the user-facing OAuth 2.1 server authorization consent flow methods:
- `getAuthorizationDetails(authorizationId)` (GET `/oauth/authorizations/{id}`)
- `approveAuthorization(authorizationId)` (POST `/oauth/authorizations/{id}/consent` with action: 'approve')
- `denyAuthorization(authorizationId)` (POST `/oauth/authorizations/{id}/consent` with action: 'deny')

These are exposed on GoTrueClient via `client.oauth`.

Closes #1475

## What is the new behavior?

## Additional context

image

### DIFF
--- a/packages/gotrue/lib/src/gotrue_oauth_api.dart
+++ b/packages/gotrue/lib/src/gotrue_oauth_api.dart
@@ -35,7 +35,7 @@ class OAuthAuthorizationDetailsResponse {
   final OAuthAuthorizedClient client;
 
   /// The OAuth User requesting authorization.
-  final User? user;
+  final User user;
 
   /// The scopes requested by the OAuth client, if any.
   final String? scope;
@@ -47,17 +47,27 @@ class OAuthAuthorizationDetailsResponse {
     required this.authorizationId,
     required this.client,
     required this.redirectUri,
-    this.user,
+    required this.user,
     this.scope,
   });
 
   factory OAuthAuthorizationDetailsResponse.fromJson(
     Map<String, dynamic> json,
   ) {
+    final user = json['user'] == null ? null : User.fromJson(json['user']);
+
+    if (user == null) {
+      throw ArgumentError.value(
+        json,
+        'json',
+        'The provided JSON should contain a parseable user object',
+      );
+    }
+
     return OAuthAuthorizationDetailsResponse(
       authorizationId: json['authorization_id'] as String,
       client: OAuthAuthorizedClient.fromJson(json['client']),
-      user: User.fromJson(json['user']),
+      user: user,
       scope: json['scope'] as String?,
       redirectUri: json['redirect_uri'] as String,
     );
--- a/packages/gotrue/lib/src/gotrue_oauth_api.dart
+++ b/packages/gotrue/lib/src/gotrue_oauth_api.dart
@@ -1,0 +1,203 @@
+part of 'gotrue_client.dart';
+
+/// OAuth client object returned from the OAuth 2.1 server when the client is
+/// authorized.
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthAuthorizedClient {
+  /// Unique identifier for the OAuth client
+  final String clientId;
+
+  /// Human-readable name of the OAuth client
+  final String? clientName;
+
+  const OAuthAuthorizedClient({
+    required this.clientId,
+    this.clientName,
+  });
+
+  factory OAuthAuthorizedClient.fromJson(Map<String, dynamic> json) {
+    return OAuthAuthorizedClient(
+      clientId: json['id'] as String,
+      clientName: json['name'] as String?,
+    );
+  }
+}
+
+/// Response type representing the details of a pending OAuth authorization
+/// request.
+///
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthAuthorizationDetailsResponse {
+  /// The unique identifier for this authorization request.
+  final String authorizationId;
+
+  /// The OAuth client requesting authorization.
+  final OAuthAuthorizedClient client;
+
+  /// The OAuth User requesting authorization.
+  final User? user;
+
+  /// The scopes requested by the OAuth client, if any.
+  final String? scope;
+
+  /// The redirect URI to be used after the authorization decision.
+  final String redirectUri;
+
+  const OAuthAuthorizationDetailsResponse({
+    required this.authorizationId,
+    required this.client,
+    required this.redirectUri,
+    this.user,
+    this.scope,
+  });
+
+  factory OAuthAuthorizationDetailsResponse.fromJson(
+    Map<String, dynamic> json,
+  ) {
+    return OAuthAuthorizationDetailsResponse(
+      authorizationId: json['authorization_id'] as String,
+      client: OAuthAuthorizedClient.fromJson(json['client']),
+      user: User.fromJson(json['user']),
+      scope: json['scope'] as String?,
+      redirectUri: json['redirect_uri'] as String,
+    );
+  }
+}
+
+/// Response type for an OAuth authorization consent decision (approve or deny).
+///
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthConsentResponse {
+  /// The URL to redirect the user to after the authorization decision.
+  ///
+  /// On approval this will contain the authorization code; on denial it will
+  /// carry an `access_denied` error — both in the form the OAuth client
+  /// registered to receive.
+  ///
+  /// This field is `null` if [skipBrowserRedirect] was `false` (the default)
+  /// and the SDK performed the redirect automatically.
+  final String? redirectUrl;
+
+  const OAuthConsentResponse({this.redirectUrl});
+
+  factory OAuthConsentResponse.fromJson(Map<String, dynamic> json) {
+    return OAuthConsentResponse(
+      redirectUrl: json['redirect_url'] as String?,
+    );
+  }
+}
+
+/// {@template gotrue_oauth_server_api}
+/// API namespace for the OAuth 2.1 authorization server consent flow.
+///
+/// Use these methods when your Supabase project is acting as an OAuth 2.1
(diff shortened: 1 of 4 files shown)


packages/gotrue/lib/src/gotrue_oauth_api.dart
```diff
@@ -0,0 +1,207 @@
+part of 'gotrue_client.dart';
+
+/// OAuth client object returned from the OAuth 2.1 server when the client is
+/// authorized.
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthAuthorizedClient {
```
Pieter-JanRobrechtCronos, line 7: Initially I wanted to reuse the `OauthClient` class. However, on closer inspection, I saw that the returned JSON from the API differs quite a lot compared rendering the use of `OauthClient.fromJson` useless.

`OauthClient.fromJson` checks the `client_id` and `client_name` fields, but the API doesn't return these values. Since both differ quite a bit, I've made the decision to introduce this new type solely for these methods.

This is an example JSON returned from the API
```json
{
  "authorization_id": "3izyptchi5y2e3mr2irauxj3khmzbddr",
  "redirect_uri": "http://localhost:50300/callback",
  "client": {
    "id": "212caa0f-3060-49f7-b7e9-d2fed05438eb",
    "name": "Local Dev App"
  },
  "user": {
    "id": "1bee2038-51fe-4f93-8fbb-442df18657ff",
    "email": "random.user@mail.com"
  },
  "scope": "email"
}
```

spydon: Makes sense I think, the `OauthClient` as a data class would've become quite bloated otherwise.

packages/gotrue/lib/src/gotrue_oauth_api.dart
```diff
@@ -0,0 +1,203 @@
...
+    this.scope,
+  });
+
+  factory OAuthAuthorizationDetailsResponse.fromJson(
+    Map<String, dynamic> json,
+  ) {
+    return OAuthAuthorizationDetailsResponse(
+      authorizationId: json['authorization_id'] as String,
+      client: OAuthAuthorizedClient.fromJson(json['client']),
+      user: User.fromJson(json['user']),
```
spydon: `User.fromJson` takes a non-nullable map and dereferences `json['id']` immediately, so this throws when the `user` key is absent, even though the field is `User?`. Guard it: `user: json['user'] == null ? null : User.fromJson(json['user'])`.

packages/gotrue/lib/src/gotrue_oauth_api.dart
```diff
@@ -0,0 +1,203 @@
...
+    required this.redirectUri,
+    this.user,
+    this.scope,
+  });
+
+  factory OAuthAuthorizationDetailsResponse.fromJson(
+    Map<String, dynamic> json,
+  ) {
+    return OAuthAuthorizationDetailsResponse(
+      authorizationId: json['authorization_id'] as String,
```
spydon, line 69: The docs below say the response may contain "only a redirect_url field", but this unconditionally casts `json['authorization_id']`/`json['redirect_uri']` to `String`, throwing a `TypeError` on that shape. Either model that response or drop the doc claim.

packages/gotrue/lib/src/gotrue_oauth_api.dart
```diff
@@ -0,0 +1,203 @@
...
+///
+/// Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+class OAuthConsentResponse {
+  /// The URL to redirect the user to after the authorization decision.
+  ///
+  /// On approval this will contain the authorization code; on denial it will
+  /// carry an `access_denied` error — both in the form the OAuth client
+  /// registered to receive.
+  ///
+  /// This field is `null` if [skipBrowserRedirect] was `false` (the default)
```
spydon: There is no `skipBrowserRedirect` parameter on these methods and no auto-redirect behaviour, so this note (and `[skipBrowserRedirect]` as a doc link) is misleading. `redirectUrl` is just whatever the server returns; please remove the reference (looks copied from the JS SDK).

packages/gotrue/lib/src/gotrue_oauth_api.dart
```diff
@@ -0,0 +1,203 @@
...
+///    details and the requested scopes for display in the consent UI.
+/// 3. Call [approveAuthorization] or [denyAuthorization] according to the
+///    user's decision.
+///
+/// ```dart
+/// // 1. Extract the authorization_id from the incoming redirect URL.
+/// final authorizationId = Uri.parse(currentUrl).queryParameters['authorization_id']!;
+///
+/// // 2. Show the consent screen.
+/// final details = await supabase.auth.oauthServer.getAuthorizationDetails(authorizationId);
```
spydon: The namespace is `oauth`, not `oauthServer` (see `gotrue_client.dart`), so this example won't compile. Applies to line 114 as well.

packages/gotrue/lib/src/gotrue_oauth_api.dart
Pieter-JanRobrechtCronos: Yes that's very true, I'll add the guard! 

I'm wondering what your view on this is. I made the `User` attribute nullable since the `User.fromJson` method can return null (when the `id` key is missing). However, I would argue that `OAuthAuthorizationDetailsResponse` should be tied to the user that is actually requesting the authorization. How do you usually handle this case? Would you keep the `User` attribute as nullable which possible could return a response that is effectively invalid? Or would you make the `User` attribute null safe?

packages/gotrue/lib/src/gotrue_oauth_api.dart
Pieter-JanRobrechtCronos: My bad, I'll remove part of these docs since they don't contain relevant information.

packages/gotrue/lib/src/gotrue_oauth_api.dart
Pieter-JanRobrechtCronos: I see that an automated check is failing due to the public API. Is this because I'm using the wrong namespace? The `sdk-compliance.yaml` contains `oauth_server`, is this the namespace that I should be using?

packages/gotrue/lib/src/gotrue_oauth_api.dart
spydon: Hmm, maybe we should just throw an exception if we don't get a user and then keep it as non-nullable?

packages/gotrue/lib/src/gotrue_oauth_api.dart
spydon: Nah, it's not because you're using the wrong namespace, the compliance check fails because the new API methods aren't covered. I can fix that up. :)
So the only thing that needs fixing here is the dartdoc comment.

spydon: Opened this stacked PR that you can pull in: https://github.com/Unlock-d/supabase-flutter/pull/1

packages/gotrue/lib/src/gotrue_oauth_api.dart
Pieter-JanRobrechtCronos: Yeah, that's what I've done right now. The `fromJson` will throw an `ArgumentError` when a null `User` object is returned during parsing.

spydon (approved):

packages/gotrue/lib/src/gotrue_oauth_api.dart
spydon: Updated to `FormatException`, to be consistent with the rest of the codebase.

Pieter-JanRobrechtCronos:
Thanks for your time as well! Do you know what the usual ETA is for this to land in a new release?

spydon:
> Thanks for your time as well! Do you know what the usual ETA is for this to land in a new release?

Should be up in ~10 minutes :)

PR #1502, opened 2026-07-01
fix(supabase_flutter): backfill requireResidentKey when omitted by the server

### LINKED ISSUE #1496: requireResidentKey causes null cast error when server omits the field
### Describe the bug

When calling supabase.auth.registerPasskey(), the app crashes because the Supabase server does not always include `requireResidentKey` in the WebAuthn response. The field is defined as non-nullable `bool` in passkeys_platform_interface, but the WebAuthn spec treats it as optional and deprecated since Level 2.

### Library affected

supabase_flutter

### Reproduction

_No response_

### Steps to reproduce

1. Enable Passkeys in Supabase Dashboard under Authentication > Configuration > Passkeys
2. Call supabase.auth.registerPasskey() on Android
3. App crashes with null cast error

### System Info

```shell
flutter --version output:
- supabase_flutter: 2.15.1
- passkeys_platform_interface: 2.7.0
```

### Platform

Windows, Android

### Logs

<details>
<summary>Click to expand!</summary>

```shell
Error: The argument type 'bool?' can't be assigned to the parameter type 'bool'.
        requireResidentKey: a.requireResidentKey,
                              ^
../../../../AppData/Local/Pub/Cache/hosted/pub.dev/passkeys_android-2.12.1/lib/passkeys_android.dart:77:31

Error: The argument type 'bool?' can't be assigned to the parameter type 'bool'.
        requireResidentKey: requestAuthSelectionType.requireResidentKey,
                                                     ^
../../../../AppData/Local/Pub/Cache/hosted/pub.dev/passkeys_windows-0.1.2/lib/passkeys_windows.dart:79:54
```
</details>

### Validations

- [x] Follow our [Code of Conduct](https://github.com/supabase/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://supabase.com/docs/reference/dart/introduction).
- [x] Check that there isn't [already an issue](https://github.com/supabase/supabase-flutter/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a Supabase Flutter/Dart Library issue and not an issue with the Supabase platform. If it's a Supabase platform related bug, it should likely be reported to [supabase/supabase](https://github.com/supabase/supabase) instead.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/supabase/supabase-flutter/discussions) or join our [Discord Chat Server](https://discord.supabase.com/).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

spydon:
## Summary
- `authenticatorSelection.requireResidentKey` has been optional and deprecated in the WebAuthn spec since Level 2, so the Supabase Auth server may correctly omit it. The `passkeys_platform_interface` plugin still requires it as a non-nullable `bool`, so passing the raw server response through crashed `registerPasskey()` with a null cast error.
- `passkeyRegisterRequestFromOptions` now normalizes `authenticatorSelection`, deriving `requireResidentKey` from `residentKey` when the server omits it (`true` if `residentKey == "required"`, otherwise `false`), matching the WebAuthn spec's backwards compatibility guidance.

Fixes #1496

## Test plan
- [x] Added unit tests in `passkey_options_mapper_test.dart` covering: default to `false`, derive `true` from a required `residentKey`, and preserving a provided value.
- [x] `flutter test test/passkey_options_mapper_test.dart` passes (10/10).
- [x] `flutter analyze` clean on changed files.

### DIFF
--- a/packages/supabase_flutter/lib/src/passkey/passkey_options_mapper.dart
+++ b/packages/supabase_flutter/lib/src/passkey/passkey_options_mapper.dart
@@ -5,10 +5,13 @@ import 'package:passkeys_platform_interface/types/types.dart';
 ///
 /// The Supabase API returns options in the W3C
 /// `PublicKeyCredentialCreationOptionsJSON` format, which already uses the same
-/// field names as [RegisterRequestType.fromJson]. The only adjustments needed
-/// are stripping base64url padding from the challenge and credential ids (the
-/// plugin rejects padded values) and ensuring every excluded credential carries
-/// a `transports` list (the plugin requires it).
+/// field names as [RegisterRequestType.fromJson]. The adjustments needed are
+/// stripping base64url padding from the challenge and credential ids (the
+/// plugin rejects padded values), ensuring every excluded credential carries
+/// a `transports` list (the plugin requires it), and backfilling
+/// `authenticatorSelection.requireResidentKey` (the plugin requires it, but
+/// the WebAuthn spec has treated it as optional since Level 2, so the server
+/// may omit it).
 RegisterRequestType passkeyRegisterRequestFromOptions(
   Map<String, dynamic> options,
 ) {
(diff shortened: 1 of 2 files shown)


Vinzent03:
Shouldn't this be fixed in the passkey package?

spydon:
> Shouldn't this be fixed in the passkey package?

Yeah, we should, and I have an open PR for it [here](https://github.com/corbado/flutter-passkeys/pull/254).
But I think it will be a while before we get a new release there so I thought we can do it here first and then we remove this once we have a new version of the passkeys package.

Tr00d (approved):

PR #1504, opened 2026-07-01
fix(postgrest): avoid leading comma in Prefer header for modifier+select

AndroidPoet:
## What

`insert()`, `update()` and `delete()` seed the request's `Prefer` header with an empty string. When a `.select()` is chained onto one of them, `PostgrestTransformBuilder.select()` appended `return=representation` onto that empty seed, producing a malformed header with a leading comma:

```
Prefer: ,return=representation
```

This forwards a stray empty preference token to PostgREST on every `insert().select()`, `update().select()` and `delete().select()` call. This changes `select()` to skip empty `Prefer` values, so only `return=representation` is sent.

## Why

`insert()`/`update()`/`delete()` set `Prefer` to `''` (query builder), then `select()` did:

```dart
final prefer = newHeaders['Prefer']; // '' — not null
newHeaders['Prefer'] = [
  if (prefer != null) prefer,          // '' survives the null check
  'return=representation',
].join(',');                            // => ',return=representation'
```

`supabase-js` avoids this because it guards with `if (this.headers['Prefer'])`, where the empty string is falsy. The Dart port only checked for `null`, so the empty seed leaked through. The fix guards with `prefer != null && prefer.isNotEmpty`, matching the reference behaviour.

## Not a breaking change

The header value is now the intended `return=representation` (and, for `upsert().select()`, still keeps `resolution=...,return=representation`). Only the erroneous leading comma is removed; no public API changes.

## Tests

Added `packages/postgrest/test/prefer_header_test.dart` covering `insert/update/delete().select()` (each asserts an exact `Prefer: return=representation`) and `upsert().select()` (asserts no leading comma and that existing preferences are preserved). The tests fail on the old code (`Actual: ',return=representation'`) and pass with the fix. `dart format` and `dart analyze --fatal-warnings` are clean.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_transform_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_transform_builder.dart
@@ -46,7 +46,10 @@ class PostgrestTransformBuilder<T> extends RawPostgrestBuilder<T, T, T> {
     final url = overrideSearchParams('select', cleanedColumns);
     final prefer = newHeaders['Prefer'];
     newHeaders['Prefer'] = [
-      if (prefer != null) prefer,
+      // `insert()`, `update()` and `delete()` seed `Prefer` with an empty
+      // string. Skipping empty values avoids emitting a malformed header with
+      // a leading comma (e.g. `,return=representation`).
+      if (prefer != null && prefer.isNotEmpty) prefer,
       'return=representation',
     ].join(',');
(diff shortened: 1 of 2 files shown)


spydon:
Superseded by https://github.com/supabase/supabase-flutter/pull/1505, good find though!

PR #1507, opened 2026-07-01
fix(functions): don't lose error status on a non-JSON body labeled JSON

AndroidPoet:
## What

`FunctionsClient.invoke()` decoded the response body as JSON whenever the `Content-Type` was `application/json`. If a response was labeled JSON but carried a non-JSON body — e.g. an upstream proxy returning an HTML error page (`500` + `<html>…502 Bad Gateway…</html>` with `Content-Type: application/json`) — `_isolate.decode` threw a `FormatException` straight out of `invoke()`. The caller never got a `FunctionException`, and the HTTP status was lost.

On an error status this now falls back to the raw text when the body isn't valid JSON, so the status is still reported. It also compares the media type case-insensitively.

## Why

Two issues in the same block:

1. The `application/json` branch decoded unconditionally, so a mislabeled non-JSON body turned a `500` into an uncaught `FormatException` instead of a `FunctionException` with `status: 500`. This is the same spirit as #1480 (drain the error body into `details` rather than letting it escape).
2. The media-type match was case-sensitive (`responseType == 'application/json'`), so `application/JSON` fell through to the raw-text branch and JSON was handed back as an undecoded `String`. Media types are case-insensitive per RFC 9110.

Fix: wrap the decode in a `try`/`on FormatException` fallback, and `.toLowerCase()` the media type. The fallback to raw text is scoped to error statuses only: on a `2xx` the `application/json` label is a promise of structured data, so a body that doesn't parse is a real anomaly and the `FormatException` is rethrown rather than silently degrading to a `String`. The narrow `FormatException` catch also preserves existing behavior (e.g. a disposed isolate still surfaces its `StateError`).

## Not a breaking change

Valid JSON responses decode exactly as before, and a malformed body on a success status still throws as it always did. Only two cases change: a mislabeled non-JSON body on an **error** status now yields a `FunctionException` (with the raw text as `details`) instead of a bare `FormatException`, and an upper-cased `application/JSON` is now parsed as JSON.

## Tests

Added to `functions_dart_test.dart`:
- `invalid-json-error` — a `500` labeled `application/json` with an HTML body now throws `FunctionException(status: 500, details: '<html>…')`. Without the fix it throws a bare `FormatException`.
- `success-invalid-json` — a `200` labeled `application/json` with an HTML body still throws `FormatException`, confirming the fallback doesn't leak into success responses.
- `uppercase-json` — `application/JSON` with a JSON body now returns the decoded map. Without the fix it returned the raw string.

Full `functions_client` suite (41 tests) passes; `dart format` and `dart analyze --fatal-warnings` are clean.

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -180,7 +180,12 @@ class FunctionsClient {
         dynamic decoded;
         try {
           decoded = await _isolate.decode(bodyText);
-        } on FormatException catch (_) {
+        } on FormatException {
+          // A body labeled JSON that doesn't parse is only tolerated on an
+          // error status, where the raw text still needs to reach the caller
+          // as the exception `details`. On a success status it's a real
+          // anomaly, so keep surfacing it instead of handing back a String.
+          if (isSuccessStatus) rethrow;
           decoded = bodyText;
         }
         data = decoded;
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -163,7 +163,8 @@ class FunctionsClient {
             response.headers['content-type'] ??
             'text/plain')
         .split(';')[0]
-        .trim();
+        .trim()
+        .toLowerCase();
 
     final isSuccessStatus =
         200 <= response.statusCode && response.statusCode < 300;
@@ -172,9 +173,18 @@ class FunctionsClient {
 
     if (responseType == 'application/json') {
       final bodyBytes = await response.stream.toBytes();
-      data = bodyBytes.isEmpty
-          ? ""
-          : await _isolate.decode(utf8.decode(bodyBytes));
+      if (bodyBytes.isEmpty) {
+        data = "";
+      } else {
+        final bodyText = utf8.decode(bodyBytes);
+        dynamic decoded;
+        try {
+          decoded = await _isolate.decode(bodyText);
+        } on FormatException catch (_) {
+          decoded = bodyText;
+        }
+        data = decoded;
+      }
     } else if (responseType == 'application/octet-stream') {
       data = await response.stream.toBytes();
     } else if (responseType == 'text/event-stream' && isSuccessStatus) {
--- a/packages/functions_client/test/custom_http_client.dart
+++ b/packages/functions_client/test/custom_http_client.dart
@@ -71,6 +71,25 @@ class CustomHttpClient extends BaseClient {
           "Content-Type": "application/json",
         },
       );
+    } else if (request.url.path.endsWith('invalid-json-error')) {
+      return StreamedResponse(
+        Stream.value(utf8.encode('<html><body>502 Bad Gateway</body></html>')),
+        500,
+        request: request,
+        headers: {
+          "Content-Type": "application/json",
+        },
+        reasonPhrase: "Internal Server Error",
+      );
+    } else if (request.url.path.endsWith('uppercase-json')) {
+      return StreamedResponse(
+        Stream.value(utf8.encode(jsonEncode({"key": "Hello World"}))),
+        200,
+        request: request,
+        headers: {
+          "Content-Type": "application/JSON",
+        },
+      );
     }
     final Stream<List<int>> stream;
     final Map<String, String> headers;
--- a/packages/functions_client/test/custom_http_client.dart
+++ b/packages/functions_client/test/custom_http_client.dart
@@ -81,6 +81,15 @@ class CustomHttpClient extends BaseClient {
         },
         reasonPhrase: "Internal Server Error",
       );
+    } else if (request.url.path.endsWith('success-invalid-json')) {
+      return StreamedResponse(
+        Stream.value(utf8.encode('<html><body>not json</body></html>')),
+        200,
+        request: request,
+        headers: {
+          "Content-Type": "application/json",
+        },
+      );
     } else if (request.url.path.endsWith('uppercase-json')) {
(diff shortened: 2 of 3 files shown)


packages/functions_client/lib/src/functions_client.dart
```diff
@@ -163,7 +163,10 @@ class FunctionsClient {
             response.headers['content-type'] ??
             'text/plain')
         .split(';')[0]
-        .trim();
+        .trim()
+        // Media types are case-insensitive (RFC 9110), so `application/JSON`
+        // must be treated the same as `application/json`.
```
spydon: Please instruct you agent to stop adding comments like this everywhere, it's not needed 😄

packages/functions_client/lib/src/functions_client.dart
```diff
@@ -172,9 +175,23 @@ class FunctionsClient {
...
+        final bodyText = utf8.decode(bodyBytes);
+        dynamic decoded;
+        try {
+          decoded = await _isolate.decode(bodyText);
+        } on FormatException catch (_) {
+          // The response is labeled `application/json` but the body isn't
+          // valid JSON (e.g. an upstream proxy returning an HTML error page
+          // with a misleading Content-Type). Fall back to the raw text so the
+          // status code is still reported through FunctionResponse /
+          // FunctionException instead of the decode throwing out of invoke().
```
spydon:
```suggestion
```
Same here, it is already apparent from the code what is happening

packages/functions_client/test/custom_http_client.dart
```diff
@@ -71,6 +71,28 @@ class CustomHttpClient extends BaseClient {
           "Content-Type": "application/json",
         },
       );
+    } else if (request.url.path.endsWith('invalid-json-error')) {
+      // An error status whose body is labeled JSON but isn't valid JSON,
+      // e.g. an upstream proxy returning an HTML error page.
```
spydon:
```suggestion
```

packages/functions_client/test/custom_http_client.dart
```diff
@@ -71,6 +71,28 @@ class CustomHttpClient extends BaseClient {
...
+        Stream.value(utf8.encode('<html><body>502 Bad Gateway</body></html>')),
+        500,
+        request: request,
+        headers: {
+          "Content-Type": "application/json",
+        },
+        reasonPhrase: "Internal Server Error",
+      );
+    } else if (request.url.path.endsWith('uppercase-json')) {
+      // A valid JSON body advertised with an upper-cased media type.
```
spydon:
```suggestion
```

packages/functions_client/test/functions_dart_test.dart
```diff
@@ -39,6 +39,26 @@ void main() {
       );
     });
 
+    test('error response labeled JSON with a non-JSON body reports the status',
+        () async {
+      // The body can't be decoded as JSON, but invoke() must still surface the
+      // status via FunctionException instead of letting the decode throw.
```
spydon:
```suggestion
```

packages/functions_client/lib/src/functions_client.dart
spydon:
```suggestion
```

spydon (approved):

packages/functions_client/lib/src/functions_client.dart
AndroidPoet: Will do 😄

spydon:
I hope you're alright with me fixing up your PRs, I figured it's fast to just do it directly then having a round-trip of comments :)

AndroidPoet:
> I hope you're alright with me fixing up your PRs, I figured it's fast to just do it directly then having a round-trip of comments :)
Totally cool, man! I believe the product should be as good as it can be, even if my PR just helps bring attention to an issue. I'm completely fine with you fixing things directly.

I'm really loving the vibe here it's a fully chill zone. Is there an non-employee community, Slack, Discord, or any other group where I can learn about future plans and areas where I can contribute? Of course, if that's not allowed, no worries. I'd still love to contribute however I can

spydon:
> Totally cool, man! I believe the product should be as good as it can be, even if my PR just helps bring attention to an issue. I'm completely fine with you fixing things directly.
> 
> I'm really loving the vibe here it's a fully chill zone. Is there an non-employee community, Slack, Discord, or any other group where I can learn about future plans and areas where I can contribute? Of course, if that's not allowed, no worries. I'd still love to contribute however I can

There is a discord server here that you can join :)
https://discord.gg/xe7u9UD9N

There is a slack too, but I don't know how to invite people to it yet.

Another recommendation while we're on it, try telling the agent to do a self review before opening the PR, it usually finds some small issues. :)

AndroidPoet:
> > Totally cool, man! I believe the product should be as good as it can be, even if my PR just helps bring attention to an issue. I'm completely fine with you fixing things directly.
> > I'm really loving the vibe here it's a fully chill zone. Is there an non-employee community, Slack, Discord, or any other group where I can learn about future plans and areas where I can contribute? Of course, if that's not allowed, no worries. I'd still love to contribute however I can
> 
> There is a discord server here that you can join :) https://discord.gg/xe7u9UD9N
> 
> There is a slack too, but I don't know how to invite people to it yet.
> 
> Another recommendation while we're on it, try telling the agent to do a self review before opening the PR, it usually finds some small issues. :)

I hope I'm not contributingAI slop either . Even though I'm really careful with these things, and I worked on full parity with the SDKs in my Supabase KMP project, I'll be even more careful with any of my future contributions Thanks  https://github.com/AndroidPoet/supabase-kmp

PR #1514, opened 2026-07-02
feat(storage): add download option to signed and public URLs

spydon:
## What

Adds an optional `download` parameter to:

- `createSignedUrl`
- `createSignedUrlsResult` (and the deprecated `createSignedUrls`)
- `getPublicUrl`

Passing `true` appends `download=` so the CDN serves the file with a `Content-Disposition: attachment` header using its original name; passing a `String` overrides the download file name. `null`/`false` leave the URL unchanged.

## Why

SDK parity: supabase-js accepts `download?: string | boolean` on these methods. The Dart client had no way to force a download.

## Tests

- Mock-based unit tests cover `getPublicUrl` (original name, custom name, disabled) and the signed-URL methods (appending `&download=` after the existing `?token=` query).
- An end-to-end test uploads to a public bucket, then fetches the signed and public URLs and asserts the real `Content-Disposition: attachment` response header.

## Compliance

Marks `create_signed_url`, `create_signed_urls`, and `get_public_url` implemented in `sdk-compliance.yaml`.

Resolves SDK-1166

### DIFF
--- a/packages/storage_client/lib/src/storage_file_api.dart
+++ b/packages/storage_client/lib/src/storage_file_api.dart
@@ -357,10 +357,15 @@ class StorageFileApi {
   /// example, `60` for a URL which are valid for one minute.
   ///
   /// [transform] adds image transformations parameters to the generated url.
+  ///
+  /// [download] triggers the file to be downloaded rather than opened in the
+  /// browser by setting the response's `Content-Disposition` header. Pass
+  /// `true` to keep the original file name or a [String] to override it.
   Future<String> createSignedUrl(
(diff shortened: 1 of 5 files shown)


grdsdev (commented):
Since we're adding this feature now, I think we can do better, I don't think JS's design with the best design.

I did this for Swift https://github.com/supabase/supabase-swift/blob/main/Sources/Storage/Types.swift#L455-L475

Maybe you can do something similar for Dart.

grdsdev (approved):

PR #1516, opened 2026-07-02
feat(postgrest): add format option to explain()

spydon:
## What

Adds an `ExplainFormat` enum (`text`, `json`) and a `format` parameter to `PostgrestTransformBuilder.explain()` (default `text`). The value is interpolated into the `Accept` media type (`application/vnd.pgrst.plan+<format>`), so a JSON plan can now be requested.

The JSON plan is returned as a raw JSON string, keeping `explain()` type-stable as `Future<String>`.

## Why

SDK parity: supabase-js `explain({ format })` supports `'text' | 'json'`; the Dart client hardcoded `+text`.

## Fix

The response parser only treated `application/vnd.pgrst.plan+text` as a raw body; the `+json` plan was JSON-decoded into a `List` and then cast to `String`, throwing at runtime. Any `application/vnd.pgrst.plan` media type is now returned as a raw body.

## Tests

- Standalone mock-based unit tests assert the `Accept` header for the default text plan, an explicit JSON plan, and that other options are still forwarded alongside the format.
- An end-to-end test runs `explain(format: json)` against Postgres and asserts the returned plan parses as a JSON list with a `Plan` key.

## Compliance

Marks `database.using_modifiers.explain` implemented in `sdk-compliance.yaml` and registers the `ExplainFormat` symbol.

Resolves SDK-1168

### DIFF
--- a/packages/postgrest/lib/src/types.dart
+++ b/packages/postgrest/lib/src/types.dart
@@ -96,6 +96,15 @@ enum CountOption {
   estimated,
 }
 
+/// The format of the plan returned by `PostgrestTransformBuilder.explain`.
+enum ExplainFormat {
+  /// Returns the plan as human readable text.
+  text,
+
+  /// Returns the plan as JSON.
+  json,
+}
+
 // coverage:ignore-[start]
 /// Returns count as part of the response when specified.
 @Deprecated('Not used anywhere. Will be removed in the next major version.')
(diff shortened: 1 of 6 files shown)


grdsdev (approved):

packages/postgrest/lib/src/types.dart
```diff
@@ -96,6 +96,15 @@ enum CountOption {
   estimated,
 }
 
+/// The format of the plan returned by `PostgrestTransformBuilder.explain`.
+enum ExplainFormat {
+  /// Returns the plan as human readable text.
+  text,
+
+  /// Returns the plan as JSON.
+  json,
```
spydon:
```suggestion
  text,
  json,
```

PR #1517, opened 2026-07-02
feat(realtime): add onHeartbeat stream to RealtimeClient

spydon:
## What

Adds `RealtimeClient.onHeartbeat`, a broadcast `Stream<RealtimeHeartbeatStatus>` that emits a status whenever a heartbeat is pushed (`sent`), acknowledged (`ok`/`error`), or when a prior heartbeat goes unanswered (`timeout`).

```dart
final subscription = supabase.realtime.onHeartbeat.listen((status) {
  print('heartbeat: $status');
});
```

Also adds the `RealtimeHeartbeatStatus` enum (`sent`, `ok`, `error`, `timeout`).

## Why

SDK parity: realtime-js `RealtimeClient.onHeartbeat` exposes the same lifecycle. The Dart client had the heartbeat machinery but no way to observe it.

Exposing it as a `Stream` (rather than a single callback like realtime-js) is the idiomatic Dart shape: consumers get `listen`/`where`/`firstWhere`/`timeout`, multiple subscribers and cancellation for free, and it matches how the auth client already surfaces events (`onAuthStateChange`).

## Tests

- Unit tests cover each emitted status and a case where a non-matching heartbeat reference emits nothing.
- An end-to-end test connects to a real Realtime server (both v1 and v2 protocols) and observes an actual `sent` then `ok` heartbeat cycle.

## Compliance

Marks `realtime.client.listen_heartbeats` implemented in `sdk-compliance.yaml` and registers the `RealtimeClient.onHeartbeat` and `RealtimeHeartbeatStatus` symbols.

Resolves SDK-1169

### DIFF
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -58,7 +58,6 @@ enum RealtimeHeartbeatStatus {
   ok,
   error,
   timeout,
-  disconnected,
 }
 
 /// Manages a persistent WebSocket connection to the Supabase Realtime server.
@@ -388,8 +387,7 @@ class RealtimeClient {
     stateChangeCallbacks['message']!.add(callback);
   }
 
-  /// Emits a status whenever a heartbeat is sent, acknowledged, times out, or
-  /// is skipped because the socket is disconnected.
+  /// Emits a status whenever a heartbeat is sent, acknowledged, or times out.
   Stream<RealtimeHeartbeatStatus> get onHeartbeat =>
       _heartbeatController.stream;
 
@@ -628,7 +626,6 @@ class RealtimeClient {
   @internal
   Future<void> sendHeartbeat() async {
     if (!isConnected) {
-      _heartbeatController.add(RealtimeHeartbeatStatus.disconnected);
       return;
     }
 
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -54,20 +54,10 @@ class RealtimeCloseEvent {
 
 /// The lifecycle status of a heartbeat reported to [RealtimeClient.onHeartbeat].
 enum RealtimeHeartbeatStatus {
-  /// A heartbeat message was pushed to the server.
   sent,
-
-  /// The server acknowledged the heartbeat.
   ok,
-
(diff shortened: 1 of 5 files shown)


grdsdev (approved):

spydon:
@grdsdev I updated this to be a stream instead of a callback, I think that makes more sense here.
What do you think?
I'd love your input too @Vinzent03 :)

packages/realtime_client/lib/src/realtime_client.dart
```diff
@@ -605,6 +628,7 @@ class RealtimeClient {
   @internal
   Future<void> sendHeartbeat() async {
     if (!isConnected) {
+      _heartbeatController.add(RealtimeHeartbeatStatus.disconnected);
```
Vinzent03: I'm wondering if this is really called here. Shouldn't the heartbeat timer be canceled if it is not connected? I think this more of a safeguard, but not something that the user should expect to receive, right?

Vinzent03 (commented):
I was just reviewing it in this very moment. I really like using a stream instead of a callback mechanism. This feels way more Dart idiomatic. (I thought about switching to streams for other parts of the realtime_client as well, maybe something to consider for v3. Makes removing a listener way easier)

packages/realtime_client/lib/src/realtime_client.dart
spydon: Very true, I'd say we remove it, if there isn't any apparent race where this can actually happen.

I'll look into it.

spydon: Okay it seems like it can happen, but let's just drop it if that happens.
So I removed the disconnected message.

Vinzent03: I wonder what is different about supabase-js implementation having this, but yeah one should depend on `onClose` for this kind of event.

Vinzent03 (approved):

packages/realtime_client/lib/src/realtime_client.dart
spydon: It can only happen in a race there too, which is kind of weird, because then sometimes the user will get it and sometimes not.

spydon: https://github.com/supabase/supabase-js/issues/2488

PR #1519, opened 2026-07-02
feat(gotrue): add custom OAuth providers admin API with custom_claims_allowlist

spydon:
## What

Implements the custom OAuth/OIDC providers admin API in gotrue, exposed as `client.auth.admin.customProviders`, mirroring the `admin.customProviders` API in supabase-js (`/admin/custom-providers`). This is distinct from the existing OAuth 2.1 server *clients* API (`admin.oauth`).

Resolves [SDK-1173](https://linear.app/supabase/issue/SDK-1173/parityauth-add-custom-claims-allowlist-to-custom-providers-admin-api).

The parity issue asked to add `custom_claims_allowlist` to the custom providers admin types, but that API did not exist in the Flutter SDK at all, so this adds the full surface including the field.

## Changes

- New `GoTrueAdminCustomProvidersApi` with `listProviders({type})`, `createProvider`, `getProvider`, `updateProvider`, `deleteProvider`.
- Wired into `GoTrueAdminApi` as `admin.customProviders`.
- New types: `CustomProviderType`, `OIDCDiscoveryDocument`, `CustomOAuthProvider`, `CreateCustomProviderParams`, `UpdateCustomProviderParams`.
- All three provider types carry `customClaimsAllowlist` (serialized as `custom_claims_allowlist`), an opt-in allowlist of raw identity provider claim keys copied verbatim into the user's `custom_claims`. Optional, absent by default.

## Usage

```dart
await supabase.auth.admin.customProviders.createProvider(
  CreateCustomProviderParams(
    providerType: CustomProviderType.oidc,
    identifier: 'custom:mycompany',
    name: 'My Company',
    clientId: '...',
    clientSecret: '...',
    customClaimsAllowlist: ['groups', 'org_id', 'mail'],
  ),
);
```

## Tests

- Unit tests for serialization/deserialization of `custom_claims_allowlist` (present, absent, empty) across the create/update/read types.
- Integration tests covering create/list/list-by-type/get/update/delete against a live gotrue server.

Integration tests use OAuth2 providers (OIDC creation triggers a server-side discovery-document fetch, which would be network-dependent). They do not assert `customClaimsAllowlist` round-trips through the server, since the auth image bundled with the pinned Supabase CLI accepts the field but does not yet echo it in responses; serialization is covered by the unit tests.


Vinzent03:
I actually already vibe coded this api locally for a project right when the api came out 2 months ago, because I needed it for a project, but never took the time to polish it and create the pr :sweat_smile:

spydon:
> I actually already vibe coded this api locally for a project right when the api came out 2 months ago, because I needed it for a project, but never took the time to polish it and create the pr 😅

Sounds like you're the perfect person to review this then 😁

packages/gotrue/lib/src/types/custom_oauth_provider.dart
```diff
@@ -0,0 +1,442 @@
...
+  final String? userinfoUrl;
+
+  /// JWKS URI for token verification
+  final String? jwksUri;
+
+  /// OIDC discovery document (OIDC providers only)
+  final OIDCDiscoveryDocument? discoveryDocument;
+
+  /// Timestamp when the provider was created
+  final String createdAt;
```
Vinzent03: I know that supabase-js uses a string as well, but I think a `DateTime` object would be more fitting. Though this is more of a fundamental discussion as all timestamp fields in `Session` are a `String` currently as well. I'm open to discuss this for v3.

packages/gotrue/lib/src/types/custom_oauth_provider.dart
```diff
@@ -0,0 +1,442 @@
...
+  });
+
+  Map<String, dynamic> toJson() {
+    return {
+      'provider_type': providerType.name,
+      'identifier': identifier,
+      'name': name,
+      'client_id': clientId,
+      'client_secret': clientSecret,
+      if (acceptableClientIds != null)
```
Vinzent03: This could be refactored to use new null-aware elements with #1527 merged now.

packages/gotrue/lib/src/gotrue_admin_custom_providers_api.dart
```diff
@@ -0,0 +1,128 @@
...
+    );
+
+    return CustomOAuthProvider.fromJson(data);
+  }
+
+  /// Updates an existing custom provider.
+  ///
+  /// When `issuer` or `discoveryUrl` is changed on an OIDC provider, the server
+  /// re-fetches and validates the discovery document before persisting. This
+  /// may throw a validation error (`error_code: "validation_failed"`) if the
```
Vinzent03: `error_code` is what is sent in the http body, but we map it to `AuthException.code` https://github.com/supabase/supabase-flutter/blob/d8dcaeb2fc9128925677e1390826cede8df2f823/packages/gotrue/lib/src/fetch.dart#L84
So I think it is easier for the user to link to that being `validation_failed`.

packages/gotrue/lib/src/types/custom_oauth_provider.dart
spydon: True, i think we can just as well start using DateTime here now since we plan to use that in v3.
Less migration work for our users later.

Vinzent03 (approved):

mandarini (approved):

PR #1525, opened 2026-07-03
chore: migrate to pub workspaces and Melos 8

spydon:
> **Stacked on #1527.** Review/merge that one first. This PR targets `chore/bump-min-dart-3.9` and its diff shows only the workspace/Melos changes; the Dart 3.9 SDK bump, reformat, and analyzer fixes live in the base PR.

## What

Migrates the monorepo from Melos 3 (`melos.yaml` + `pubspec_overrides.yaml` linking) to [Dart pub workspaces](https://dart.dev/tools/pub/workspaces) with the latest Melos (8.1.0), following the [Melos migration guide](https://melos.invertase.dev/guides/migrations) and mirroring how [flame](https://github.com/flame-engine/flame/blob/main/pubspec.yaml) is set up.

## Changes

- **Root `pubspec.yaml`** now holds the workspace: a literal `workspace:` member list (globs would require Dart 3.11, we target 3.9), `dev_dependencies: melos: ^8.1.0`, and the Melos config moved under the `melos:` key. `melos.yaml` is deleted.
- **Every member** declares `resolution: workspace`, so pub links the local packages automatically. `pubspec_overrides.yaml` and `runPubGetInParallel`/`usePubspecOverrides` are gone.
- **All examples are workspace members**, including `examples/launcher`. Launcher was previously excluded because the old minimum Flutter shipped a Dart older than launcher required; now that the whole workspace is on Dart 3.9 it joins like the rest (bumped to `>=3.9.0`, `resolution: workspace`).
- **Built-in commands**: drops the hand-written `analyze`/`format`/`lint:all`/`test:coverage`/`upgrade`/`outdated` scripts in favor of the built-in `melos analyze`, `melos format` (line length configured in the `melos:` block) and `melos test`. Only the `update-version` script (used by the version hook) remains.
- **CI**: the Dart test matrix can no longer use `melos bootstrap --no-flutter` (removed in Melos, and `resolution: workspace` resolves the whole workspace at once, which includes `supabase_flutter` and therefore requires Flutter). It now uses the Flutter `stable`/`beta`/`master` channels, which bundle Dart `stable`/`beta`/`dev`, preserving multi-channel coverage (beta/master only on push to main, as #1524 set up). Affected-package detection now watches the root `pubspec.yaml`. Melos stays unpinned (the action/CLI install the latest, which supports workspaces).
- **passkeys example**: updated to pass a `PasskeyAuthenticator`, matching the current `supabase_flutter` API (a pre-existing break from #1444 that CI never covered, surfaced now that `melos analyze` runs the whole workspace).

## Validation

Locally with Melos 8.1.0 and Flutter 3.44 (Dart 3.12):

- `melos bootstrap` resolves the whole workspace with a single `flutter pub get` (13 packages).
- `melos analyze --fatal-infos` is green across all packages.
- `melos format --set-exit-if-changed` is clean.
- `melos test` runs (verified on a backend-free package).


grdsdev (approved):
lgtm!

PR #1530, opened 2026-07-03
feat(supabase_flutter): default debug logging off under flutter test

spydon:
## What

Makes `Supabase.initialize`'s `debug` option default to `false` when running under `flutter test`, so tests no longer emit `INFO`/`CONFIG` log noise unless they explicitly opt in with `debug: true`.

## Why

`debug` defaults to `debug ?? kDebugMode`. Under `flutter test`, `kDebugMode` is `true`, so every test that doesn't pass `debug: false` attaches the log subscription and spams the console (`***** Supabase init completed *****`, etc.). Rather than sprinkling `debug: false` across every test, this fixes the default at the source.

This is the more general fix behind #1529 (which added `debug: false` per call). With this change those manual additions are no longer needed, so **#1529 can be closed in favor of this**.

## How

Reuses the existing web-safe `isRunningInFlutterTest` getter (already used in `supabase_auth.dart`, defined via the `platform_stub.dart` / `platform_io.dart` conditional import, which reads the `FLUTTER_TEST` env var):

```dart
_instance._debugEnable = debug ?? (kDebugMode && !isRunningInFlutterTest);
```

Production behavior is unchanged: `isRunningInFlutterTest` is `false` outside tests (and the web stub returns `false`), so the expression reduces to the previous `debug ?? kDebugMode`.

## Verification

- New `test/debug_default_test.dart` asserts the default is off under `flutter test` and that explicit `debug: true` still logs.
- Full `packages/supabase_flutter` suite: all 59 tests pass with no stray supabase log output; `flutter analyze --no-fatal-infos` is clean.

### DIFF
--- a/packages/supabase_flutter/test/initialization_test.dart
+++ b/packages/supabase_flutter/test/initialization_test.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:supabase_flutter/supabase_flutter.dart';
@@ -9,6 +10,10 @@ void main() {
 
   const supabaseUrl = '';
   const supabaseKey = '';
+
+  setUpAll(() {
+    debugPrint = (String? message, {int? wrapWidth}) {};
+  });
 
   group('Supabase initialization', () {
     setUp(() {
@@ -29,6 +34,7 @@ void main() {
         await Supabase.initialize(
           url: supabaseUrl,
           publishableKey: supabaseKey,
+          debug: false,
         );
       });
     });
@@ -39,6 +45,7 @@ void main() {
         await Supabase.initialize(
           url: supabaseUrl,
           publishableKey: supabaseKey,
+          debug: false,
           authOptions: const FlutterAuthClientOptions(
             localStorage: localStorage,
           ),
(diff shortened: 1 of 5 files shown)


packages/supabase_flutter/test/initialization_test.dart
```diff
@@ -29,6 +34,7 @@ void main() {
         await Supabase.initialize(
           url: supabaseUrl,
           publishableKey: supabaseKey,
+          debug: false,
```
Vinzent03, line 37: What is the reason for specifying false here multiple times? Is it because those tests are also run on web?

spydon: Yeah exactly, there are some hacks that we could use to detect that it is in a test on web but I think they are so ugly that this is probably the better choice, and then in the long run we shouldn't do logging in this way at all (I'll open a v3 issue)

spydon: https://github.com/supabase/supabase-flutter/issues/1533

scripts/customer_testing.dart
Vinzent03, line 1: I suppose this should live in #1528?

spydon: Oops, must have based this off that branch instead of main. Good catch

Tr00d (approved):

PR #1531, opened 2026-07-03
chore: adopt null-aware elements now that minimum Dart SDK is 3.9

spydon:
## What

Now that the minimum Dart SDK is 3.9, this adopts the null-aware elements language feature (stabilized in Dart 3.8) across the codebase.

- Enable the previously SDK-gated `prefer-null-aware-elements` DCM rule in `supabase_lints`.
- Rewrite all 65 flagged collection literals to use null-aware elements (`?element`) instead of `if (x != null)` guards:
  - List/set: `if (prefer != null) prefer` → `?prefer`
  - Map value: `if (email != null) 'email': email` → `'email': ?email`
  - Guarded chains: `if (grantTypes != null) 'grant_types': grantTypes!.map(...)` → `'grant_types': ?grantTypes?.map(...)`

## Dot shorthands

The config previously claimed the three dot-shorthand rules needed "Dart 3.9". That is inaccurate: dot shorthands are still an **experimental** language feature and cannot be used by code targeting the 3.9 floor (verified: `Color c = .green;` still errors at language 3.9 on Dart 3.12). Those rules stay disabled, and their comment is corrected to reflect this.

## Verification

- `dcm analyze packages examples` → 0 `prefer-null-aware-elements` violations
- `dart analyze` → clean across all touched packages
- `dart format` → no reformatting needed
- `realtime_client` `mock_test.dart` passes (including the `with filter` case that exercises a changed path)

### DIFF
--- a/packages/gotrue/lib/src/gotrue_admin_api.dart
+++ b/packages/gotrue/lib/src/gotrue_admin_api.dart
@@ -118,8 +118,8 @@ class GoTrueAdminApi {
     final options = GotrueRequestOptions(
       headers: _headers,
       query: {
-        if (page != null) 'page': page.toString(),
-        if (perPage != null) 'per_page': perPage.toString(),
+        'page': ?page?.toString(),
+        'per_page': ?perPage?.toString(),
       },
     );
     final response = await _fetch.request(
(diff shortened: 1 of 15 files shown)


Vinzent03 (approved):
I immediately thought about the supabase packages when I read the release notes back then. Great to finally make use of it.

Tr00d (approved):
Not familiar with Dart, but is this a breaking change?

spydon:
> Not familiar with Dart, but is this a breaking change?

Nope, just a syntax change

PR #1526, opened 2026-07-03
feat(realtime_client): support new postgres changes filter operators, multi-filter, column selection, and replication-ready events

filipecabaco:
## What kind of change does this PR introduce?

Brings the realtime client to parity with the new Realtime server features:

* **New `postgres_changes` filter operators**: `like`, `ilike`, `is`, `match`, `imatch`, and `isdistinct`, exposed via `PostgresChangeFilterType`.
* **Operator negation**: `PostgresChangeFilter(negate: true)` prefixes any operator with `not.` (e.g. `status=not.in.(draft,archived)`).
* **Multiple filters**: `onPostgresChanges(filters: [...])` combines several `PostgresChangeFilter`s with an `AND`.
* **Column selection**: `onPostgresChanges(select: [...])` restricts the change payload to a subset of columns.
* **Replication-ready notification**: `RealtimeChannelConfig(replicationReady: true)` asks the server to emit a `system` event once the Postgres replication connection is ready. The new `RealtimeSystemPayload` type parses that payload.

PostgREST-style value quoting is applied to scalar and `in` filter values so reserved characters (`,()"\`) and surrounding whitespace don't break the server's filter parser.

https://linear.app/supabase/issue/SDK-1170/parityrealtime-add-postgres-changes-filter-builder-new-operators-and
https://linear.app/supabase/issue/SDK-1179/parityrealtime-add-replication-ready-broadcast-option-and-system-event

PR #1532, opened 2026-07-03
refactor: adopt null-aware elements and enable the DCM rule

spydon:
## What

Enables the DCM `prefer-null-aware-elements` rule and migrates all 65 existing violations to Dart 3.8 null-aware collection elements.

## Why

The minimum Dart SDK is now 3.9.0 (#1527), but the shared lint config still disabled several rules with a stale comment claiming the floor was 3.4.0. `prefer-null-aware-elements` relies on the Dart 3.8 null-aware elements feature, so it can be turned on now.

While investigating, I found the comment was also wrong about the other three rules: `prefer-returning-shorthands`, `prefer-shorthands-with-enums`, and `prefer-shorthands-with-static-fields` rely on **dot shorthands**, which the analyzer only accepts from **Dart 3.10.0**, above our current 3.9.0 floor. Enabling those would require a breaking minimum-SDK bump, so they stay disabled with a corrected comment.

## Changes

- `supabase_lints`: enable `prefer-null-aware-elements` (drop the `false` override so the recommended preset governs it) and correct the stale comment (null-aware = 3.8, dot shorthands = 3.10).
- Migrate 65 sites across `gotrue`, `postgrest`, `realtime_client`, `storage_client`, and `supabase` tests. All are behavior-preserving:
  - `if (x != null) 'k': x` → `'k': ?x`
  - `if (x != null) 'k': x!.map(...)` → `'k': ?x?.map(...)`
  - `if (x != null) x` (list/set) → `?x`

## Verification

- `dcm analyze packages`: no issues found (zero `prefer-null-aware-elements` remaining, no new issues).
- `dart analyze` across the workspace: no errors.
- Touched hermetic tests pass (`supabase` and `realtime_client` mock suites, including the realtime filter serialization path). The backend-dependent package tests are covered by CI.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_admin_api.dart
+++ b/packages/gotrue/lib/src/gotrue_admin_api.dart
@@ -118,8 +118,8 @@ class GoTrueAdminApi {
     final options = GotrueRequestOptions(
       headers: _headers,
       query: {
-        if (page != null) 'page': page.toString(),
-        if (perPage != null) 'per_page': perPage.toString(),
+        'page': ?page?.toString(),
+        'per_page': ?perPage?.toString(),
       },
     );
     final response = await _fetch.request(
@@ -138,7 +138,7 @@ class GoTrueAdminApi {
   }) async {
     final body = {
       'email': email,
-      if (data != null) 'data': data,
+      'data': ?data,
     };
     final fetchOptions = GotrueRequestOptions(
       headers: _headers,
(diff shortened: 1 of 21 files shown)


spydon:
Closing in favor of #1531, which lands the same null-aware migration on top of the latest main (this branch conflicts with #1526) and has passing CI.

PR #1534, opened 2026-07-06
fix(gotrue): OAuth server throws FormatException instead of ArgumentE…

Pieter-JanRobrechtCronos:

As mentioned in [this discussion](https://github.com/supabase/supabase-flutter/pull/1499#discussion_r3506322065) for #1499, the  `OAuthAuthorizationDetailsResponse.fromJson` best throws a `FormatException` to be inline with the rest of the codebase.

Currently, when a JSON without a `user` key is parsed by the `OAuthAuthorizationDetailsResponse.fromJson` factory, the factory throws an `ArgumentError`.

## What is the new behavior?

Now, when a JSON without a `user` key is parsed by the `OAuthAuthorizationDetailsResponse.fromJson` factory, the 
factory throws a `FormatException`.

## Additional context

@spydon from the discussion I had the idea that you updated this, but I haven't found any related changes, so I've made a small PR to fix this.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_oauth_api.dart
+++ b/packages/gotrue/lib/src/gotrue_oauth_api.dart
@@ -24,11 +24,37 @@ class OAuthAuthorizedClient {
   }
 }
 
-/// Response type representing the details of a pending OAuth authorization
-/// request.
+/// Result returned by [GoTrueOAuthApi.getAuthorizationDetails].
+///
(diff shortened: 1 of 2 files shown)


Pieter-JanRobrechtCronos:
I'm closing this PR because I unfortunately pointed this PR from the Unlock'd main branch where I also created a commit for a small bug fix which I wanted to create a separate PR for. I'm closing this so that I can properly create a PR for both, unrelated changes.

PR #1535, opened 2026-07-06
fix(gotrue): OAuth server throws FormatException instead of ArgumentE…

Pieter-JanRobrechtCronos:

As mentioned in [this discussion](https://github.com/supabase/supabase-flutter/pull/1499#discussion_r3506322065) for #1499, the  `OAuthAuthorizationDetailsResponse.fromJson` best throws a `FormatException` to be inline with the rest of the codebase.

Currently, when a JSON without a `user` key is parsed by the `OAuthAuthorizationDetailsResponse.fromJson` factory, the factory throws an `ArgumentError`.

## What is the new behavior?

Now, when a JSON without a `user` key is parsed by the `OAuthAuthorizationDetailsResponse.fromJson` factory, the 
factory throws a `FormatException`.

## Additional context

@spydon from the discussion I had the idea that you updated this, but I haven't found any related changes, so I've made a small PR to fix this.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_oauth_api.dart
+++ b/packages/gotrue/lib/src/gotrue_oauth_api.dart
@@ -58,10 +58,9 @@ class OAuthAuthorizationDetailsResponse {
     final user = json['user'] == null ? null : User.fromJson(json['user']);
 
     if (user == null) {
-      throw ArgumentError.value(
-        json,
-        'json',
+      throw FormatException(
         'The provided JSON should contain a parseable user object',
+        json.toString(),
       );
     }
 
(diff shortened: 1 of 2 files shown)


spydon (approved):
This is technically a breaking change, but I think we can let it slip.

Pieter-JanRobrechtCronos:
Indeed, the signature changes so it makes sense to make it a breaking change. I can change the commit to mark it as breaking since the automatic checks are failing as well on the commit message?

spydon:
> Indeed, the signature changes so it makes sense to make it a breaking change. I can change the commit to mark it as breaking since the automatic checks are failing as well on the commit message?

It's failing because the first letter is capital, we really should remove the casing check 😅

PR #1536, opened 2026-07-06
fix(gotrue): handle already-consented OAuth authorization responses

Pieter-JanRobrechtCronos:
> Stacked on #1535. Review/merge that one first.

Bug fix

`getAuthorizationDetails` crashed with a `FormatException` when the OAuth 2.1 server short-circuited the consent flow. This happens when a user who already approved a client starts a new authorization for it: the server returns a redirect-only body (`{"redirect_url": "...?code=..."}`) with no user or client, which the parser rejected because it required a user object.

## What is the new behavior?

`getAuthorizationDetails` now checks whether a `redirect_url` is present returning `OAuthAuthorizationRedirectResponse` when it is present. When the key isn't present the old `OAuthAuthorizationDetailsResponse` object is returned.

Callers now switch on the result and forward the redirect instead of losing it to an exception. Genuinely malformed detail bodies (no redirect_url and no user) still throw, so validation isn't weakened.

## Additional context

BREAKING CHANGE: `getAuthorizationDetails` now returns the sealed `OAuthAuthorizationResponse` instead of `OAuthAuthorizationDetailsResponse`. Callers must switch/cast to access client, user, scope and redirectUri.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_oauth_api.dart
+++ b/packages/gotrue/lib/src/gotrue_oauth_api.dart
@@ -24,11 +24,37 @@ class OAuthAuthorizedClient {
   }
 }
 
-/// Response type representing the details of a pending OAuth authorization
-/// request.
+/// Result returned by [GoTrueOAuthApi.getAuthorizationDetails].
+///
+/// The OAuth 2.1 server responds in one of two ways, depending on whether the
+/// signed-in user has already granted consent to the requesting client:
+///
+/// * [OAuthAuthorizationDetailsResponse] — consent is still required, so the
+///   requesting client, user and requested scopes are returned for display in
+///   a consent screen.
+/// * [OAuthAuthorizationRedirectResponse] — the user already granted consent
+///   to this client, so the server short-circuits the consent flow and returns
+///   only the redirect URL the caller should navigate to.
 ///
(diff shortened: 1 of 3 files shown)


Pieter-JanRobrechtCronos:
@spydon I spotted this bug when I was working with the code that I've introduced for #1499. I hope you don't mind me creating a PR before opening an issue. I'm happy to create an issue if you still require one!

spydon:
> @spydon I spotted this bug when I was working with the code that I've introduced for #1499. I hope you don't mind me creating a PR before opening an issue. I'm happy to create an issue if you still require one!

It's all good, when it's not a new feature it's fine to create a PR directly.

spydon (approved):
Generally we shouldn't pull in breaking changes now, but since this fixes a bug it's fine.

Pieter-JanRobrechtCronos:
Ah I see, I can still make some changes if you prefer? I don't see how I would be able to keep out the breaking change though.

spydon:
@Pieter-JanRobrechtCronos can you turn on so that maintainers can modify the PR? Then I can fix up the compliance matrix

Pieter-JanRobrechtCronos:
I've checked, and I don't see any checkmarks that I can toggle to give you the necessary access... Is that something that I would need to toggle in my fork or a setting in this PR?

In the meantime I've made the necessary changes to the compliance matrix.

PR #1545, opened 2026-07-08
feat(functions): expose FunctionsFetchException/FunctionsRelayException/FunctionsHttpException subtypes

spydon:
## What

Introduces three error subtypes as subclasses of the existing `FunctionException`, so consumers can match on the failure mode:

- **`FunctionsFetchException`** — network/transport failure reaching the function, before any response is received. The originating error is available in `details`; `status` is `0`.
- **`FunctionsRelayException`** — error returned by the relay, detected via the `x-relay-error: true` response header.
- **`FunctionsHttpException`** — non-2xx response from the function itself.

`invoke()` now throws the appropriate subtype depending on the failure mode.

## Naming

The subtypes use the `Exception` suffix (not `Error`) to follow Dart's `Error`/`Exception` distinction: they are meant to be caught, extend `FunctionException` (which implements `Exception`), and wrap an underlying `ClientException`. This also keeps them consistent with the base class name.

## Why non-breaking

Existing `catch` blocks that catch `FunctionException` continue to work unchanged; the new subtypes only refine the type consumers can optionally match on. No existing signature or default behavior changes. `toString()` now uses `\$runtimeType` so subtypes print their own name.

A follow-up for making `FunctionException` a `sealed` class (breaking, enabling exhaustive `switch`) is tracked in #1550 for v3.

## Tests

Added `network-error` and `relay-error` fixtures to the custom HTTP client, plus cases asserting each subtype is thrown, that `status`/`details`/`reasonPhrase` are populated, and that the subtypes remain catchable as `FunctionException`.

Closes SDK-1258

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -165,7 +165,7 @@ class FunctionsClient {
     try {
       response = await (_httpClient?.send(request) ?? request.send());
     } catch (error) {
-      throw FunctionsFetchError(details: error);
+      throw FunctionsFetchException(details: error);
     }
     final responseType =
         (response.headers['Content-Type'] ??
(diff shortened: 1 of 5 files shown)


grdsdev (approved):

Vinzent03:
Isn't the suffix `Exception` instead of `Error` more correct here? They originate from an exception (`ClientException`), even extend `Exception` and are meant to be catched. Dart distinguishes between Error and Exception.

spydon:
> Isn't the suffix `Exception` instead of `Error` more correct here? They originate from an exception (`ClientException`), even extend `Exception` and are meant to be catched. Dart distinguishes between Error and Exception.

Very true! This just matched the JS side, I'll update it

PR #1546, opened 2026-07-08
chore(functions): mark invoke() timeout parity as not_applicable

spydon:
## What

Marks `functions.invocation.timeout` as `not_applicable` in `sdk-compliance.yaml` instead of implementing it.

## Why

This started as an implementation of a per-invocation `timeout` on `FunctionsClient.invoke()`. As raised in review, wrapping the returned `Future` in `Future.timeout` does **not** abort the in-flight request. So a caller could still see the function succeed after the timeout fires, and the behavior is identical to the caller applying `.timeout()` on the result themselves.

Exposing it as an `invoke()` option therefore adds no real capability and would be misleading, so the implementation is dropped and the capability is recorded as `not_applicable` with an explanatory note. A genuinely useful version would require request cancellation (tracked separately as `functions.invocation.request_cancellation`).

## Changes

- Added a `not_applicable` status to the legend in `sdk-compliance.yaml`.
- Set `functions.invocation.timeout` to `not_applicable` with a note explaining the no-op wrapper reasoning.

Closes SDK-1262.

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -1,4 +1,3 @@
-import 'dart:async';
 import 'dart:convert';
 import 'dart:typed_data';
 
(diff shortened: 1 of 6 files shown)


grdsdev (dismissed):

Vinzent03:
Is this any different from applying the timeout to the future of the invoke in the users code directly?  The timeout does not abort the request, which users may expect. So even if you get a timeout exception a function call could still succeed.

spydon:
> Is this any different from applying the timeout to the future of the invoke in the users code directly? The timeout does not abort the request, which users may expect. So even if you get a timeout exception a function call could still succeed.

Indeed, this is kind of useless. I'll tag it as `not_applicable` instead (didn't know we had that state).

Tr00d (approved):

PR #1548, opened 2026-07-08
docs(auth): document getOAuthSignInUrl for URL-without-launch, mark oauth parity implemented

spydon:
## What

Marks `auth.sign_in.sign_in_with_oauth` as `implemented` in the SDK compliance matrix and documents the URL-without-launch path.

- `sdk-compliance.yaml`: `auth.sign_in.sign_in_with_oauth` moves from `partially_implemented` to `implemented`.
- `signInWithOAuth` doc comment now points callers to `getOAuthSignInUrl` when they want the OAuth URL without launching a browser.

## Why

The compliance gap was "the `skipBrowserRedirect` option is not publicly exposed." In supabase-js `skipBrowserRedirect` does two things: (1) suppresses the automatic `window.location.assign(url)` browser navigation, and (2) adds a `skip_http_redirect` server hint.

In Flutter, half (1) is already the default behavior of `getOAuthSignInUrl`, which builds and returns the OAuth `OAuthResponse` (including the URL) without ever launching a browser. That is the meaningful Flutter equivalent of "skip the browser redirect and hand me the URL."

Adding an explicit `skipBrowserRedirect` flag was considered and dropped:
- On `signInWithOAuth` it would only build the URL and discard it, since that method returns `Future<bool>` (launch success), not the URL.
- On `getOAuthSignInUrl` it would only toggle the `skip_http_redirect` server hint, which is a niche flag that makes the server return JSON instead of a 302, undesirable for the common flow of opening the URL in a browser/webview, and oddly named on a method that never touches a browser.

So the capability is satisfied by the existing API; this PR reclassifies it and adds a doc pointer.

## Non-breaking

Documentation and compliance-metadata only; no API or behavior change.

Closes SDK-1261

### DIFF
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -161,7 +161,12 @@ class FunctionsClient {
 
     _log.finest('Request: ${request.method} ${request.url} ${request.headers}');
 
-    final response = await (_httpClient?.send(request) ?? request.send());
+    final http.StreamedResponse response;
+    try {
+      response = await (_httpClient?.send(request) ?? request.send());
+    } catch (error) {
+      throw FunctionsFetchException(details: error);
+    }
     final responseType =
         (response.headers['Content-Type'] ??
                 response.headers['content-type'] ??
@@ -170,8 +175,11 @@ class FunctionsClient {
             .trim()
             .toLowerCase();
 
+    final isRelayError = response.headers['x-relay-error'] == 'true';
     final isSuccessStatus =
-        200 <= response.statusCode && response.statusCode < 300;
+        200 <= response.statusCode &&
+        response.statusCode < 300 &&
+        !isRelayError;
 
     final dynamic data;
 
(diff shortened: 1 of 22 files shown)


grdsdev (commented):
I don't think this makes much sense in the way it is implemented, it's different then JS.

In JS the browser redirect is a side-effect of the method, which you can disable using the `skipBrowserRedirect`, but it still returns the URL.

Here it has no side-effect, the result of the method call is the redirect, so passing `skipBrowserRedirect` you're just saying "build the URL and if skipBrowserRedirect is true, discard it"

spydon:
> I don't think this makes much sense in the way it is implemented, it's different then JS.
> 
> In JS the browser redirect is a side-effect of the method, which you can disable using the `skipBrowserRedirect`, but it still returns the URL.
> 
> Here it has no side-effect, the result of the method call is the redirect, so passing `skipBrowserRedirect` you're just saying "build the URL and if skipBrowserRedirect is true, discard it"

Very true, this should pretty much just be marked as implemented in the matrix then.

Tr00d (approved):

PR #1563, opened 2026-07-08
feat(auth): add async getSession() with on-demand refresh

spydon:
## Summary

Adds an async `getSession()` to `GoTrueClient`, alongside the existing synchronous `currentSession` getter. It returns the current session and, when the access token has expired, refreshes it on demand before returning. Concurrent callers share a single refresh through the same completer-based de-duplication used by `refreshSession`, so an expired session's refresh token is only spent once (the Dart equivalent of supabase-js's `_acquireLock`).

When a refresh fails but the access token is still within its real validity window (it was only inside the 30s expiry margin), the still-valid session is returned instead of surfacing the error, mirroring supabase-js's proactive-preserve behavior. Otherwise the `AuthException` propagates. Returns `null` when there is no session.

Nothing existing changes; `currentSession` stays.

## Outcome

Implemented.

- **Linear:** SDK-1274
- **supabase-js reference:** `GoTrueClient.getSession` / `__loadSession` (default lockless path)
- **Compliance matrix:** `auth.session.get_session` → `implemented` (was `partially_implemented`), registered `GoTrueClient.getSession`

## Tests

New `packages/gotrue/test/get_session_test.dart` covers: no session, valid session (no request), on-demand refresh of an expired session, concurrent de-duplication, throw when unrefreshable, and the still-valid proactive-preserve fallback. All pass.


Tr00d (approved):

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -192,6 +192,51 @@ class GoTrueClient {
   /// Returns the current session, if any;
   Session? get currentSession => _currentSession;
 
+  /// Returns the current session, refreshing it on demand when the access
+  /// token has expired.
+  ///
+  /// Unlike the synchronous [currentSession] getter, this waits for any
+  /// in-flight refresh to settle and, when the session has expired, attempts
+  /// to refresh it before returning. Concurrent callers share a single refresh
```
Vinzent03: I feel like it is not important for the user of the method that it properly deduplicates the token refresh and the refresh is shared. The user shouldn't care about these aspects.
I would move the last sentence to line 224 here.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -192,6 +192,51 @@ class GoTrueClient {
   /// Returns the current session, if any;
   Session? get currentSession => _currentSession;
 
+  /// Returns the current session, refreshing it on demand when the access
+  /// token has expired.
+  ///
+  /// Unlike the synchronous [currentSession] getter, this waits for any
+  /// in-flight refresh to settle and, when the session has expired, attempts
```
Vinzent03: What exactly do you mean with it "waits for any in-flight refresh to settle"? I don't think this is the case here. If a refresh is ongoing via `refreshSession` it would still return the previous session if it is not expired yet.

packages/gotrue/lib/src/gotrue_client.dart
```diff
@@ -192,6 +192,51 @@ class GoTrueClient {
...
+      final response = await _callRefreshToken(refreshToken);
+      return response.session;
+    } on AuthException {
+      final current = _currentSession;
+      final expiresAt = current?.expiresAt;
+      if (current != null &&
+          expiresAt != null &&
+          DateTime.fromMillisecondsSinceEpoch(
+            expiresAt * 1000,
+          ).isAfter(DateTime.now())) {
```
Vinzent03: While this works fine here, there is also this code section https://github.com/supabase/supabase-flutter/blob/61a353c29e1b6950fd628b44f405ade91044100e/packages/supabase/lib/src/supabase_client.dart#L269 where the same is implemented. I think a new getter on the Session object to check whether it is expired or not without margin could be beneficial. While I think this could be beneficial for the user as well, we could also mark it as internal if we prefer that.

packages/gotrue/lib/src/gotrue_client.dart
spydon: I think it is correct, see [here](https://github.com/supabase/supabase-flutter/blob/25d958cb378880eec8bd4050ddec38bd3c6e314e/packages/gotrue/lib/src/gotrue_client.dart#L1506), it checks whether there is already a refresh in-flight and if so it doesn't do another call, but waits for that one.

packages/gotrue/lib/src/gotrue_client.dart
spydon: Fixed in 58a7669

packages/gotrue/lib/src/gotrue_client.dart
spydon: Fixed it a bit differently in [4eca2c1](https://github.com/supabase/supabase-flutter/pull/1563/commits/4eca2c1771b0b60264db7fec2b3305ccba411cf8), the supabase_client shouldn't need to handle the token expiration manually since it's already done in `getSession`, and I don't think we need to widen the API surface with another getter on the session object just yet (I think almost all users' usecases should be possible to handle without that).

packages/gotrue/lib/src/gotrue_client.dart
Vinzent03: Yeah but that only prevents creating another refresh call.

```dart
client.refreshSession(); // manually refresh the session by the user while the session is still valid
client.getSession(); // Before the refresh is finished, already returns the still valid session, but does not wait for the already running refresh. The returned session may soonly be outdated.
```

Maybe I'm just reading the comment differently, but I understand it in the way it waits for any currently running refresh before returning a session. Even if the current session is still valid.

spydon: It is a little bit vague I guess, I'll re-word it a bit to be more clear. :)

Vinzent03 (approved):

PR #1564, opened 2026-07-08
feat(client): add opt-in trace context propagation headers

spydon:
## Summary

Adds opt-in W3C trace context propagation across outgoing Supabase requests, matching supabase-js observability support. Disabled by default, so existing clients send no additional headers.

When enabled via a new `TracePropagationOptions`, the client injects `traceparent`, `tracestate`, and `baggage` headers onto requests targeting Supabase hosts (`*.supabase.co`, `*.supabase.in`, the project host, and loopback addresses). Third-party hosts never receive trace headers. The upstream sampling decision is respected by default.

Idiomatic difference from supabase-js: JS auto-extracts the active context from the OpenTelemetry global API. Dart has no ubiquitous ambient trace context, so the context is supplied via a `traceContextProvider` callback (the same pattern the SDK already uses for `accessToken`). This wires cleanly into the existing `AuthHttpClient` so postgrest, storage, functions, and gotrue transports are all covered.

**Outcome:** implemented

## Changes

- `packages/supabase/lib/src/trace_propagation.dart` — new public `TraceContext`, `TracePropagationOptions`, `TraceContextProvider`.
- `packages/supabase/lib/src/trace_http_client.dart` — internal `TracePropagationClient` that injects trace headers for Supabase targets, respecting sampling.
- `packages/supabase/lib/src/supabase_client.dart` — new `tracePropagationOptions` parameter wired through the auth/gotrue HTTP clients.
- `packages/supabase_flutter/lib/src/supabase.dart` — forwards `tracePropagationOptions` through `Supabase.initialize`.
- Tests in `packages/supabase/test/trace_propagation_test.dart`.

## Reference

supabase-js `@supabase/tracing` package (`packages/shared/tracing`) and its wiring in `packages/core/supabase-js/src/lib/fetch.ts`.

## Compliance matrix

`client.observability.trace_propagation` → `implemented` (the traceContextProvider callback is the idiomatic Dart equivalent of supabase-js auto-extracting from the OpenTelemetry global API).

SDK-1275


packages/supabase/lib/src/supabase_client.dart
```diff
@@ -145,14 +149,25 @@ class SupabaseClient {
...
        _hasCustomIsolate = isolate != null {
+    final baseHttpClient = httpClient ?? Client();
+    final tracedHttpClient = tracePropagationOptions.enabled
+        ? TracePropagationClient(
+            baseHttpClient,
+            tracePropagationOptions,
+            supabaseUrl,
+          )
+        : baseHttpClient;
+    _gotrueHttpClient = tracePropagationOptions.enabled
```
Vinzent03: If no tracing is enabled and no custom http client was passed to the `SupabaseClient` constructor this results in null. This works fine, but I think it is more efficient if we always supply a Client even if it is just the default one like it is done with the `tracedHttpClient` above. So I would replace the `httpClient` here with `baseHttpClient`.

Vinzent03 (approved):

sdk-compliance.yaml
```diff
@@ -514,4 +514,18 @@ features:
       - FlutterAuthClientOptions.persistSession
   client.request_configuration.custom_http_client: implemented
   client.request_configuration.global_headers: implemented
-  client.observability.trace_propagation: not_implemented
+  client.observability.trace_propagation:
+    status: partially_implemented
+    note: "Opt-in W3C trace context propagation (traceparent/tracestate/baggage) to Supabase hosts only, with sampling-decision respect. Unlike supabase-js, which auto-extracts the active context from the OpenTelemetry global API, the Dart client takes a traceContextProvider callback since there is no ubiquitous ambient trace context in Dart."
```
grdsdev: I'd consider it as `implemented` then, if this is how it works with Dart.

spydon: True, fixed in [fe9513f](https://github.com/supabase/supabase-flutter/pull/1564/commits/fe9513f25d133f07e74921d900c7ab078c2bc983)

mandarini (approved):

Tr00d (approved):

PR #1571, opened 2026-07-09
refactor: drop retry dependency in favor of an in-house helper

spydon:
## What

Removes the `retry` package from both `gotrue` and `storage_client`, replacing it with a small self-contained `RetryOptions`/`retry` helper (`lib/src/retry.dart`) in each package.

- `gotrue` used the top-level `retry()` (token refresh).
- `storage_client` used `RetryOptions(...).retry(...)` (request retries).

## Why

Both packages pulled in `retry` for the same narrow use: exponential backoff gated by a `retryIf` predicate. The helper is ~40 lines and mirrors the package's behavior exactly:

- 200 ms delay factor, `2^attempt` backoff
- 25% jitter (`randomizationFactor`)
- 30 s max delay
- 8 max attempts by default

`gotrue`'s time-budget `retryIf` depends on these exact defaults, so they are preserved.

## Testing

- New `retry_test.dart` in `gotrue` covers success-after-retries, `maxAttempts` cutoff, `retryIf` short-circuit, `RetryOptions.retry` (the storage usage shape), and exponential/capped delay math — all pass.
- `dart analyze` clean for both packages.

Note: the helper is duplicated in each package because they are published independently and there is no shared internal utility package.

### DIFF
--- a/packages/gotrue/lib/src/gotrue_client.dart
+++ b/packages/gotrue/lib/src/gotrue_client.dart
@@ -13,7 +13,7 @@ import 'package:gotrue/src/types/fetch_options.dart';
 import 'package:http/http.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
-import 'package:retry/retry.dart';
+import 'package:gotrue/src/retry.dart';
 
 import 'broadcast_stub.dart'
     if (dart.library.js_interop) './broadcast_web.dart'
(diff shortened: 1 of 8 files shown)


Tr00d (approved):

o-santi (commented):
> Note: the helper is duplicated in each package because they are published independently and there is no shared internal utility package.

I would strongly recommend on creating a shared utility package for this use case. Even though it is the only one currently, in my experience a lot more shared code will appear, and you'll see the number of duplicated "utilities" grow over time.

Tr00d:
@o-santi  There's another PR specifically for this, a new `common` package for reusability

PR #1577, opened 2026-07-10
feat(postgrest): support aborting requests

spydon:
## Summary

Adds request aborting to the PostgREST client, following up on the discussion in #1560 (comment): now that the repo is on `http ^1.6.0`, in-flight requests can actually be cancelled via the [`Abortable`](https://pub.dev/packages/http#aborting-requests) API, so the reasoning that removed cancellation-based features from #1560 no longer applies.

> Stacked on top of #1560 (base branch `worktree-SDK-1270`). Review/merge that first.

New per-request API on `PostgrestBuilder`:

```dart
final abort = Completer<void>();
Timer(const Duration(seconds: 5), abort.complete);
try {
  final data = await supabase.from('users').select().abortCompleter(abort);
} on RequestAbortedException {
  // The request was aborted before it completed.
}
```

- `PostgrestBuilder.abortCompleter(Completer<void> completer)` — aborts the request when `completer` completes. Available on the query, filter and transform builders, and preserved across further chaining.
- When triggered, the request throws `RequestAbortedException` (re-exported from `package:http`).
- Aborting short-circuits the auto-retry loop: an aborted request is intentional and is never retried.
- The abortable path is only used when a completer is set, so all existing requests keep their exact current send behavior.

## Design notes

- The API name follows the proposal in #1368 (`abortCompleter`). `supabase-js` uses `abortSignal`; since Dart has no `AbortSignal`, a `Completer<void>` is the idiomatic equivalent. Open to `abortSignal(Future<void>)` instead if preferred, which would also allow `Future.delayed`-based timeouts directly.
- Aborting depends on the underlying `Client` supporting it (the platform `IOClient` / `BrowserClient` in `http` 1.6 do).

## Request timeout

`database.configuration.request_timeout` is still `not_applicable` in the matrix. It is now technically feasible to build a real per-request timeout on top of this (a `Future.delayed` as the abort trigger), so that entry could be revisited in a follow-up if we want a first-class timeout option.

## Compliance matrix

- `database.using_modifiers.request_cancellation` → `implemented`

## Tests

`packages/postgrest/test/abort_test.dart`: aborting a GET, aborting an insert (POST), no-retry-on-abort, normal completion when never triggered, and preservation across chaining.

Closes #1368

### DIFF
--- a/packages/postgrest/lib/postgrest.dart
+++ b/packages/postgrest/lib/postgrest.dart
@@ -1,6 +1,8 @@
 /// PostgREST client for Dart. Provides an ORM interface to PostgREST.
 library;
 
+export 'package:http/http.dart' show RequestAbortedException;
+
 export 'src/postgrest.dart';
 export 'src/postgrest_builder.dart';
 export 'src/types.dart';
--- a/packages/postgrest/lib/src/postgrest_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_builder.dart
@@ -91,6 +91,7 @@ class PostgrestBuilder<T, S, R> implements Future<T> {
   final YAJsonIsolate? _isolate;
   final CountOption? _count;
   final _RetryConfig _retry;
+  final Future<void>? _abortTrigger;
   final _log = Logger('supabase.postgrest');
 
   /// HTTP status codes that trigger an automatic retry by default.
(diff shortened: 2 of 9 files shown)


spydon:
https://github.com/supabase/supabase-flutter/pull/1368 already deals with this, so I'm closing this one.

PR #1610, opened 2026-07-16
feat: support more realtime filters and multiple filters for stream

Vinzent03:

feature; catching up with newer capabilities of the realtime client for the stream method.

- The `stream` did not catch up with newly supported realtime filters like `like`, `ilike`, `match`, `imatch`, `isDistinct`
- Previously realtime did only allow one filter, but that has changed and is not yet reflected in the stream method.

## What is the new behavior?

- Allow the new filters as well
- Allow any count of filters
- Improve documentation of the `stream` method by mentioning two realtime caveats affecting the behavior of the stream method

I've also added the first integration tests for the stream method. Ensuring the values are all properly serialized as well.

## Additional context

- realtime filters documentation: https://supabase.com/docs/guides/realtime/postgres-changes?queryGroups=language&language=dart#available-filters

This pr does not yet cover adding negated stream filters. Will add that in a future pr.

## Summary by CodeRabbit

* **New Features**
  * Added support for combining multiple filters in real-time `stream()` calls using AND semantics.
  * Added stream filters including `like`, `ilike`, `match`, `imatch`, `isFilter`, and `isDistinct`.
* **Bug Fixes**
  * Improved reliability of Docker-based real-time integration tests with longer timeouts.
* **Documentation**
  * Expanded streaming guidance covering filtering, behavior, supported operators, and event handling.
* **Chores**
  * Enabled real-time streaming updates for the `public.users` table.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,7 +72,7 @@ jobs:
               continue
             fi
             case "$package" in
-              gotrue|postgrest|realtime_client|storage_client) backend=true ;;
+              gotrue|postgrest|realtime_client|storage_client|supabase) backend=true ;;
               *) backend=false ;;
             esac
             case "$package" in
--- a/packages/supabase/dart_test.yaml
+++ b/packages/supabase/dart_test.yaml
@@ -1,0 +1,5 @@
+tags:
+  integration:
+    # Integration tests talk to a Dockerized Realtime server, which can be
+    # slower than the default timeout allows.
+    timeout: 2x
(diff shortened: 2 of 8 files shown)


Vinzent03:
This pr should be ready now. Adding the tests took way longer than expected to get to my liking's. I got really strange side effects of trying to use `isDistinct` or `isFilter` on list columns. 
There's also a strange extra delay for the first test case, because the replication needs more time to set up on the first one. I will probably come up with another pr to better time the postgrest call and first event if the replication is actually established and not when only the channel is subscribed.

spydon (approved):

PR #1626, opened 2026-07-21
feat(realtime): surface upgrade rejection details as RealtimeConnectException

spydon:
## Summary

When the Realtime server rejects the WebSocket upgrade, for example on a missing or invalid `apikey`, the client previously surfaced only a bare status code buried in a `WebSocketChannelException`:

```
WebSocketChannelException: WebSocketException: Connection to '...' was not upgraded to websocket, HTTP status code: 401
```

The server returns a helpful body, but the WebSocket handshake discards it (only the status is visible on native, nothing at all on web). This PR recovers that information and surfaces it as a new `RealtimeConnectException` on both native and web:

```
type: RealtimeConnectException
statusCode: 401
errorCode: UNAUTHORIZED_INVALID_API_KEY
message: Invalid API key
hint: Double check your Supabase `anon` or `service_role` API key.
```

## What changed

- `RealtimeClient`: when `WebSocketChannel.ready` fails, a new `_fetchConnectionError` helper re-requests the same endpoint over HTTP to recover the server's error response, then surfaces it through the existing `onError` and `subscribe` error paths.
- New public `RealtimeConnectException` (`statusCode`, `message`, `hint`, `errorCode`), exported from the package barrel.
- New integration test `test/invalid_apikey_test.dart` asserting an invalid apikey yields a `RealtimeConnectException`.

## How it works

On a failed upgrade, `_fetchConnectionError` issues a plain `GET` to the same endpoint (scheme mapped `ws`/`wss` to `http`/`https`, with the apikey already present in the query). A non-2xx response is parsed into `RealtimeConnectException(statusCode, message, hint, errorCode)`, reading `message`/`hint` from a JSON body and the code from the `sb-error-code` header. On a 2xx response or an unreachable server it falls back to the original error. The request reuses the injected `httpClient` when provided, otherwise a short-lived client with a 5 second timeout.

The standard WebSocket transport is left untouched, so this works identically on native and web with no new dependency and no reimplementation of the WebSocket handshake.

## Testing

- New `test/invalid_apikey_test.dart` (integration): an invalid apikey yields a `RealtimeConnectException` with a 401 or 403 status.
- Full existing `realtime_integration_test.dart` suite passes for both protocol versions (subscribe, broadcast, presence, postgres changes, heartbeats), plus `socket_test`, `mock_test`, and `websocket_io_test`.
- Verified end to end against a hosted project: `message`, `hint`, and `sb-error-code` are surfaced as shown above.

## Commit history

The first commit implements a native-only alternative that reads the body during a manual `HttpClient` handshake. The second commit reverts it in favor of the cross-platform follow-up fetch, which is the net content of this PR. The first commit is kept only so the two approaches can be compared, and can be squashed away on merge.

## Open items

- [ ] Compliance symbols: `RealtimeConnectException` is a new public symbol; the `sdk-compliance.yaml` check needs a canonical capability ID from `supabase/sdk` to register it against.
- [ ] Confirm the one extra request on connection failure is acceptable, since it also fires during reconnect backoff.

### DIFF
--- a/packages/realtime_client/lib/realtime_client.dart
+++ b/packages/realtime_client/lib/realtime_client.dart
@@ -11,5 +11,6 @@ export 'src/realtime_channel.dart';
 export 'src/realtime_client.dart';
 export 'src/realtime_presence.dart';
 export 'src/transformers.dart' hide getEnrichedPayload, getPayloadRecords;
+export 'src/websocket/realtime_connect_exception.dart';
 export 'src/types.dart'
     hide ToType, PostgresChangeEventMethods, ChannelFilter, RealtimeListenTypes;
--- a/packages/realtime_client/lib/src/realtime_client.dart
+++ b/packages/realtime_client/lib/src/realtime_client.dart
@@ -283,7 +283,16 @@ class RealtimeClient {
         if (connState != SocketStates.disconnected &&
             connState != SocketStates.disconnecting) {
           connState = SocketStates.closed;
-          _onConnError(error);
+          // A rejected WebSocket upgrade only exposes a status code to the
+          // client (and not even that on web), because the WebSocket API
+          // discards the handshake response body. Re-request the same endpoint
+          // over HTTP to recover the server's error body (for example the
+          // "Invalid API key" message and hint) and surface it.
+          final enrichedError = await _fetchConnectionError(error);
+          if (conn != localConn) {
+            return;
+          }
+          _onConnError(enrichedError);
           reconnectTimer.scheduleTimeout();
         }
         return;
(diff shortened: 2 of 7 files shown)


mandarini (approved):

Vinzent03:
In its current implementation it retries for every kind of exception, but I think we should make this retry more specific. The example in your issue description is a [WebSocketChannelException](https://pub.dev/documentation/web_socket_channel/latest/web_socket_channel/WebSocketChannelException-class.html), which exposes an `inner` exception, which in your example is a [WebSocketException](https://api.flutter.dev/flutter/dart-io/WebSocketException-class.html). So I think we should check whether these conditions are the case.
As a counterexample when I have no internet connection I get this error:

`WebSocketChannelException: SocketException: Failed host lookup: 'xxxx.supabase.co' (OS error: No address associated with hostname, errno = 7)`
So the inner one is a [SocketException](https://api.flutter.dev/flutter/dart-io/SocketException-class.html).
And retrying the endpoint with http only would only delay the error reporting not produce any new insights.

This would also be a breaking change as the previous exception is no longer thrown. So I would check whether we can extend the WebSocketException to be compatible with existing exception checks.

spydon:
> In its current implementation it retries for every kind of exception, but I think we should make this retry more specific. The example in your issue description is a [WebSocketChannelException](https://pub.dev/documentation/web_socket_channel/latest/web_socket_channel/WebSocketChannelException-class.html), which exposes an `inner` exception, which in your example is a [WebSocketException](https://api.flutter.dev/flutter/dart-io/WebSocketException-class.html). So I think we should check whether these conditions are the case. As a counterexample when I have no internet connection I get this error:
> 
> `WebSocketChannelException: SocketException: Failed host lookup: 'xxxx.supabase.co' (OS error: No address associated with hostname, errno = 7)` So the inner one is a [SocketException](https://api.flutter.dev/flutter/dart-io/SocketException-class.html). And retrying the endpoint with http only would only delay the error reporting not produce any new insights.
> 
> This would also be a breaking change as the previous exception is no longer thrown. So I would check whether we can extend the WebSocketException to be compatible with existing exception checks.

Hmm, after further research I think we should just close this one, the native users will already be able to get the status code and that ought to be enough for them to figure out what is wrong. Users on web don't get any information, even if we keep the GET, and there is nothing we can do about that due to the underlying that the browser's WebSocket API deliberately hides it.

We should really fix a list of exceptions we shouldn't retry though, but that's a separate issue.

PR #1640, opened 2026-08-04
refactor(supabase_common): share the local stack test configuration

spydon:
Follow-up to #1639.

The URL, ports and API keys of the local Supabase CLI stack were copied into every test suite that talks to it. The service role JWT alone existed in five places, port 54421 in more than twenty.

- Adds `package:supabase_common/testing.dart`, holding the host, ports, per-service URLs, the mail server address, the anon and service role keys and the JWT secret. Every package already depends on `supabase_common`, so no pubspec changes.
- Annotates the constants with `@visibleForTesting`, so they are usable from any package's test directory but rejected in library code, which CI catches through `dart analyze --fatal-infos`.
- Points postgrest, realtime_client, storage_client, gotrue and supabase at them directly. Only values that are actually derived stay local: `apiHeaders`, `resetAuthDataUrl` and the Postgres endpoint record.
- Removes `packages/supabase/test/utils/local_stack.dart`, which #1639 had added.
- Simplifies the verify link pattern in the gotrue tests, which no longer needs double escaping.

No hardcoded stack address or key is left outside the shared library.

## Summary by CodeRabbit

* **Tests**
  * Standardized integration tests across Auth, REST, Realtime, Storage, and database services to use shared local stack settings.
  * Improved portability across configurable environments.
  * Updated email-link matching and authentication fixture setup.
  * Removed duplicated local URLs and credentials from test configurations.
* **Chores**
  * Added centralized testing utilities for service endpoints, credentials, mail, and database settings.


Tr00d (approved):

Vinzent03 (approved):
I like that most of them should be usable by normal package users as well for their own testing suites. Unsure if that should/could be documented somewhere such that users are aware of this.

spydon:
> I like that most of them should be usable by normal package users as well for their own testing suites. Unsure if that should/could be documented somewhere such that users are aware of this.

Right now the package is marked as internal, but later we can for sure document it and open it up. :)

Issue #1572, opened 2026-07-09
Extract shared code from packages into a supabase_common package


spydon:
## Background

While auditing dependencies in the monorepo we found several pieces of code that are duplicated or near-duplicated across packages. Some duplication was even introduced by the recent dependency-removal work (e.g. the in-house `retry` helper now lives identically in two packages). This issue tracks consolidating that shared code into a single internal (but published) `supabase_common` package.

Packages under `packages/`: `functions_client`, `gotrue`, `postgrest`, `realtime_client`, `storage_client`, `supabase`, `supabase_flutter`, `yet_another_json_isolate`.

Linear: [SDK-1281](https://linear.app/supabase/issue/SDK-1281/extract-shared-code-from-supabase-flutter-packages-into-a-supabase)

## Goal

Create a `supabase_common` package and move genuinely shared code into it, starting with the lowest-risk, highest-confidence items and deferring anything that would break public API.

## Tier 1 — identical / near-identical, lift-and-shift (no public-API risk)

- [ ] **`retry` / `RetryOptions`** — byte-for-byte identical in `packages/gotrue/lib/src/retry.dart` and `packages/storage_client/lib/src/retry.dart`. Move to one shared copy.
- [ ] **`ToSnakeCase` extension on `Enum`** — byte-for-byte identical in `packages/gotrue/lib/src/types/auth_response.dart` (~line 106) and `packages/storage_client/lib/src/types.dart` (~line 469).
- [ ] **Platform detection** — `platform_io.dart` / `platform_stub.dart` plus the conditional-import trick are near-identical between `packages/supabase/lib/src/` and `packages/supabase_flutter/lib/src/` (flutter only adds `isRunningInFlutterTest`). Move with the extra flag preserved.
- [ ] **Replay subject** — `_ReplaySubject` in `packages/gotrue/lib/src/gotrue_client.dart` (~line 1708) and `_BehaviorSubject` in `packages/supabase/lib/src/supabase_stream_builder.dart` (~line 455) reimplement the same broadcast + latest-event-replay behavior. Unify into one `ReplaySubject` supporting `sync`, `onListen`/`onCancel`, and `addStream`.

## Tier 2 — same shape, unify with a small parameter (low risk, some reconciliation)

[...]

Issue #1550, opened 2026-07-08
Make FunctionException a sealed class for v3


spydon:
In #1545 the `FunctionsFetchException`, `FunctionsRelayException`, and `FunctionsHttpException` subtypes were added under `FunctionException` in a non-breaking way (plain inheritance), so existing `catch (FunctionException)` blocks keep working.

For v3 we should make `FunctionException` a `sealed` class so consumers get a compile-checked exhaustive `switch` over the failure mode:

```dart
try {
  await supabase.functions.invoke('hello');
} on FunctionException catch (error) {
  final message = switch (error) {
    FunctionsFetchException() => 'Network failure',
    FunctionsRelayException() => 'Relay error',
    FunctionsHttpException()  => 'Function returned ${error.status}',
  };
}
```

### Why this is breaking (hence v3)

`sealed` implies `abstract`, so:
- The public `const FunctionException(...)` constructor can no longer be used by external code to instantiate a bare `FunctionException` (e.g. in tests/mocks).
- External packages can no longer extend or implement it.

A non-breaking intermediate sealed base does not help: as long as a bare `FunctionException` remains constructable, any switch must allow for one, so the three subtypes can never form the closed set the compiler checks for exhaustiveness. Real three-way exhaustiveness requires `FunctionException` itself to be `sealed`.

### Where

- `packages/functions_client/lib/src/types.dart`

hiSandog:
Sealing the hierarchy gives useful exhaustiveness, but the v3 migration should provide a replacement for tests that currently instantiate or subclass `FunctionException`. Public const constructors for each concrete subtype, with stable fields for cause/status/response, would let mocks create realistic failures without reopening the hierarchy. A compile test for the exhaustive switch plus serialization/string-output compatibility would make the breaking change predictable across the functions client and downstream packages.

spydon:
@hiSandog I don't see what the point would be to subclass or instantiate `FunctionException`.
The only thing a mock would no longer be able to do is to construct a bare, typeless FunctionException. But that's an anti-pattern the seal is deliberately encouraging you away from: if you're testing error handling, you want a specific failure mode, not an abstract stand-in.

PR #1644, opened 2026-08-05
refactor!: split every service exception into a client and an api base

spydon:
Tier 3 of #1572, first of four PRs. Under the v3 umbrella #1278.

## What

`AuthException`, `StorageException`, `PostgrestException` and `FunctionException` each reimplemented the same message plus status shape under different field names and types. They now share one hierarchy in `supabase_common`:

```dart
abstract class SupabaseException implements Exception {
  final String message;
  final String? errorCode;
}

mixin SupabaseApiException on SupabaseException {
  int get statusCode;
}
```

A plain `SupabaseException` is a failure the client raised on its own, before or without a request. A `SupabaseApiException` is a failure a service reported, so `statusCode` is a non-nullable `int` that exists exactly when a service answered, and the layer a failure came from is visible from its type alone. Both are re-exported from every package, so one catch handles a failure from any service:

```dart
try {
  await supabase.from('countries').select();
} on SupabaseApiException catch (error) {
  print('${error.statusCode}: ${error.message}');
} on SupabaseException catch (error) {
  print(error.message);
}
```

The two level split came out of [@Vinzent03's review](https://github.com/supabase/supabase-flutter/pull/1644#issuecomment-5255301142). A single base handed every client raised failure a `statusCode` and an `errorCode` that were always null, and storage went further than that: a request that never reached the server produced `StorageException(statusCode: 'ClientException')`, the exception's own type name in the status field.

`SupabaseApiException` is a mixin rather than a second base class because the two axes are orthogonal. `AuthApiException` has to be an `AuthException`, since that is what callers catch, and also the cross service type for "a service answered". Dart gives you one superclass, so one of the two has to be a mixin or an interface, and neither can hold state. That is why each API exception declares `statusCode` itself.

## The hierarchy

| Package | Client raised | Service reported |
| --- | --- | --- |
| `gotrue` | `AuthException`, `AuthPKCEGrantCodeExchangeError`, `AuthSessionMissingException`, `AuthInvalidJwtException`, `AuthUnknownException`, `AuthRetryableFetchException` | `AuthApiException`, `AuthWeakPasswordException`, `AuthRetryableApiException` (new) |
| `postgrest` | none, every failure is a response | `PostgrestApiException` |
| `storage_client` | `StorageException` | `StorageApiException` (new) |
| `functions_client` | `FunctionException`, `FunctionsFetchException` | `FunctionsApiException`, `FunctionsRelayException` |

`PostgrestApiException` mixes `SupabaseApiException` in directly instead of gaining a subclass, since every Postgrest failure comes from a response. It keeps the `Api` part of the name so it lines up with the other packages, which leaves the plain `PostgrestException` name free for a client raised base if one is ever needed.

## Breaking changes

| Before | After |
| --- | --- |
| `AuthException.statusCode` (`String?`) | `AuthApiException.statusCode` (`int`, required) |
| `AuthException.code` | `AuthException.errorCode` |
| `AuthSessionMissingException.statusCode == '400'` | no status, `errorCode` is `session_missing` |
| `AuthInvalidJwtException.statusCode == '400'` | no status, `errorCode` is `invalid_jwt` |
| `AuthWeakPasswordException extends AuthException` | `extends AuthApiException` |
| `AuthRetryableFetchException.statusCode` | gone; a 5xx is an `AuthRetryableApiException`, which carries it |
| `AuthUnknownException.statusCode` | gone; read it from `originalError` |
| `StorageException.statusCode` (`String?`) | `StorageApiException.statusCode` (`int`, required) |
| `StorageException.error` | `errorCode` |
| `StorageException.fromJson(json, '404')` | `StorageApiException.fromJson(json, 404)` |
| `PostgrestException` | `PostgrestApiException` |
| `PostgrestException.code` (PostgREST code, or the HTTP status when the body was not JSON) | `errorCode` (PostgREST or PostgreSQL code only) and `statusCode` (HTTP status, non-nullable) |
| `PostgrestException.fromJson(json, code: 409)` | `PostgrestApiException.fromJson(json, statusCode: 409)` |
| `PostgrestException.toJson()` key `code` | keys `statusCode` and `errorCode` |
| `FunctionsHttpException` | `FunctionsApiException` |
| `FunctionException.status` (`int`) | `FunctionsApiException.statusCode` (`int`, required) |
| `FunctionException.reasonPhrase` | folded into `message` |
| `FunctionsFetchException.status == 0` | no status at all |
| `FunctionResponse.status` | `FunctionResponse.statusCode` |

Notes on the less mechanical ones:

- **Postgrest no longer overloads `code`.** It used to stuff the HTTP status into `code` when the error body was not JSON, so `code` was sometimes `PGRST116` and sometimes `409`. The status now has its own field and `errorCode` is only ever a PostgREST or PostgreSQL code. A duplicate key violation reads as `statusCode: 409, errorCode: '23505'`.
- **Two auth exceptions stop fabricating a status.** `AuthSessionMissingException` and `AuthInvalidJwtException` are raised by the client, never by the service, so the `400` they reported was invented. They report an error code instead.
- **`AuthRetryableFetchException` split in two.** It was thrown both for a transport failure, which has no status, and for a 5xx, which has one. It now covers only the transport case, and `AuthRetryableApiException` extends it for a 5xx. `e is AuthRetryableFetchException` still catches both, so the retry and refresh paths in `GoTrueClient` are unchanged.
- **Functions gained a message.** `FunctionException` had no message, only `status`, `details` and `reasonPhrase`. The response's reason phrase becomes the message, and when the response carries none, as over HTTP/2, each subtype falls back to its own default, matching supabase-js: `'Failed to send a request to the Edge Function'`, `'Relay error invoking the Edge Function'`, `'Edge Function returned a non-2xx status code'`. The response body is still in `details`.
- **`FunctionsHttpException` is gone.** It and `FunctionsApiException` both meant "the Edge Function answered with a non-2xx status", which is the definition of an api exception, so `FunctionsApiException` took over the name and the default message. `FunctionsRelayException` stays, because a relay error means the function may never have run.
- **`FunctionResponse.status` is renamed too.** With the exceptions in that package reporting `statusCode`, the successful response was the only thing left calling it `status`. It also lines up with `http.Response.statusCode`.
- **`AuthException` compares the runtime type.** `statusCode` moved to the subclasses, so equality had to move with it. The base compares the concrete type, message and error code, and the API subclasses add the status, which keeps equality symmetric across the hierarchy.

`getSessionFromUrl` also had to change how it reads an error callback: `error_code` holds either a numeric status (older links) or a code such as `otp_expired`. The numeric form now throws an `AuthApiException` and anything else a plain `AuthException` whose `errorCode` falls back to the `error` parameter. Both shapes are covered by tests.

Every auth exception subclass used to repeat the same `toString`; the shared classes now print the concrete runtime type, so only the subtypes with extra fields (`originalError`, `reasons`, `details`, `hint`) override it.

## A bug this surfaces

`Fetch._handleError` in storage cast the decoded error body to `Map<String, dynamic>` inside a `try`/`on FormatException`. A body that parses as JSON but is not an object, for example the `["upstream connect error"]` a gateway can return, throws a `TypeError` on that cast, which the `on FormatException` does not catch, so it escaped instead of surfacing as a `StorageException`. Fixed here, with a regression test; #1647 later replaces the inline guard with the shared `tryDecodeJsonObject` helper.

## Testing

`dart analyze`, `dcm analyze` and `dart format` are clean. Full test suites pass for `gotrue`, `postgrest`, `storage_client`, `functions_client`, `realtime_client`, `supabase`, `supabase_common` (against the local stack) and `supabase_flutter`, plus the examples analyzer. The capability matrix symbol, drift and schema checks pass with the new symbols registered.

`auth_exception_test.dart` loses more lines than it gains, which is deliberate. It had grown a large number of tests that could not fail: constructor round trips reading back the values just passed in, `isA` checks the compiler already guarantees, a whole `Error handling scenarios` group restating other tests with different string literals, and one test named after equality that compared field by field and never invoked `==`. What is left covers behaviour that can break, and its `toString` assertions are exact matches rather than `contains`. The real coverage for this change is in `fetch_test.dart`, which drives HTTP responses through the client and asserts the exception that comes out.

## Migration

`MIGRATION.md` gains a section for the new hierarchy: the shape, the full rename table, and the one case that needs a code change rather than a rename, which is reading a status off a per-service base.

## Out of scope

`RealtimeSubscribeException` has no `message` at all, only a `RealtimeSubscribeStatus` and `details`, and it reports a channel subscription outcome rather than a request failure. It does not fit the base without inventing a message, so it keeps its own shape.

The sealed `IcebergException` is a closer fit and is worth migrating, in its own PR: it already has `message`, a non-nullable `statusCode`, a `type` that maps onto `errorCode`, and `details`. It also carries the same `statusCode == 0` sentinel for a network failure that this PR removed from `FunctionsFetchException`, so it wants the same client and api split rather than a straight rename.

Making `FunctionException` sealed is tracked separately in #1550.


Tr00d (approved):

Vinzent03:
We have cases where an exception is thrown with only a message for some bad state in the client e.g. https://github.com/supabase/supabase-flutter/blob/bbf9b74bb8878057d56caade0c0eea0c0d81f46e/packages/gotrue/lib/src/gotrue_client.dart#L1069-L1079

So these cases have no status or error code. Currently they are null in these cases. Error codes and status codes are only present for http/supabase api calls. So I'm wondering if it makes sense to separate those api exceptions from these client only exceptions and maybe even have a class with status code being non-nullable. Such that we have a `SupabaseApiException` with error and status code and `SupabaseException` with message only. This could make it easier to know on first sight from which layer or part an exception is coming from.

spydon:
> We have cases where an exception is thrown with only a message for some bad state in the client e.g.
> 
> https://github.com/supabase/supabase-flutter/blob/bbf9b74bb8878057d56caade0c0eea0c0d81f46e/packages/gotrue/lib/src/gotrue_client.dart#L1069-L1079
> 
> So these cases have no status or error code. Currently they are null in these cases. Error codes and status codes are only present for http/supabase api calls. So I'm wondering if it makes sense to separate those api exceptions from these client only exceptions and maybe even have a class with status code being non-nullable. Such that we have a `SupabaseApiException` with error and status code and `SupabaseException` with message only. This could make it easier to know on first sight from which layer or part an exception is coming from.

Sounds like a good idea!

spydon:
@Vinzent03 what do you think about this solution?

```dart
abstract class SupabaseException implements Exception {
  final String message;
  final String? errorCode;

  const SupabaseException(this.message, {this.errorCode});
}

mixin SupabaseApiException on SupabaseException {
  int get statusCode;
}
```

Vinzent03:
I think I really like the new style. But some small things:
- What is the real difference between `FunctionsApiException` and a `FunctionsHttpException`. I can see the reason for the relay exception, but I'm wondering if the http on can just be the `FunctionsApiException`. Primarily because it is not very aligned with the other exception classes from the other packages. The definition of an api exception is that it comes from a http response with no 2xx status code so I feel like the api exception one is enough.
- I'm thinking whether the `PostgrestException` should be called `PostgrestApiException` as well like the others such that it is more aligned and we are freer to introduce an `PostgrestException` in the future.

Also may I ask for the reason of excluding the realtimesubscribe exception and iceberg here? Haven't looked into it really, but just wondering.

spydon:
> I think I really like the new style. But some small things:
> 
>     * What is the real difference between `FunctionsApiException` and a `FunctionsHttpException`. I can see the reason for the relay exception, but I'm wondering if the http on can just be the `FunctionsApiException`. Primarily because it is not very aligned with the other exception classes from the other packages. The definition of an api exception is that it comes from a http response with no 2xx status code so I feel like the api exception one is enough.
> 
>     * I'm thinking whether the `PostgrestException` should be called `PostgrestApiException` as well like the others such that it is more aligned and we are freer to introduce an `PostgrestException` in the future.

Sounds like good suggestions, I'll fix those up!

> Also may I ask for the reason of excluding the realtimesubscribe exception and iceberg here? Haven't looked into it really, but just wondering.

Because they have zero overlap with the other exception structure.

EDIT:
Indeed much cleaner with just `FunctionsApiException`!

EDIT2:
iceberg seems like it would be worth doing after all, doing that in stacked follow-up PR though.

Vinzent03 (approved):

PR #1658, opened 2026-08-06
fix!: make order() default to ascending

### LINKED ISSUE #1638: v3: flip order() default to ascending: true
## Summary

`PostgrestTransformBuilder.order()` defaults to **descending**:

```dart
PostgrestTransformBuilder<T> order(
  String column, {
  bool ascending = false,
  bool nullsFirst = false,
  String? referencedTable,
})
```

So `.order('created_at')` returns newest rows first. This diverges from the two things a developer is most likely to expect:

- SQL `ORDER BY`, which defaults to `ASC`
- `postgrest-js`, where `order()` destructures `{ ascending = true }`

The failure mode is silent. The query succeeds, the row count is right, the types are right, the analyzer says nothing, and only the order is wrong. It survives unit tests that assert on set membership rather than sequence, which is what was reported in #1637.

## How the default got here

It was never a deliberate choice for Dart. It is an inherited copy of a JavaScript signature that stopped existing ten days later:

- **2020-09-10**, `supabase/postgrest-dart` commit `7a8f8b7` introduced the first `order()` as a port of the pre-TypeScript `postgrest-js`, which then had `order(columnName, ascending = false, nullsFirst = false)`. The self-contradictory doc line came across verbatim: `@param {bool} ascending True for descending results, false by default.`
- **2020-09-20**, `postgrest-js` commit `2960d09` ("feat: Transition to TypeScript") rewrote the signature to `{ ascending = true, ... }`. Dart never followed, and this is where the divergence begins.
- **2020-10-13**, `postgrest-dart` commit `37eb9fc` moved the options map to named parameters and carried `ascending = false` over unchanged.
- **2023-05-11**, #461 consolidated the package into this monorepo, still unchanged.

## Proposed change (v3)

- `packages/postgrest/lib/src/postgrest_transform_builder.dart`: `bool ascending = true` in `order()`.
- `packages/supabase/lib/src/supabase_stream_builder.dart`: `SupabaseStreamBuilder.order()` has the same `bool ascending = false` default and should be flipped in the same change, otherwise `.stream()` and `.select()` disagree on what an unqualified `.order()` means.

`nullsFirst` stays at `false`. That already matches `postgrest-js` and PostgREST, so there is no divergence to fix.

Update the doc comments and both packages' tests, and the migration guide.

## Why breaking

Every existing call that omits `ascending` silently reverses its result order. There is no compile error, no deprecation window that can surface it, and no runtime failure. That combination is why it needs a major version and a prominent entry in the migration guide rather than a quiet fix.

## Migration path

Callers who want the current behavior add `ascending: false` explicitly. This is worth calling out in the migration guide as something that can be done **before** upgrading, since it is a no-op against the current default and makes the upgrade a non-event for that call site.

Until v3, #1637 documents the current default in the doc comment so it stops being a surprise.

## Context

Tracked under the v3 umbrella (#1278).

spydon:
Closes #1638

Breaking change.

`order()` defaults `ascending` to `false` in two places, so omitting the parameter sorts descending:

- `PostgrestTransformBuilder.order()`
- `SupabaseStreamBuilder.order()`

`.order('created_at')` therefore returns newest rows first, which disagrees with SQL's `ORDER BY` (defaults to `ASC`) and with `postgrest-js` (`{ ascending = true }`). The failure is silent: the query succeeds, the row count and types are right, only the order is wrong.

The default was never chosen for Dart. `postgrest-dart` `7a8f8b7` (2020-09-10) ported the first `order()` from the pre-TypeScript `postgrest-js`, which then had `order(columnName, ascending = false, nullsFirst = false)`. `postgrest-js` `2960d09` moved to `{ ascending = true }` ten days later and Dart never followed. #1637 documented the current behavior in the meantime.

Two examples in this repository were already wrong because of it, which is the clearest evidence that the default surprises people:

- `examples/realtime_room/lib/room_repository.dart` says "SELECT the existing messages once, oldest first" and returned them newest first.
- `examples/database_crud/lib/tasks_repository.dart` says "ordered alphabetically by name" and returned projects in reverse alphabetical order.

Both are now correct without touching their query code.

## What is the new behavior?

- `PostgrestTransformBuilder.order()`: `bool ascending = true`.
- `SupabaseStreamBuilder.order()`: `bool ascending = true`, so `.stream()` and `.select()` agree on what an unqualified `.order()` means.
- Doc comments on both state the default and drop the note about diverging from SQL and `postgrest-js`, since there is no longer a divergence.

`nullsFirst` is unchanged at `false`. That already matched `postgrest-js` and PostgREST.

## Migration

`MIGRATION.md` gets an `order() now sorts ascending by default` section under v2 to v3, covering both call sites, why the analyzer will not catch it, and how to keep the old behavior:

```dart
await supabase.from('messages').select().order('created_at', ascending: false);
```

The section also points out that `ascending: false` already means descending on v2, so consumers can add it before upgrading and leave this change out of the upgrade itself.

## Tests

- New `packages/postgrest/test/order_default_test.dart` asserts the generated `order` query parameter without a backend: the default, explicit `ascending: false`, `nullsFirst: true`, a referenced table, and that chained `order()` calls keep their own direction.
- `transforms_test.dart` gains `order defaults to ascending` and `order descending` against the local stack.
- `mock_test.dart` gains `order defaults to ascending` for the stream builder, and the mock server now answers `order=id.asc.nullslast`.
- Tests that were exercising the old implicit default (multi-column order, order with filters, referenced-table order, embedded order, stream order with limit) now pass `ascending: false` explicitly, so their coverage of descending order is preserved rather than silently reinterpreted.

`melos analyze` and `melos format` are clean. `packages/postgrest` (196 tests) and `packages/supabase` (136 tests) pass in full against a local Supabase stack.

## Additional context

Part of the v3 umbrella (#1278).

## Summary by CodeRabbit

* **New Features**
  * Database queries and real-time streams now default to ascending order.
  * Descending order remains available with `ascending: false`.
  * Ordering continues to support null placement, referenced data, multiple columns, and filters.

* **Documentation**
  * Updated ordering guidance, examples, and migration notes to reflect the new default.

* **Tests**
  * Expanded coverage for default, ascending, descending, nested, referenced-table, filtered, and stream ordering.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -112,3 +112,36 @@ await catalog.loadTableResult(
   const LoadTableOptions(snapshots: TableSnapshotScope.refs),
 );
 ```
+
+### `order()` now sorts ascending by default
+
+`PostgrestTransformBuilder.order()` and `SupabaseStreamBuilder.order()` defaulted `ascending` to
+`false`, so a call that left the parameter out sorted descending. That was the opposite of SQL's
+`ORDER BY` and of the other Supabase client libraries. Both now default to `true`.
+
+Like the `connectionState` change above, this one does not produce a compile error. The call still
+compiles and still returns the same rows, only in the reverse order, so it is worth checking every
+`.order()` call that does not pass `ascending` explicitly, in both `select()` queries and
+`stream()`.
+
+```dart
+// Before: newest first
+final messages = await supabase.from('messages').select().order('created_at');
+
+// After: oldest first
+final messages = await supabase.from('messages').select().order('created_at');
+```
+
+To keep the previous behaviour, ask for descending order explicitly:
+
+```dart
+final messages = await supabase
+    .from('messages')
+    .select()
+    .order('created_at', ascending: false);
+```
+
+`ascending: false` already means descending on v2, so you can add it to your current code before
+upgrading and leave this change out of the upgrade itself.
+
+`nullsFirst` is unchanged and still defaults to `false`.
--- a/examples/database_crud/lib/tasks_repository.dart
+++ b/examples/database_crud/lib/tasks_repository.dart
@@ -26,7 +26,7 @@ class TasksRepository {
   /// * [search] matches the title case-insensitively (`ilike`).
   /// * [onlyIncomplete] hides finished tasks (`eq`).
   ///
-  /// Results are ordered by priority (highest first) then creation time.
+  /// Results are ordered by priority (highest first) then newest first.
   Future<List<Task>> fetchTasks({
     String? projectId,
     String? search,
@@ -47,7 +47,7 @@ class TasksRepository {
 
     final rows = await query
         .order('priority', ascending: false)
-        .order('created_at');
+        .order('created_at', ascending: false);
     return rows.map(Task.fromJson).toList();
   }
 
--- a/packages/postgrest/lib/src/postgrest_transform_builder.dart
+++ b/packages/postgrest/lib/src/postgrest_transform_builder.dart
@@ -73,14 +73,13 @@ class PostgrestTransformBuilder<T> extends RawPostgrestBuilder<T, T, T> {
 
   /// Orders the result with the specified [column].
   ///
-  /// [ascending] defaults to `false`, so results come back in **descending**
-  /// order unless `ascending: true` is passed. Note that this is the opposite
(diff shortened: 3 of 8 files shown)


Tr00d (approved):

QuintinWillison (approved):
💪

PR #1677, opened 2026-08-10
chore!: give each package's Constants class a package-specific name

spydon:
Stacked on #1673, addressing https://github.com/supabase/supabase-flutter/pull/1673#discussion_r3747503513.

## Why

Six packages declared a class named `Constants`: `functions_client`, `gotrue`, `realtime_client`, `storage_client`, `supabase` and `supabase_flutter`. The capability matrix keys symbols by bare name, so an entry like `Constants.defaultHeaders` could not say which package it meant. The same ambiguity applies to anyone reading the code across packages.

## What changed

Each class is named after its package, and the file declaring it is named to match:

| Package | Class | File |
|---|---|---|
| `functions_client` | `FunctionsConstants` | `src/functions_constants.dart` |
| `gotrue` | `GoTrueConstants` | `src/gotrue_constants.dart` |
| `realtime_client` | `RealtimeConstants` | `src/realtime_constants.dart` |
| `storage_client` | `StorageConstants` | `src/storage_constants.dart` |
| `supabase` | `SupabaseConstants` | `src/supabase_constants.dart` |
| `supabase_flutter` | `SupabaseFlutterConstants` | `src/supabase_flutter_constants.dart` |

`gotrue` and `realtime_client` keep a `constants.dart`, since both also declare enums there. Only the class moved out into its own file, so `constants.dart` now holds exactly the enums and, for `gotrue`, `ApiVersions`. `gotrue.dart` no longer needs `hide Constants` on its export as a result.

`realtime_client` holds the only public class of the six, so it changed the most:

- The `typedef RealtimeConstants = Constants` is gone and the class carries that name directly. `realtime_client.dart` only ever exported the alias, so the name consumers write is unchanged.
- `defaultHeaders`, `defaultHeartbeatIntervalMs` and `wsCloseNormal` are now `@internal`. Nothing outside the package reads them, and the other five packages' default header maps were already internal. This is the breaking part: code outside the package that reached for them will now get `invalid_use_of_internal_member`.
- `defaultTimeout` and `defaultConnectionCloseTimeout` stay public, because `supabase` reads both when filling in unset `RealtimeClientOptions`.

In `sdk-compliance.yaml`, the three now-internal symbols are dropped, `realtime.configuration.heartbeat_interval` loses its `supporting_symbols` list along with them, the duplicated `Constants` and `Constants.defaultTimeout` pair is removed from the top-level list, and `RealtimeConstants.defaultConnectionCloseTimeout` and `RealtimeConstants.defaultTimeout` are registered under the new name.

## Verification

- `dart analyze` clean across all packages and examples.
- Full suites pass against a local Supabase stack, at the `--concurrency=1` the workflow uses: `gotrue` (479), `realtime_client` (205, integration included), `storage_client` (210), `supabase` (134), `supabase_flutter` (65), `functions_client` (48).
- The compliance symbol, drift and schema checks pass locally against this branch's base.

## Summary by CodeRabbit

* **New Features**
  * Realtime configuration constants are now available through the public library interface.

* **Refactor**
  * Standardized configuration naming across authentication, database, functions, realtime, storage, and Flutter integrations.
  * Centralized default headers, timeouts, session settings, and client metadata under clearer service-specific constants.
  * Preserved existing defaults and runtime behavior.

* **Tests**
  * Updated automated coverage to validate the renamed configuration references and existing expected values.

### DIFF
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -2,7 +2,7 @@ import 'dart:async';
 
 import 'package:http/http.dart';
 import 'package:logging/logging.dart';
-import 'package:supabase/src/constants.dart';
+import 'package:supabase/src/supabase_constants.dart';
 import 'package:supabase/src/version.dart';
 import 'package:supabase/supabase.dart';
 import 'package:yet_another_json_isolate/yet_another_json_isolate.dart';
@@ -93,7 +93,7 @@ class SupabaseClient {
   set headers(Map<String, String> newHeaders) {
     _headers.clear();
     _headers.addAll({
-      ...Constants.defaultHeaders,
+      ...SupabaseConstants.defaultHeaders,
       ...newHeaders,
     });
 
@@ -113,7 +113,7 @@ class SupabaseClient {
       auth.headers
         ..clear()
         ..addAll({
-          ...Constants.defaultHeaders,
+          ...SupabaseConstants.defaultHeaders,
           ..._getAuthHeaders(),
           ...headers,
         });
@@ -153,7 +153,7 @@ class SupabaseClient {
        _functionsUrl = '$supabaseUrl/functions/v1',
        _postgrestOptions = postgrestOptions,
        _headers = {
-         ...Constants.defaultHeaders,
+         ...SupabaseConstants.defaultHeaders,
          ...?headers,
        },
        _httpClient = httpClient,
--- a/packages/supabase/lib/src/supabase_constants.dart
+++ b/packages/supabase/lib/src/supabase_constants.dart
@@ -3,7 +3,7 @@ import 'package:supabase_common/supabase_common.dart';
 import 'package:meta/meta.dart';
 
 @internal
-class Constants {
+class SupabaseConstants {
   static final Map<String, String> defaultHeaders = Map.unmodifiable({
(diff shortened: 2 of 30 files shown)


QuintinWillison (commented):
Needing to rename an internal type to fit a downstream, external system smells wrong to me. 🤔 

This raises questions:

1. Should the symbols/identifiers being enumerated in your manifest include the package they're in also, so each package (as is ok in Flutter/Dart) can have its own `Constants` internal?
2. But, actually, should internal symbols/identifiers even be enumerated in a public-facing manifest?

spydon:
> Needing to rename an internal type to fit a downstream, external system smells wrong to me. 🤔 

The PR description is a little bit misdirecting, the main clean-up reason is that we shouldn't have  classes named in a generic fashion like that in the repo, the class names should entail what they are for.

> This raises questions:
> 
> 1. Should the symbols/identifiers being enumerated in your manifest include the package they're in also, so each package (as is ok in Flutter/Dart) can have its own `Constants` internal?

That would be a good improvement, but I don't think we need to do it until it becomes an actual problem.

> 2. But, actually, should internal symbols/identifiers even be enumerated in a public-facing manifest?

They are not.

Vinzent03 (approved):
I like this change.

Tr00d (approved):

PR #1678, opened 2026-08-10
refactor!: share a single HttpMethod enum across the packages

spydon:
## Why

Four different representations of an HTTP method coexisted across the packages:

| Package | Representation | Values | Wire string |
|---|---|---|---|
| `functions_client` | public `enum HttpMethod` | get, post, put, delete, patch | `method.name.toUpperCase()`, open-coded at two call sites |
| `postgrest` | public `enum HttpMethod` | get, **head**, post, put, patch, delete | `String get value => name.toUpperCase()` |
| `gotrue` | `@internal enum RequestMethodType` | get, post, put, patch, delete | none, switches to `http.Client`'s `get` / `post` / … |
| `storage_client` | bare `String` | `'GET'`, `'POST'`, `'PUT'`, `'DELETE'`, `'HEAD'` | the literal itself |

The two public enums shared a name while disagreeing on contents, which forced `supabase.dart` to export postgrest with `hide HttpMethod`, making postgrest's enum unreachable through the umbrella library. It also confused the capability matrix, which keys symbols by bare name: it registered `HttpMethod` for functions and `HttpMethod.value` for postgrest as if they were one type.

The stringly-typed side had its own cost. `storage_client` compared `method != 'GET'` to decide whether to set a JSON content type, in both its `Fetch` helper and its Iceberg REST catalog.

## What changed

One `HttpMethod` in `supabase_common`, taking postgrest's shape: the six methods plus the `value` getter that `functions_client` used to open-code.

Public surface:

- `functions_client` and `postgrest` re-export the shared enum, so callers of either library see no change.
- `supabase.dart` exports postgrest whole again.
- `functions.invocation.method_override` registers `FunctionsClient.invoke`, since the shared enum lives in `supabase_common`, which `.sdk-parse-ignore` excludes from the scanned public API surface. A note records that the Dart enum also offers `head`, which supabase-js does not expose for function invocation.
- The `edge_functions` example's "every HTTP method" test skips `head`, whose response carries no body for the echo function to reflect, and asserts against `method.value`.

Internal adoption, no public API change since both declarations were `@internal`:

- `gotrue`: `RequestMethodType` is gone and its roughly 65 call sites use the shared enum. `GotrueFetch`'s dispatch switch gains a `head` branch to stay exhaustive over six values, wired to `http.Client.head`, and now skips the JSON content type for `head` as well as for `get`. No gotrue call site issues a HEAD today, so that branch is currently unexercised.
- `storage_client`: `Fetch`'s private request helpers and the Iceberg catalog's `_request` take an `HttpMethod` instead of a `String`, the wire strings come from `.value`, and the content-type branch compares enum values. The public wrapper method names (`get`, `post`, `head`, and so on) are unchanged.

Every log line that interpolates a method uses `.value`, so they keep printing `GET` rather than the enum's default `toString`.

The `supabase_common` pins stay at `0.1.2`; `melos version` rewrites dependents' pins at release time.

## Merge order

This targets `main` directly and does not depend on #1673 or #1677, though it came out of the same review thread. #1673 previously registered `HttpMethod` and `HttpMethod.value`, which this PR makes dangling; those registrations have been pruned on that branch, so the two can merge in either order.

## Verification

- `dart analyze` clean across all packages and examples.
- `dcm analyze` reports no issues in any file this PR touches.
- Suites pass against a local Supabase stack, at the `--concurrency=1` the workflow uses: `gotrue` (479), `storage_client` (210), `postgrest` (196), `functions_client` (48).
- The compliance symbol, drift and schema checks pass locally with `main` as the base.

## Summary by CodeRabbit

- **New Features**
  - Added a shared `HttpMethod` API covering GET, HEAD, POST, PUT, PATCH, and DELETE.
  - Made `HttpMethod` available through the relevant client libraries.
  - Added explicit HEAD request support where applicable.

- **Bug Fixes**
  - Improved HTTP method serialization for requests.
  - Corrected request content handling for GET and HEAD calls.
  - Updated integration coverage to accurately validate supported methods.

### DIFF
--- a/examples/edge_functions/integration_test/invoke_test.dart
+++ b/examples/edge_functions/integration_test/invoke_test.dart
@@ -57,7 +57,12 @@ void main() {
     });
 
     testWidgets('every HTTP method reaches the function', (_) async {
-      for (final method in HttpMethod.values) {
+      // HEAD responses carry no body, so the echo function cannot report back
+      // which method it saw.
+      final methods = HttpMethod.values.where(
+        (method) => method != HttpMethod.head,
+      );
+      for (final method in methods) {
         final response = await functions.invoke(
           'echo',
           method: method,
@@ -66,7 +71,7 @@ void main() {
               ? null
               : {'value': 1},
         );
-        expect(response.data['method'], method.name.toUpperCase());
+        expect(response.data['method'], method.value);
       }
     });
 
--- a/packages/functions_client/lib/functions_client.dart
+++ b/packages/functions_client/lib/functions_client.dart
@@ -3,6 +3,7 @@ library;
 
 export 'package:http/http.dart'
     show ByteStream, MultipartFile, RequestAbortedException;
+export 'package:supabase_common/supabase_common.dart' show HttpMethod;
 
 export 'src/functions_client.dart';
 export 'src/types.dart';
--- a/packages/functions_client/lib/src/functions_client.dart
+++ b/packages/functions_client/lib/src/functions_client.dart
@@ -169,7 +169,7 @@ class FunctionsClient {
 
       request =
           http.AbortableMultipartRequest(
-              method.name.toUpperCase(),
+              method.value,
               uri,
               abortTrigger: abortSignal,
             )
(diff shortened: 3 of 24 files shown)


QuintinWillison (approved):
My Dart's not good enough to comment fully on correctness there, but fully support the reduction of repetition! 💪

PR #1683, opened 2026-08-10
refactor!: apply Dart 3 class modifiers to the core public types

spydon:
Closes SDK-819.

Most public classes in the SDK were plain `class`, so consumers could extend or implement any of them, including internal plumbing. This declares the intent explicitly with Dart 3 class modifiers.

### `interface class` — implementable, not extendable

The client entry points and the request builder base. `interface class` closes off `extends` while leaving `implements` open, so the standard mocking idiom keeps compiling:

```dart
// Still compiles
class MockSupabaseClient extends Mock implements SupabaseClient {}

// No longer compiles
class MyClient extends SupabaseClient { /* ... */ }
```

`SupabaseClient`, `GoTrueClient`, `PostgrestClient`, `PostgrestBuilder`, `RealtimeClient`, `FunctionsClient`, `SupabaseStorageClient`, `StorageFileApi`.

The ticket proposed `final class` for these. `final` also blocks `implements`, which would break every consumer test suite that mocks a client with no escape hatch, so `interface` is the narrower break that still delivers "not designed for extension".

### `abstract interface class` — the two consumer extension points

`LocalStorage` and `GotrueAsyncStorage`. Both declare only abstract members, so callers change one word:

```dart
- class MyLocalStorage extends LocalStorage { /* ... */ }
+ class MyLocalStorage implements LocalStorage { /* ... */ }
```

### `final class` — fully closed

`AuthState`, plus the internal `GotrueFetch`, `Fetch` and `SupabaseAuth` helpers.

## Notes on the rest of the ticket's proposal

- **The `AuthException` hierarchy is deliberately untouched.** The ticket asks for `base class AuthException` with `final` subclasses, but the exceptions are being restructured by a separate chain of PRs, so the modifiers belong there rather than here where they would just conflict.
- `AuthChangeEvent` is already an `enum`, so the proposed `sealed class` treatment is moot; exhaustive switches over auth events are unchanged. `AuthState` is a single class rather than a hierarchy, so it is `final` rather than `sealed`.
- `StorageBucketApi` is not exported from `package:storage_client`, so consumers cannot name it and its modifier would have no public effect. Closing it would also force `SupabaseStorageClient` to be `base`/`final` and lose mockability, so it stays a plain class.
- The concrete postgrest builders (`PostgrestQueryBuilder` and friends) stay open. `PostgrestQueryBuilder` is extended across package boundaries by `SupabaseQueryBuilder`, and marking the chain `base`/`final` would make the whole `from(...)` chain unmockable.

## Test changes

Two realtime test stubs extended `RealtimeClient` and had to go:

- The setAuth error-handling tests in `channel_test.dart` now use a mocktail mock of the socket, which is exactly the escape hatch `interface class` preserves.
- The `remove` and `setAuth` tests in `socket_test.dart` register their mock channels on `RealtimeClient.channels` directly rather than overriding `channel()` to hand them back.

The `LocalStorage` and `GotrueAsyncStorage` test stubs switched from `extends` to `implements`.

## Verification

- `dart analyze packages`: no issues
- `dart format`: no changes
- Full test suites pass for gotrue (479), postgrest (196), realtime_client (205), storage_client (210), supabase (134), functions_client (48) and supabase_flutter (65), with the local Supabase stack running for the backend-dependent ones
- The extracted public symbol set is byte-identical to `main` (1754 symbols, no additions or removals), so `sdk-compliance.yaml` needs no changes

`MIGRATION.md` documents each modifier under the v2 to v3 section.

## Summary by CodeRabbit

* **Documentation**
  * Added migration guidance for Dart 3 class modifiers and updated inheritance examples.

* **API Updates**
  * Clarified which public classes can be extended, implemented, or neither through interface, abstract interface, and final declarations.
  * Preserved existing members, method signatures, and runtime behavior.

* **Tests**
  * Updated storage and realtime test implementations to follow the new class-modifier rules.
  * Refined realtime authentication and channel test coverage using mocks.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -326,3 +326,69 @@ unaffected. For `functions.invoke`, two things changed:
   (3 to 5). This one is not a compile error, so replace any persisted `index` with `name`.
 
 The enum also exposes `value`, the uppercase wire form, in place of `method.name.toUpperCase()`.
+
+### Core types declare their subtyping intent with class modifiers
+
+The core public types now carry Dart 3 class modifiers that state whether they are meant to be
+extended, implemented, or neither. Nothing about their behaviour changed, but subtyping them in a
+way the modifier disallows is now a compile error.
+
+The clients and the request builders are `interface class`, so they can still be implemented (this
+is what `mockito` and `mocktail` need) but no longer extended:
+
+| Type | Modifier |
+| --- | --- |
+| `SupabaseClient` | `interface class` |
+| `GoTrueClient` | `interface class` |
+| `PostgrestClient` | `interface class` |
+| `PostgrestBuilder` | `interface class` |
+| `RealtimeClient` | `interface class` |
+| `FunctionsClient` | `interface class` |
+| `SupabaseStorageClient` | `interface class` |
+| `StorageFileApi` | `interface class` |
+
+If you were extending one of these to change its behaviour, wrap it instead:
+
+```dart
+// Before
+class LoggingSupabaseClient extends SupabaseClient {
+  LoggingSupabaseClient(super.url, super.key);
+}
+
+// After
+class LoggingSupabaseClient {
+  LoggingSupabaseClient(this._inner);
+
+  final SupabaseClient _inner;
+
+  SupabaseQueryBuilder from(String table) {
+    print('from($table)');
+    return _inner.from(table);
+  }
+}
+```
+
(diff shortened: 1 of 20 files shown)


Tr00d (approved):

spydon:
Decided against this, I think it's overengineering for very little win.

PR #1706, opened 2026-08-13
refactor(realtime)!: replace listener callbacks with streams

### LINKED ISSUE #1520: Move to streams instead of callbacks where suitable for v3
I was just reviewing it in this very moment. I really like using a stream instead of a callback mechanism. This feels way more Dart idiomatic. (I thought about switching to streams for other parts of the realtime_client as well, maybe something to consider for v3. Makes removing a listener way easier)

_Originally posted by @Vinzent03 in https://github.com/supabase/supabase-flutter/pull/1517#pullrequestreview-4617492130_

spydon:
## What

Converts the callback-based listener APIs in `realtime_client` to broadcast streams, following the precedent set by `RealtimeClient.onHeartbeat` (#1517).

### `RealtimeClient`

The `onOpen`, `onClose`, `onError`, and `onMessage` callback-registration methods are now `Stream` getters (the internal `stateChangeCallbacks` registry is gone):

```dart
client.onOpen.listen((_) => print('socket opened'));
client.onClose.listen((event) => print('socket closed: $event'));
client.onError.listen((error) => print('socket error: $error'));
client.onMessage.listen((message) => print('message: $message'));
```

### `RealtimeChannel`

`onPostgresChanges` and `onBroadcast` no longer take a `callback` and instead return typed streams, and `onPresenceSync`, `onPresenceJoin`, `onPresenceLeave`, and `onSystemEvents` are now stream getters. `onSystemEvents` also emits a typed `RealtimeSystemPayload` instead of a raw payload:

```dart
final channel = supabase.channel('room');
channel
    .onPostgresChanges(
      event: PostgresChangeEvent.insert,
      schema: 'public',
      table: 'messages',
    )
    .listen((payload) => print(payload.newRecord));
channel.onBroadcast(event: 'cursor-pos').listen(print);
channel.onPresenceSync.listen((_) => print(channel.presenceState()));
channel.subscribe();
```

`subscribe()` no longer takes a status callback. Status changes (with the error that caused a `channelError`) are exposed on the new `onStatusChange` stream via the new `RealtimeSubscribeStatusChange` class:

```dart
channel.onStatusChange.listen((change) {
  print('${change.status} ${change.error ?? ''}');
});
channel.subscribe();
```

All channel streams complete when the channel closes.

## Why

Streams are the idiomatic Dart shape for event listeners: consumers get `listen`/`map`/`where`/`firstWhere`/`timeout`, multiple subscribers, and, most importantly, easy listener removal via `StreamSubscription.cancel()`, which the callback API had no public equivalent for. This is a v3 breaking change requested in #1520.

The `examples/realtime_room` app shows the payoff: its hand-rolled `StreamController` wrappers around the callbacks are replaced with direct `map`/`where` transforms of the channel streams.

## Notes

- For `postgres_changes`, the stream must still be created before `subscribe()` (the requested changes are part of the join payload), but it can be listened to at any point.
- `SupabaseStreamBuilder` (`.stream()`) is migrated internally; its public API is unchanged.
- Registers `RealtimeChannel.onStatusChange` and `RealtimeSubscribeStatusChange` in `sdk-compliance.yaml`; the converted members keep their symbol names. Local symbol and drift checks pass.

## Tests

- All existing unit tests are migrated to the stream API (`realtime_client`: 191 passing, `supabase`: 134 passing, `supabase_flutter`: 76 passing).
- The realtime integration suite was run locally against a real Realtime server and passes for both protocol v1 and v2.

Resolves #1520

## Summary by CodeRabbit

- **New Features**
  - Realtime client and channel events are now available through typed, reusable streams.
  - Added stream-based connection, subscription-status, Postgres, broadcast, presence, system-event, and error notifications.
  - Subscription status changes include status details and optional errors.
  - Realtime room examples expose message, typing, deletion, and presence events as streams.

- **Documentation**
  - Updated examples and migration guidance for stream listeners and separate channel subscriptions.

- **Bug Fixes**
  - Improved subscription error reporting, event delivery, cleanup, and reconnect behavior.


Tr00d (approved):

Vinzent03:
Some general thoughts about the design that existed already previously as well. For open/closed/error on a channel we have one stream with different events via the `onStatusChange` method, but for the connection level on the socket we have dedicated `onClose`, `onError`, `onOpen`, `onMessage` methods. This is a different design for very similar aspects. Since the new stream way is a bit more boilerplate with `listen` and properly closing the subscription I think it could make more sense for just two streams. A message stream and a status change stream. Errors could be sent with the status change stream via `addError`. I think they are primarily for debugging purpose either way since errors and closed events are sent through to the channel.

For channel errors `onStatusChange` emits a `RealtimeSubscribeStatusChange` with a nullable error. I would consider handling them as stream errors as well. I think this would be even more idiomatic to make use of that feature of streams. Though this means every listener needs to listen for errors which would otherwise be uncaught.

I primarily want these two aspects to be considered. One could also keep it the current way to be more aligned with the js sdk as well.

spydon:
@Vinzent03 Definitely agree on the connection streams. The socket now has just `onStatusChange` (open and closed, with connection errors delivered via `addError`) and `onMessage`, so it has the same shape as a channel.

I kept the channel errors as they are, on `RealtimeSubscribeStatusChange`. Stream errors are more idiomatic in isolation, but channel errors are routine here: a network blip, a token that expired into a row level security denial, or a `postgres_changes` replication failure all land on that stream. With `addError`, every listener that does not pass `onError` leaks them to the zone, which is the failure from #1363, where an escaping async error was reported as fatal by Sentry and Crashlytics. Making every listener write an `onError` handler to avoid crash reports for something the client recovers from on its own seemed like the worse trade, and keeping the error on the payload also stays closer to supabase-js. What do you think?

Vinzent03:
Sounds reasonable, LGTM!

Vinzent03 (approved):

PR #1717, opened 2026-08-14
chore: use caret constraints for internal package dependencies

spydon:
## What

Every internal package dependency in this repo now uses a caret constraint instead of an exact pin.

| Package | Dependencies changed |
| --- | --- |
| `supabase` | `functions_client`, `postgrest`, `realtime_client`, `storage_client`, `supabase_auth`, `supabase_common`, `yet_another_json_isolate` |
| `supabase_flutter` | `supabase`, `supabase_common` |
| `functions_client` | `supabase_common`, `yet_another_json_isolate` |
| `postgrest` | `yet_another_json_isolate`, `supabase_common` |
| `supabase_auth` | `supabase_common` |
| `storage_client` | `supabase_common` |
| `realtime_client` | `supabase_common` |

## Why

Exact pins meant a patch release of any leaf package required republishing every package above it in the tree just to let users pick the fix up. With carets, a patch or minor release of an internal package flows through to consumers without a coordinated bump across the workspace.

## Notes

- `supabase_common: ^0.1.2` resolves to `>=0.1.2 <0.2.0` under pub's pre-1.0 caret rules, so a `0.2.0` release still requires bumping dependents. That is the standard convention for a pre-1.0 package.
- `supabase_auth: ^3.0.0-dev.1` resolves to `>=3.0.0-dev.1 <4.0.0`, which admits later dev prereleases and the eventual stable `3.0.0`.
- `supabase_lints` dev dependencies and the example apps already used carets, so they are untouched.

## Testing

`dart pub get` resolves the workspace cleanly.

### DIFF
--- a/packages/functions_client/pubspec.yaml
+++ b/packages/functions_client/pubspec.yaml
@@ -21,8 +21,8 @@ dependencies:
   http: ^1.6.0
   logging: ^1.3.0
   meta: ^1.16.0
-  supabase_common: 0.1.2
-  yet_another_json_isolate: 2.1.1
+  supabase_common: ^0.1.2
+  yet_another_json_isolate: ^2.1.1
 
 dev_dependencies:
   supabase_lints: ^0.1.1
--- a/packages/postgrest/pubspec.yaml
+++ b/packages/postgrest/pubspec.yaml
@@ -19,10 +19,10 @@ resolution: workspace
 
 dependencies:
   http: ^1.6.0
-  yet_another_json_isolate: 2.1.1
+  yet_another_json_isolate: ^2.1.1
   meta: ^1.16.0
   logging: ^1.3.0
-  supabase_common: 0.1.2
+  supabase_common: ^0.1.2
 
 dev_dependencies:
   collection: ^1.19.0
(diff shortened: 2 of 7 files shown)


grdsdev (changes requested):
We discourage this in the SDKs for the reason that when someone opens a support ticket with the supabase-flutter version, we can't know which sub package version they have if ^ is used.

spydon:
> We discourage this in the SDKs for the reason that when someone opens a support ticket with the supabase-flutter version, we can't know which sub package version they have if ^ is used.

That is seen as extremely bad practice in the Dart ecosystem (I'm pretty sure that's the case in other ecosystems too), because if you use community libraries that use our packages together then you won't be able to resolve the the dependencies without everything being pinned to the same version or using `pubspec_overrides`.

If there is an issue opened where this is unclear (usually quite uncommon) we usually just ask for the lockfile, and we'll get all information that we need.

EDIT: Here is the best practices link for Dart: https://dart.dev/tools/pub/dependencies#use-caret-syntax

spydon:
@grdsdev Furthermore, that leads to that we always have to release all packages, even if a fix only is needed in one package. Like in this case where we only would have had to release gotrue.

PR #1730, opened 2026-08-17
feat(storage)!: return the file id from upload and update methods

### LINKED ISSUE #1274: Get file id on storage upload
> Return file id on upload for storage. js implementation [here](https://github.com/supabase/storage/pull/332).
> 
> 
> 
> Basically this: https://github.com/supabase/supabase-flutter/pull/789#issuecomment-1879400440 

 _Originally posted by @dshukertjr in [#790](https://github.com/supabase/supabase-flutter/issues/790#issuecomment-1880016373)_

spydon:

Feature (breaking change, part of v3, #1278).

Closes #1274

`upload`, `uploadBinary`, `update`, `updateBinary`, `uploadToSignedUrl` and `uploadBinaryToSignedUrl` return a plain `String` key, so the id of the created object is discarded even though the storage server has been returning it since [supabase/storage#332](https://github.com/supabase/storage/pull/332).

## What is the new behavior?

All six methods return an `UploadResponse` with:

- `id`: the identifier of the stored object (`null` for signed URL uploads, where the server does not report it)
- `path`: the path of the object within the bucket
- `fullPath`: the path prefixed with the bucket id

This matches the `{ id, path, fullPath }` shape supabase-js returns from its upload methods.

## Additional context

- The upload and update methods now normalize the object path the same way the signed URL uploads already did (leading, trailing and duplicate slashes are removed), matching supabase-js.
- The new `UploadResponse` symbols are registered in `sdk-compliance.yaml`; the symbol, drift and schema checks pass locally.
- Unit tests and the storage integration suite pass against the local stack, including a new assertion that the real server returns the id.

### DIFF
--- a/examples/storage_transforms/lib/storage_repository.dart
+++ b/examples/storage_transforms/lib/storage_repository.dart
@@ -25,12 +25,13 @@ class StorageRepository {
   Future<String> uploadImage({
     required String name,
     required Uint8List bytes,
-  }) {
-    return _files.uploadBinary(
+  }) async {
+    final response = await _files.uploadBinary(
       name,
       bytes,
       fileOptions: const FileOptions(contentType: 'image/png', upsert: true),
     );
+    return response.path;
   }
 
   /// Lists the images in the bucket, newest first.
--- a/packages/supabase_storage/lib/src/fetch.dart
+++ b/packages/supabase_storage/lib/src/fetch.dart
@@ -69,7 +69,7 @@ class Fetch {
 
     _log.finest('Request: ${method.value} $url ${request.headers.redacted}');
     final streamedResponse = await request.sendWith(httpClient);
-    return _handleResponse<T>(streamedResponse, options);
+    return _handleResponse(streamedResponse, options);
   }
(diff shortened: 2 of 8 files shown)


packages/storage_client/lib/src/storage_file_api.dart
```diff
@@ -54,6 +54,15 @@ class StorageFileApi {
 
   FetchOptions get _fetchOptions => FetchOptions(headers);
 
+  UploadResponse _uploadResponse(String cleanPath, dynamic response) {
+    final data = response as Map;
```
Vinzent03: This line smells a bit. I know this existed before as well, but a better typing of the internal api by removing many `dynamic` would be beneficial.

Vinzent03 (approved):

packages/storage_client/lib/src/storage_file_api.dart
spydon: Very true, fixed!

Tr00d (approved):

PR #1731, opened 2026-08-18
refactor(postgrest)!: expose retry configuration as PostgrestRetryOptions

### LINKED ISSUE #1576: v3: expose PostgrestRetryOptions as a single public retry config
## Summary

In #1560 the PostgREST auto-retry settings became configurable (`retryEnabled`, `retryCount`, `retryableStatusCodes`). That PR is intentionally non-breaking, so those settings are exposed as individual named parameters on `PostgrestClient`, `PostgrestClientOptions`, and the builder constructors, and are bundled into a single immutable config only internally.

For v3 we should promote that internal bundle to a public options object and drop the individual parameters, so retry configuration is expressed as one value everywhere.

## Proposed change (breaking)

- Add a public `PostgrestRetryOptions` (or similar) with `enabled`, `count`, `statusCodes`, and the test-only `delay`.
- Replace the individual `retryEnabled` / `retryCount` / `retryableStatusCodes` parameters on `PostgrestClient` with a single `retryOptions` parameter.
- Do the same on `PostgrestClientOptions` in the `supabase` package (and surface it on `SupabaseClient`).
- Keep the per-request override `PostgrestBuilder.retry({bool enabled, int? count})` as the ergonomic shortcut; it can delegate to `retryOptions.copyWith(...)`.

## Why v3

Replacing the released `retryEnabled` parameter (shipped in postgrest v2.8.0) and the new `retryCount` / `retryableStatusCodes` parameters with a single object removes public constructor parameters, which is a breaking change for consumers.

## Notes

The internal `_RetryConfig` added in #1560 already models this shape, so the v3 work is mostly promoting it to public API and swapping the constructor surface. This also removes the last places where retry fields are threaded individually (the `PostgrestClient` -> builder boundary and the sub-builder public constructors).

spydon:
## What

Promotes the internal `_RetryConfig` added in #1560 to a public `PostgrestRetryOptions` and replaces the individual retry parameters with a single `retryOptions` value, so the retry configuration is one object everywhere.

```dart
final supabase = SupabaseClient(
  supabaseUrl,
  supabaseKey,
  postgrestOptions: const PostgrestClientOptions(
    retryOptions: PostgrestRetryOptions(count: 5, statusCodes: {503}),
  ),
);

// Per-request override, unchanged:
await supabase.from('todos').select().retry(enabled: false);
```

`PostgrestRetryOptions` has a const constructor and carries `enabled`, `count`, `statusCodes`, the backoff knobs `initialDelay`, `maxDelay` and `randomizationFactor`, a `delay(attempt)` computed from them with the shared `exponentialBackoff` helper, and a `copyWith`. Tuning the backoff was not possible before, and the field names match the `RetryOptions` of `supabase_common` so a later unification is a rename. Sharing one retry configuration across every client is tracked separately in #1735. `PostgrestBuilder.retry()` delegates to `retryOptions.copyWith(...)`, so the ergonomic shortcut keeps working and now composes on top of the client configuration.

### Breaking changes

| Before | After |
| --- | --- |
| `PostgrestClient(retryEnabled:, retryCount:, retryableStatusCodes:)` | `PostgrestClient(retryOptions:)` |
| `PostgrestClientOptions(retryEnabled:, retryCount:, retryableStatusCodes:)` | `PostgrestClientOptions(retryOptions:)` |
| `PostgrestQueryBuilder(retryEnabled:, …)` / `PostgrestRpcBuilder(retryEnabled:, …)` | `…(retryOptions:)` |
| `PostgrestClient.defaultRetryableStatusCodes` | `PostgrestRetryOptions.defaultStatusCodes` |

A negative retry count is now rejected by an assertion on `PostgrestRetryOptions` instead of an `ArgumentError` from the `PostgrestClient` constructor, and the retry loop treats a count below one as "send once" so a release build cannot reach the previously unreachable `StateError`.

### Fixed along the way

`SupabaseClient.from()` and `SupabaseClient.schema(…).from()` built their `SupabaseQueryBuilder` without forwarding the PostgREST options, so the retry settings and `requestTimeout` of `PostgrestClientOptions` only ever applied to `supabase.rest`. A client configured with `retryEnabled: false` still retried `supabase.from('todos').select()`. Both paths now pass the client's `retryOptions` and `requestTimeout` through. Users who configured those options and relied on the old behavior will see them take effect, which is noted in the migration guide.

## Tests

- `packages/postgrest/test/retry_test.dart` is ported to the new API, plus new cases for `copyWith` preserving unset fields and the client keeping the options it was given.
- New `packages/supabase/test/postgrest_options_test.dart` covers the threading fix for `from()`, `schema().from()` and `rpc()`, including disabled retries.
- `MIGRATION.md` gets a v3 section, and `sdk-compliance.yaml` registers the new symbols for `database.configuration.auto_retry` (symbol, drift and schema checks run clean locally).

Closes #1576

## Summary by CodeRabbit

- **New Features**
  - Added centralized retry settings for automatic request retries.
  - Supports retry limits, status codes, exponential backoff, randomization, defaults, and copying.
  - Retry settings apply consistently across queries, schemas, and RPC requests.

- **Improvements**
  - Requests with no retry attempts execute once without entering a retry loop.
  - Retry configuration is publicly available through the PostgREST library.
  - Made `AuthInvalidJwtException` constructor usable in constant expressions.

- **Documentation**
  - Updated migration guidance for the consolidated retry configuration.


Vinzent03:
The values itself of the `PostgrestRetryOptions` are not very specific to postgrest. Can it be shared with the other sub libraries as well? Such that we can unify it more? We already have the retry helper within `supabase_common` which I see is currently not used for the postgrest package. Using it would allow specifying values like the `maxDelay`. Having to almost distinct retry option classes is a bit strange. Is there a specific reason the common retry helper is not used? I suppose there is no specific reason, but is just the way it went historically.

spydon:
@Vinzent03 I agree, we should try to unify these (as long as it doesn't become generic to a point where it gets clunkier to use). I opened https://github.com/supabase/supabase-flutter/issues/1735 that I plan to do as a follow-up once this one is merged.

packages/postgrest/lib/src/postgrest_retry_options.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  /// The fraction between 0 and 1 the delay is randomized by.
+  ///
+  /// Defaults to `0`, which waits exactly as long as the backoff says.
+  final double randomizationFactor;
+
+  const PostgrestRetryOptions({
+    this.enabled = true,
+    this.count = 3,
+    this.statusCodes = defaultStatusCodes,
```
grdsdev, line 51: 503 and 520 are the only status codes that should be retried, let's not allow clients to pass a different status code set as that can lead to broader issues.

grdsdev (changes requested):

packages/postgrest/lib/src/postgrest_retry_options.dart
spydon: Ah, that's not introduced in this PR though (thus shouldn't be removed here), but I'll open a follow-up issue and remove that option tomorrow once this is merged.

spydon: https://github.com/supabase/supabase-flutter/issues/1736

grdsdev (approved):

PR #1737, opened 2026-08-19
refactor(postgrest)!: fix the retried status codes to 503 and 520

### LINKED ISSUE #1736: 503 and 520 are the only status codes that should be retried
503 and 520 are the only status codes that should be retried, let's not allow clients to pass a different status code set as that can lead to broader issues.

_Originally posted by @grdsdev in https://github.com/supabase/supabase-flutter/pull/1731#discussion_r3805358847_

spydon:
## What

`PostgrestRetryOptions.statusCodes` is a static constant now instead of a configurable field, so `503 Service Unavailable` and `520 Unknown Error` are the only responses that ever trigger a retry.

Letting callers pass their own set invites retrying responses that cannot change on a second attempt, a `500` from a failing query for example, which only multiplies the load on the project without a chance of a different answer.

```dart
// Before
retryOptions: PostgrestRetryOptions(count: 5, statusCodes: {500, 503, 520}),

// After
retryOptions: PostgrestRetryOptions(count: 5),
```

Everything else about the retry configuration is unchanged: `enabled`, `count`, the backoff knobs and the per-request `.retry()` override all stay.

Since the set is a `const` now, the two places that snapshotted it with `Set.unmodifiable` so a caller could not mutate a live client's behavior are gone as well.

### Breaking changes

| Before | After |
| --- | --- |
| `PostgrestRetryOptions(statusCodes: …)` | removed |
| `PostgrestRetryOptions.copyWith(statusCodes: …)` | removed |
| `PostgrestRetryOptions.statusCodes` (instance field) | `PostgrestRetryOptions.statusCodes` (static constant) |
| `PostgrestRetryOptions.defaultStatusCodes` | `PostgrestRetryOptions.statusCodes` |

The v2 `PostgrestClient(retryableStatusCodes:)` and `PostgrestClientOptions(retryableStatusCodes:)` were already replaced by `retryOptions` in #1731, and now have no replacement at all. `MIGRATION.md` gets a section for that and the existing v3 retry section drops its `statusCodes` references.

## Tests

- `packages/postgrest/test/retry_test.dart` loses the "configurable retryable status codes" group, gains an explicit "GET does not retry on 500" case and an assertion that the fixed set is `{503, 520}`.
- `packages/supabase/test/postgrest_options_test.dart` proved that the configured options were threaded through `from()`, `schema().from()` and `rpc()` by retrying a custom status code. It now uses a retry count above the default of three instead, so the request only recovers on the fifth response.
- That file also fails on `main` since #1489 landed, because `SupabaseClient` asserts on a PKCE flow without `asyncStorage`. It passes a `MemoryAuthAsyncStorage` now, which is unrelated to the retry change but needed for the file to run at all.
- `sdk-compliance.yaml` drops `PostgrestRetryOptions.defaultStatusCodes`; symbol, drift and schema checks run clean locally.

Closes #1736

## Summary by CodeRabbit

* **Breaking Changes**
  * Retry configuration now uses a single `PostgrestRetryOptions` value across clients and request builders.
  * Automatic retries are limited to HTTP 503 and 520 responses.
  * Custom retry status codes are no longer supported; handle other statuses by catching API exceptions.
  * Per-request retry count overrides remain available.
* **Tests**
  * Updated coverage confirms retry counts and status-code behavior.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1237,22 +1237,52 @@ The individual retry parameters are replaced by one `PostgrestRetryOptions` valu
 
 | Before | After |
 | --- | --- |
-| `PostgrestClient(retryEnabled: …, retryCount: …, retryableStatusCodes: …)` | `PostgrestClient(retryOptions: …)` |
-| `PostgrestClientOptions(retryEnabled: …, retryCount: …, retryableStatusCodes: …)` | `PostgrestClientOptions(retryOptions: …)` |
-| `PostgrestBuilder`, `PostgrestQueryBuilder` and `PostgrestRpcBuilder` constructors, same three parameters | `retryOptions: …` |
-| `PostgrestClient.defaultRetryableStatusCodes` | `PostgrestRetryOptions.defaultStatusCodes` |
+| `PostgrestClient(retryEnabled: …, retryCount: …)` | `PostgrestClient(retryOptions: …)` |
+| `PostgrestClientOptions(retryEnabled: …, retryCount: …)` | `PostgrestClientOptions(retryOptions: …)` |
+| `PostgrestBuilder`, `PostgrestQueryBuilder` and `PostgrestRpcBuilder` constructors, same parameters | `retryOptions: …` |
 
 ```dart
 // Before
 postgrestOptions: const PostgrestClientOptions(
   retryCount: 5,
-  retryableStatusCodes: {503},
 ),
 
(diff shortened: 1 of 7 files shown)


spydon:
@grdsdev fix regarding what we spoke about yesterday :)

Tr00d (approved):

Issue #376, opened 2023-02-27
Outdated auth headers on sub library clients


Vinzent03:
**Describe the bug**
When you save your `StorageClient` or `FunctionsClient` in a local variable. E.g. to pass it to a class only responsible for storage, its auth headers get outdated.
```dart
final storage = supabase.storage;
supabase.auth.signOut();
storage.from();//calls still use the old headers;
```
`storage` is just a getter.
```dart
SupabaseStorageClient get storage => SupabaseStorageClient(
        storageUrl,
        _getAuthHeaders(),
        httpClient: _httpClient,
        retryAttempts: _storageRetryAttempts,
      );
```

**Expected behavior**
Even though you save a client in a separate variable, I expect its auth headers still to be updated.

**My proposal**
~~My idea to solve this is that we pass a function to the `SupabaseStorageClient` constructor, which returns the latest auth headers. So the storage client call the function, when it needs the headers.
What do you think?~~

New approach is identical to how we already handle this for realtime, by adding a `setAuth` method and calling it on auth state change.

PR #1739, opened 2026-08-19
refactor!: remove the dead setAccessToken mutators

spydon:
## Summary

Removes `PostgrestClient.setAccessToken()`, `SupabaseStorageClient.setAccessToken()` and `FunctionsClient.setAccessToken()`. `RealtimeClient.setAccessToken()` stays.

Nothing in the repository called the three that are gone, but they were not inert either. Despite the name they pinned a token rather than kept one in sync: `AuthHttpClient` applies the current session token with `putIfAbsent`, so an `Authorization` header set through one of these setters took precedence over it, and nothing ever cleared it. The pinned token kept shadowing the session across refreshes and sign-outs for the rest of the client's life. A method named `setAccessToken` on a client that is otherwise authenticated per request is a poor way to spell that.

The capability survives without them. The header maps are still mutable, so a token can be pinned with `supabase.rest.headers['Authorization'] = ...` or `storage.setHeader(...)`, passed to a constructor, or scoped to a single call with `invoke(headers: ...)` / `.setHeader(...)`. What goes away is four inconsistent signatures for the same idea, one of which reads like auth synchronization.

Realtime is the real case for a setter: it holds a live socket and has to push a new token over it rather than attach one per request, so `SupabaseClient` genuinely calls its setter on auth state changes.

## Why

Mirrors https://github.com/supabase/supabase-swift/pull/1233, which removed the equivalent `FunctionsClient.setAuth(token:)` from supabase-swift in favour of a per-request access token closure. Flutter already resolves the token per request through `AuthHttpClient`, so only the cleanup half applies here.

The header-clobbering bug that PR also fixed does not exist in this SDK: `putIfAbsent` means an `Authorization` passed to a single `invoke` or query survives the live token. `client_test.dart` now covers that end to end, since it is the replacement this migration points callers at.

## Notes

The three header maps stay mutable. `SupabaseClient.headers` still rewrites them in place on assignment, and `SupabaseStorageClient.setHeader()` still mutates its own. Making them unmodifiable is a separate change that has to deal with that setter first.

`sdk-compliance.yaml` is updated for both affected capabilities:

- `functions.invocation.set_auth_token` now points at the constructor and `invoke`, which are the two ways to supply the header, with a note on the deviation from the reference implementation.
- `client.authentication_integration.cross_client_token_sync` now lists `RealtimeClient.setAccessToken` and `SupabaseClient.accessToken`, with a note explaining why the other three clients need no setter.

`MIGRATION.md` also drops a step that had gone stale independently of this change: the v2 deprecation table told readers to replace `PostgrestClient.auth()` with `PostgrestClient.setAuth()`, which was renamed earlier in v3 and is now removed outright.

## Test plan

- `dart test` in `postgrest` (202), `supabase_functions` (50) and `supabase` (143), all passing
- New test in `client_test.dart`: a per-request `Authorization` beats the live session token for both functions and rest. Verified it fails if `AuthHttpClient` is changed to overwrite instead of `putIfAbsent`
- `dart analyze` clean and `dart format -l 80` clean on the four affected packages
- Capability matrix validated against supabase/sdk: compliance file valid, symbol check reports all new public symbols covered
- `supabase_storage` has 21 failures locally from a dirty local stack, identical on clean `origin/main`; CI runs against a fresh stack

Closes SDK-1521

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed `setAccessToken()` from PostgREST, Storage, and Functions clients.
  * Renamed Realtime’s `setAuth()` method to `setAccessToken()`.
  * PostgREST authentication now requires an `Authorization` header.

* **New Features**
  * Added chainable `setHeader()` support for updating Storage client headers.

* **Documentation**
  * Added migration guidance for constructor and per-request authorization headers.
  * Clarified authorization requirements for standalone Functions clients and session-based invocations.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1339,16 +1339,19 @@ try {
 `PostgrestClient.setAccessToken()`, `SupabaseStorageClient.setAccessToken()` and
 `FunctionsClient.setAccessToken()` are removed. `RealtimeClient.setAccessToken()` stays.
 
-`SupabaseClient` never called any of the three. It wraps the HTTP client the rest, storage and
-functions clients use, and resolves the current session token on every request, so a token pushed in
-by hand was overwritten by the live one anyway. Nothing changes for you if you go through
-`SupabaseClient`, and the token you set with `SupabaseClient.accessToken` or by signing in still
-reaches all three.
+Despite the name, these three pinned a token rather than kept one in sync. `SupabaseClient` gives
+the rest, storage and functions clients an HTTP client that attaches the current session token to
+every request, but only when the request does not already carry an `Authorization` header. A token
+set through `setAccessToken` did carry one, so it won, and nothing ever cleared it: it kept
+overriding the session token across refreshes and sign-outs for the rest of the client's life.
 
-Realtime keeps its setter because it holds a live socket and has to push a new token over it rather
-than attach one per request.
+If you never called them, nothing changes. If you did, the replacement depends on what you were
+after.
 
-If you construct one of these clients yourself, pass the header to the constructor instead:
+To authenticate as the signed-in user, do nothing. `SupabaseClient` already resolves that token on
+every request.
+
+To pin a token on a client you construct yourself, pass it as a constructor header:
 
 ```dart
 // Before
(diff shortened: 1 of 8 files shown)


Tr00d (approved):

Vinzent03:
I think the origin reason for these methods was to support their individual usages better. So if one does not use the `supabase` package no auth http client is available. I think but I'm not sure that the headers were modifiable back then as well, and having a dedicated method is easier than having to deal with the exact header semantics.
So I would say there are valid reasons to keep it for individual package use. Especially the postgrest package could be used without supabase and providing an easier authentication mechanism than manipulating headers would be great.

spydon:
@Vinzent03 I was a bit curious of the history of this so I let the clanker dig through it: 

> I dug through the history, and they were actually added *for* the composed case rather than for standalone use.
> 
> #376 (Feb 2023) was `supabase.storage` rebuilding the client on every access, so a stored sub-client went stale after sign-out. Your first proposal in that issue was a closure returning the current headers; it got struck through in favour of "identical to how we already handle this for realtime, by adding a `setAuth` method". supabase/storage-dart#52 says "In preparation for #376", and supabase/supabase-dart#182 wired it up. supabase/supabase-dart#189 and supabase/supabase-dart#192 followed within a month (header bleed between queries, then a query after sign-in using the previous session). #540 (Jul 2023) replaced the whole thing with `AuthHttpClient` and deleted the six `setAuth` calls, which is the closure idea from #376. The setters themselves just never got removed.

Independently of that your point still stands though. Standalone users do need token rotation, and constructor headers only cover the static case.

We want to avoid today's shape though, where the method means two different things. Since #540, `AuthHttpClient` applies the session token with `putIfAbsent`, so calling `setAccessToken` on a client obtained from `SupabaseClient` pins a token that shadows the session permanently, across refreshes and sign-outs.

Good news though is that we're going to do it in the same way as supabase/supabase-swift#1233 just landed: an `accessToken` callback resolved per request. It handles rotation, and it composes correctly instead of fighting the parent client. (Coming up as a stacked PR very soon)

Vinzent03:
Okay interesting, thanks. Didn't remember I added them to begin with, but with a new callback approach this seems to be solved even better!

Vinzent03 (approved):

spydon:
> Okay interesting, thanks. Didn't remember I added them to begin with, but with a new callback approach this seems to be solved even better!

Hehe yeah, it was a very long time ago, I would never have remembered either. 😄

Issue #1749, opened 2026-08-21
Shared JSON isolate is an unrecoverable single point of failure, and Supabase.initialize won't forward isolate:


shellyneira:
## Summary

`SupabaseClient` routes work through one internal `YAJsonIsolate` — every
`functions.invoke` body and reply, and every postgrest response over 10 KB. That
worker has no request ids, no reply deadline, no liveness check and no respawn,
and `onExit`/`onError` are delivered on the same port as replies. If it ever
stops answering, `_sendPort.send` goes into the void and **every affected call
never completes** — no exception, no log, no recovery for the lifetime of the
process. The app just spins.

There is also no way to supply or supervise your own instance:
`SupabaseClient` accepts an `isolate:` parameter, but `Supabase.initialize` in
`supabase_flutter` does not forward it, so a `supabase_flutter` app cannot get
a handle on the isolate its client depends on.

Call sites:

- `functions_client/lib/src/functions_client.dart` — `await _isolate.encode(body)`
  runs before the request is sent, and `_isolate.decode` on the JSON reply.
- `postgrest/lib/src/postgrest_builder.dart` — `if ((response.contentLength ?? 0) > 10000 && _isolate != null) body = await _isolate.decode(response.body)`.
- `yet_another_json_isolate` 2.1.1 `lib/src/_isolates_io.dart` — one `ReceivePort`,
  one `StreamQueue`, replies matched by arrival order.

## Impact

Because `functions.invoke` awaits the isolate *before* the HTTP request, a
non-responsive worker means the request never reaches the network at all. In our
case that was the passwordless sign-in flow: the app sat on
`functions.invoke('auth-start')` until its own timeout, and the backend showed
zero requests — no gateway log, no function invocation, nothing. Meanwhile GoTrue
calls and every postgrest response under 10 KB kept working normally, which makes
this very hard to attribute: it looks like a network or backend problem, and it
is neither.

## Reproduction of the failure mode

A worker that terminates without the cooperative `dispose()` path (crash, OOM,
`Isolate.kill`) leaves the port open, so the single stray `onExit` message
poisons whichever caller is first in the queue and every later call waits on a
reply that will never arrive. Running the 2.1.1 protocol with a handle on the
isolate so it can be killed the way the OS would kill it:

```
alive -> {a: 1}
isolate killed
call 0 -> _TypeError: type 'Null' is not a subtype of type 'List<dynamic>'
call 1 -> *** HUNG, silent, forever ***
[...]

PR #1750, opened 2026-08-21
feat(supabase_flutter): forward the isolate parameter through Supabase.initialize

shellyneira:
Closes the remaining item from #1749.

`SupabaseClient` already takes an `isolate:` parameter and already tracks
ownership — `_hasCustomIsolate` keeps `dispose()` from disposing an instance it
did not create. `Supabase.initialize` just never forwarded it, so a
`supabase_flutter` app cannot reach the parameter at all.

`YAJsonIsolate` was also not exported by `supabase` or `supabase_flutter`, which
made the existing `SupabaseClient` parameter unusable without adding a direct
dependency on `yet_another_json_isolate`. This exports the type (`show
YAJsonIsolate`) so the public API is callable, and drops three now-redundant
imports the analyzer flagged as a result.

### Changes

- `supabase`: export `YAJsonIsolate`.
- `supabase_flutter`: `isolate:` on `Supabase.initialize`, threaded to
  `SupabaseClient` through `_init`, with dartdoc noting that a supplied instance
  is owned by the caller.
- Two tests: initialization with a caller supplied isolate, and that such an
  isolate is still usable after `Supabase.instance.dispose()`.

### Why

Sharing one instance with code outside Supabase, and — for anyone still on the
released `yet_another_json_isolate` 2.1.1, where a single persistent worker
backs every `functions.invoke` and every large postgrest decode — being able to
hold and supervise that instance. #1746 makes the second reason far less
pressing once it ships; the first stands on its own.

### Verification

`flutter analyze` clean on both packages. `packages/supabase` and
`packages/supabase_flutter` test suites pass. `stream_integration_test.dart`
fails identically before and after this change (it needs a local stack).

## Summary by CodeRabbit

* **New Features**
  * Added support for supplying a custom JSON-processing isolate during Supabase initialization.
  * Exposed the JSON isolate type for use by applications.
  * Caller-provided isolates remain under caller ownership and are not disposed automatically.

* **Tests**
  * Added coverage confirming custom isolate initialization succeeds and lifecycle ownership is preserved.

### DIFF
--- a/packages/supabase/lib/src/supabase_client.dart
+++ b/packages/supabase/lib/src/supabase_client.dart
@@ -5,7 +5,6 @@ import 'package:supabase/src/logger.dart';
 import 'package:supabase/src/supabase_constants.dart';
 import 'package:supabase/src/version.dart';
 import 'package:supabase/supabase.dart';
-import 'package:yet_another_json_isolate/yet_another_json_isolate.dart';
 
 import 'api_key.dart';
 import 'auth_http_client.dart';
--- a/packages/supabase/lib/src/supabase_query_schema.dart
+++ b/packages/supabase/lib/src/supabase_query_schema.dart
@@ -1,6 +1,5 @@
 import 'package:http/http.dart';
 import 'package:supabase/supabase.dart';
-import 'package:yet_another_json_isolate/yet_another_json_isolate.dart';
 
 import 'counter.dart';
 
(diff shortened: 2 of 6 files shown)


spydon (changes requested):
I don't see the point of this now when the isolate isn't long-lived...?
The export will be done soon though (the other isolate PR was just merged yesterday.

spydon:
Thanks for the contribution, but as mentioned in the V3 issue (not very visible, sorry for that) we're not taking in contributions for V3 code at this time.
Superseded by #1751

PR #1764, opened 2026-08-25
fix!: return unmodifiable header views with controlled replacement

spydon:

Breaking change, plus a small feature that keeps every header use case reachable.

`AuthClient.headers`, `FunctionsClient.headers`, `StorageBucketApi.headers` (so `SupabaseStorageClient.headers`) and `StorageFileApi.headers` all handed out the live internal header map, so any consumer could rewrite the headers of a client from anywhere, including the sub-clients a `SupabaseClient` manages:

```dart
supabase.functions.headers['Authorization'] = 'Bearer $jwt';
supabase.storage.headers.clear();
```

`SupabaseClient._replaceHeaders` depended on that leak, mutating each client with `..clear()..addAll(...)`. This is the same anti-pattern that was fixed for `RealtimeClient.headers` in #1758, and `PostgrestClient.headers` before that.

Closes SDK-1600.

## What is the new behavior?

- Every `headers` getter returns `Map.unmodifiable(_headers)`, so mutating one throws an `UnsupportedError`.
- Each client gained an `@internal void replaceHeaders(Map<String, String>)`, and `SupabaseClient.headers =` now calls that instead of mutating the maps in place. Propagation to all sub-clients is unchanged.
- `AuthClient` and `FunctionsClient` gained a public `setHeader(String, String)`, mirroring `SupabaseStorageClient.setHeader`. Without it, a standalone client built directly (not through `SupabaseClient`) would have had no way to add a header after construction, since in-place mutation was previously the only one. On auth the header map is shared with `admin`, so admin requests carry it too; on functions a header passed to `invoke()` still wins for that single call.
- `SupabaseStorageClient.setHeader()` goes through `replaceHeaders` as well, so it keeps working. `StorageBucketApi` copies the headers it is constructed with rather than aliasing the caller's map.
- `SupabaseStorageClient.vectors` is a getter that builds a client per access instead of a `late final` field, so it always carries the current headers. It used to alias the live map, which a cached snapshot would have silently broken.
- `StorageBucketApi.storageFetch` is `late final`.
- `MIGRATION.md` documents the change, and the outdated `supabase.functions.headers[...] = ...` advice in the `setAccessToken()` section is corrected to use `setHeader`.

Adding custom headers therefore stays possible everywhere: the constructor for a whole set, `setHeader()` for a single one on auth, functions and storage (and per bucket), `invoke(headers: …)` or a builder's `setHeader()` for a single call, and `SupabaseClient.headers =` to reach every sub-client at once.

## Additional context

`sdk-compliance.yaml` registers the two new `setHeader` symbols under `client.request_configuration.global_headers`. The symbol, drift and schema checks from supabase/sdk all pass locally.

## Summary by CodeRabbit

- **New Features**
  - Added chainable `setHeader()` methods for authentication and Edge Functions clients.
  - Header updates now propagate consistently across related clients and requests.
  - Storage vector clients receive the latest configured headers on access.

- **Bug Fixes**
  - Prevented accidental in-place modification of client and bucket header maps.
  - Preserved custom headers while maintaining required API key headers.

- **Documentation**
  - Updated migration guidance with the new header configuration approach and immutable header behavior.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1531,7 +1531,7 @@ builds its sub-clients, so there is no `accessToken` callback to pass:
 
 ```dart
 supabase.storage.setHeader('Authorization', 'Bearer $jwt');
-supabase.functions.headers['Authorization'] = 'Bearer $jwt';
+supabase.headers = {...supabase.headers, 'Authorization': 'Bearer $jwt'};
 ```
 
 The rest client is stateless and its header map unmodifiable (see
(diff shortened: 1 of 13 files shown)


grdsdev (approved):
Reviewed the unmodifiable-headers change carefully given the breaking-change marker.

- Map.unmodifiable is used correctly everywhere (fresh snapshot per getter call, not cached), so it can't go stale.
- Traced every internal call site across supabase, supabase_auth, supabase_functions, supabase_storage (and iceberg/vector sub-clients) via full-repo grep for '..clear()/addAll', 'headers[…]=' and similar: all former in-place mutations of client header maps now go through the new replaceHeaders()/setHeader() APIs. No missed mutation sites found.
- StorageBucketApi now copies constructor-provided headers instead of aliasing the caller's map; StorageFileApi/SupabaseVectorsClient already copied. auth_client.dart's two remaining internal 'headers' getter reads were switched to '_headers' to avoid needless unmodifiable wrapper churn, consistent with the rest of the admin/mfa/oauth/passkey sub-APIs which already used '_headers' directly.
- replaceHeaders() preserves the old clear+addAll (full replace) semantics, so no merge-vs-replace regression. setHeader() implementations preserve prior add/overwrite-single-key semantics.
- vectors becoming a per-access getter (vs a cached late final) is a real, intentional behavior change - it's called out explicitly in MIGRATION.md with guidance, and has new test coverage.
- Good test coverage: immutability (throwsUnsupportedError) for auth/functions/storage/bucket, setHeader propagation to auth admin, invoke() header precedence over setHeader, and vectors picking up updated headers.

No correctness issues or missed call sites found. Approving.

PR #1765, opened 2026-08-25
refactor(realtime)!: make RealtimeChannel internal mutable fields private with controlled accessors

spydon:

Breaking change (v3), follow-up to #1758 / SDK-818.

#1758 privatized the mutable internals of `RealtimeClient`, but `RealtimeChannel` still exposes its join bookkeeping as public mutable fields that are only guarded by `@internal` annotations:

- `joinedOnce` is a mutable flag, so external code can unlock a second `subscribe()`.
- `joinPush` and `presence` are `late` and reassignable, so the join push or the presence tracker can be swapped out underneath the channel.
- `parameters` is both reassignable and a live mutable map, so callers can corrupt the join payload directly instead of going through `updateJoinPayload`.

The unexported `Push` and `RealtimePresence` classes have the same shape (`payload`/`sent`, `state`/`pendingDiffs`/`joinRef`/`caller`).

## What is the new behavior?

Same treatment as #1758, private backing fields with controlled accessors:

- `joinedOnce`: private `_joinedOnce`, read-only `@internal` getter (read by `RealtimeClient` and tests).
- `joinPush`: private `late final _joinPush`, read-only `@internal @visibleForTesting` getter (only tests need it).
- `presence`: fully private `late final _presence`, nothing outside the class read it; presence state is exposed through `presenceState()` and the presence streams.
- `parameters`: private final `_parameters`, the `@internal` getter returns an `UnmodifiableMapView`; internal mutation keeps going through `updateJoinPayload`.

Optional cleanup while in the area (both classes are `@internal` and not exported):

- `Push.payload` is a getter over a private field that is only replaced through `updatePayload`; `Push.sent` was write-only dead state and is removed.
- `RealtimePresence.state` is a getter over a private field; `pendingDiffs`, `joinRef`, and `caller` are private.

MIGRATION.md documents the change since the fields were reachable by consumers. No public symbols changed for the compliance matrix (everything involved is `@internal`), verified with the local symbol check.

## Additional context

Resolves SDK-1601.

## Summary by CodeRabbit

* **Breaking Changes**
  * Channel configuration and internal presence state are now read-only after creation.
  * Push payloads and session/auth request data can no longer be modified directly.
  * Configure channel parameters and request headers through supported APIs.

* **Documentation**
  * Added v2-to-v3 migration guidance, including replacements for removed mutable properties and updated configuration practices.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -330,6 +330,25 @@ final channels = client.channels;
 `setAccessToken()` is the intentional path for token rotation: unlike the removed field write, it
 also propagates the new token to every joined channel.
 
+### The mutable internals of `RealtimeChannel` are private
+
+`RealtimeChannel` received the same treatment as `RealtimeClient`. Its join bookkeeping used to be
+public mutable fields that were only marked `@internal`, so external code could reassign the join
+push, replace the presence tracker, or mutate the join parameters underneath the channel:
+
+| Before | After |
+| --- | --- |
+| `channel.joinedOnce = value` | removed, the getter remains (internal) |
+| `channel.joinPush = push` | removed, the getter remains (internal and test-only) |
+| `channel.presence` | removed |
+| `channel.parameters` (mutable map) | `channel.parameters` (unmodifiable view, internal) |
+
+These members were never part of the supported API surface, but they were reachable. If you read
+presence state through `channel.presence`, use `channel.presenceState()` and the
+`onPresenceSync`, `onPresenceJoin`, and `onPresenceLeave` streams instead. The join payload is
+only updated internally; there is no supported way to mutate `parameters` after the channel is
+created, so pass the configuration through `RealtimeChannelConfig` when creating the channel.
+
 ### Broadcasts no longer fall back to the REST API
 
 `sendBroadcastMessage()` used to silently post to the REST broadcast endpoint whenever the channel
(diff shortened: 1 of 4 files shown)


grdsdev (approved):
Reviewed the diff and confirmed: (1) accessor semantics match the old fields exactly — Push.payload/updatePayload and RealtimePresence state/caller/pendingDiffs/joinRef are unchanged aside from visibility, Push.sent was genuinely dead (write-only, no readers) so its removal is safe; (2) grepped the whole repo for every removed/changed field (joinedOnce, joinPush, presence, parameters, sent) and all remaining call sites (realtime_client.dart:908, channel_test.dart, socket_test.dart mocks) are read-only through the new getters, so nothing relies on reassignment; (3) parameters' UnmodifiableMapView still exposes the same live underlying map for reads, and internal mutation continues through the private _parameters field via updateJoinPayload, so no state-machine/race behavior changed; (4) this was never public API (@internal-only), and MIGRATION.md documents the removed/replaced accessors with guidance (presenceState()/streams instead of .presence, RealtimeChannelConfig instead of mutating .parameters); (5) no new tests were added but none were needed — this is a pure visibility refactor and the full existing test suite (channel_test.dart, socket_test.dart) exercises the new getters unchanged. No correctness or breaking-change-scope issues found.

PR #1781, opened 2026-08-28
fix(realtime): parse Postgres array literals instead of splitting on commas

fahaddoc:
## What

`toArray` in `supabase_realtime` decoded a Postgres array literal by trying `json.decode` on the body and, when that threw, splitting the string on every comma. Both paths carried a `TODO` and a `WARNING: splitting on comma does not cover all edge cases`, and the edge cases turn into wrong data in realtime payloads:

| Literal Postgres sends | Before | After |
| --- | --- | --- |
| `{"a,b",c}` | `['"a', ' b"', 'c']` (three elements) | `['a,b', 'c']` |
| `{NULL,a}` on `text[]` | `['NULL', 'a']` | `[null, 'a']` |
| `{"",a}` | `['""', 'a']` | `['', 'a']` |
| `{{1,2},{3,4}}` on `int4[]` | `[null, null, null, null]` | `[[1, 2], [3, 4]]` |

A `text[]` column whose values contain commas is the common case here: subscribers got extra elements with stray quote characters in them, and nothing signalled that the value had been mangled.

This replaces both paths with a parser for the literal itself: comma separated elements, optional double quotes where `\` escapes the next character, whitespace around unquoted elements dropped, an unquoted `NULL` as the null element (a quoted `"NULL"` stays the four character string), and nested arrays keeping their shape. A literal that doesn't parse now returns the raw string instead of data that looks structured but isn't.

`{}`, `{1,2,3}` and quoted ranges like `{"[2021-01-01,2021-12-31)"}` behave exactly as before — the existing tests cover those and are unchanged.

## Note on parity with supabase-js

`realtime-js` still has the same `JSON.parse`-then-split logic (`src/lib/transformers.ts`), including the same TODO and warning comments, so after this change the Dart client parses these literals where the JS client does not. I went with correctness here since the current behaviour silently returns wrong values, but happy to align differently if you'd rather keep the two in lockstep, and I can open the matching issue on `realtime-js`.

## Tests

Added a `toArray with quoting` group in `packages/supabase_realtime/test/transformers_test.dart` covering quoted commas, `NULL` vs `"NULL"`, empty string elements, nested arrays, escaped quotes and backslashes, quoted braces, whitespace handling, and malformed literals.

`dart test -j 1 --exclude-tags integration` in `packages/supabase_realtime` passes (239 tests), and `dart analyze packages/supabase_realtime` is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL array parsing for quoted values, escaped characters, whitespace, nested arrays, empty strings, and `NULL` values.
  * Added support for explicit dimension prefixes and PostgreSQL `box` array delimiters.
  * Malformed array literals, including invalid unquoted elements, are now safely returned unchanged instead of being incorrectly parsed.

* **Tests**
  * Added coverage for multidimensional arrays, quoting, escaping, whitespace, embedded braces, dimension prefixes, `box` values, and malformed input.

### DIFF
--- a/packages/supabase_realtime/lib/src/transformers.dart
+++ b/packages/supabase_realtime/lib/src/transformers.dart
@@ -346,12 +346,21 @@ dynamic toArray(dynamic value, String type) {
     return value;
   }
 
+  // Postgres prefixes the literal with explicit dimensions, for example
+  // `[0:1]={1,2}`, when a dimension's lower bound is not 1.
+  final literal = value.replaceFirst(_arrayDimensions, '');
+
   // Confirm value is a Postgres array by checking curly brackets
-  if (value.length < 2 || !value.startsWith('{') || !value.endsWith('}')) {
+  if (literal.length < 2 ||
+      !literal.startsWith('{') ||
+      !literal.endsWith('}')) {
     return value;
   }
 
-  final elements = _ArrayLiteralParser(value).parse();
+  final elements = _ArrayLiteralParser(
+    literal,
+    _elementDelimiter(type),
+  ).parse();
   if (elements == null) {
     // Not a literal this parser understands, hand the raw value back rather
     // than splitting it into something that looks like data but isn't.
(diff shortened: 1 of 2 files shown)


spydon (approved):
Thanks for your contribution!

fahaddoc:
thanks for taking it the rest of the way, the dimension decorations and the box delimiter hadn't occurred to me.

that offer on the js side still stands if you want it, realtime-js still has the same json parse then split fallback with the same TODO on it. happy to open the issue there.

spydon:
> thanks for taking it the rest of the way, the dimension decorations and the box delimiter hadn't occurred to me.
> 
> that offer on the js side still stands if you want it, realtime-js still has the same json parse then split fallback with the same TODO on it. happy to open the issue there.

I opened an internal issue for it, let's see what Katerina says on Monday :)

PR #1793, opened 2026-09-07
feat(postgrest): add column expressions and the filter tree

spydon:
Stack 1/7 for SDK-1741, the Dart port of the supabase-swift SDK-1624 stack (supabase/supabase-swift#1288). Base: `main`.

Adds the vocabulary for a typed PostgREST surface where casts, JSON paths, aggregates and embedded relations are all expressible and compile-time checked. Nothing consumes it yet; `where` and `order` move onto it in 3/7.

**`PostgrestColumnExpression<Row, Value>`** is anything valid in a `select` list, with two independent refinements: `PostgrestFilterableExpression` (left of an operator) and `PostgrestOrderableExpression` (an `order` key), because PostgREST accepts a different expression set in each position. `Row` is the table's row type, so a filter built from one table's column is a `PostgrestFilter<ThatRow>` and cannot be passed to a query on another table; `Value` makes `.eq('seven')` on an `int` column a compile error. `PostgrestNullableExpression` is where `isNull()` lives, so a `NOT NULL` column does not get a filter that can never match.

**`PostgrestColumn`** and **`PostgrestNullableColumn`** are the stored column kinds. They stay static members of the table's namespace class (`Books.title`), where the generator already puts them: Dart has no `$0` shorthand for a closure over a columns value, and the row type parameter gives the relation check the Swift `Root` associated type gives.

**`PostgrestFilter<Row>`** is the tree, composed with `&`, `|` and `not()` (Dart cannot overload `&&`, `||` or `!`), rendered to query parameters. Rendering is position-dependent, following what the Swift stack verified against PostgREST 16.1:

- a top-level operand goes unescaped, the same operand inside `or=(…)` must be escaped;
- a negated leaf in a group is `id.not.eq.2`; `not.id.eq.2` is a PGRST100 parse error;
- double negation collapses in both positions.

Values render through one function: `DateTime` in ISO 8601, a `List` as a `{a,b}` array literal with element escaping, a `Map` as JSON, everything else through `toString`, which for the enums `supabase_typegen` generates is the wire name. Quoting is applied only where a value needs it, unlike the string builder's `in`, which quotes every string.

Nothing is stringly typed until the request is rendered: the operator is a `PostgrestFilterOperator` enum that carries its wire token, an `IS` check carries `bool?` (`null`, `true`, `false`) rather than the words, and an operand is kept as the value the caller passed until it is rendered.

Comparison, boolean `IS`, `isNull` and `raw` operators come with it. The remaining operators are 2/7.

## Summary by CodeRabbit

- **New Features**
  - Added typed PostgREST column expressions for filtering and ordering.
  - Added comparison, nullability, boolean, raw, and composable filters.
  - Added support for combining, grouping, and negating filters.
  - Added rendering for strings, numbers, booleans, dates, enums, lists, and maps.
- **Documentation**
  - Updated SDK capability coverage for the new typed query and filter APIs.
- **Tests**
  - Added comprehensive coverage for column expressions, filter composition, escaping, and serialization.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_column_expression.dart
+++ b/packages/postgrest/lib/src/postgrest_column_expression.dart
@@ -1,0 +1,183 @@
+part of 'postgrest_typed_builder.dart';
+
+/// Anything that can appear in a `select` list: a stored column, or an
+/// expression derived from one.
+///
+/// [Row] is the row type of the table the expression belongs to. Every
+/// filter and ordering built from the expression carries it, so using a
+/// column of one table against a query on another is a compile error rather
+/// than a PostgREST 400.
+///
+/// [Value] is the Dart type the expression produces. Operators require a
+/// matching operand, so `.eq('seven')` on an `int` column does not compile.
+///
+/// Stored columns are declared as [PostgrestColumn] or
+/// [PostgrestNullableColumn].
+@experimental
+sealed class PostgrestColumnExpression<Row, Value extends Object> {
+  const PostgrestColumnExpression._();
+
+  /// The text PostgREST expects, in a `select` list or on the left of an
+  /// operator.
+  String get expression;
+
+  @override
+  String toString() => expression;
+}
+
+/// A column expression that can also sit on the left of a filter operator.
+///
+/// Each PostgREST operator is a method here, returning a [PostgrestFilter]
+/// that [PostgrestTypedFilterBuilder.where] applies and that composes with
+/// `&`, `|` and [PostgrestFilter.not].
+///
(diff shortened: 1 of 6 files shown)


mandarini (approved):

grdsdev (approved):

grdsdev (approved):
Reviewed the diff (filter-tree/comparison rendering, AND/OR flattening, escaping). Well-tested, correct. Minor style-only nit found (a wire-format inconsistency in null-array-element rendering vs the existing string-based builder, currently dead code) — worth a look when #1795 activates that path, not blocking here.

PR #1794, opened 2026-09-07
feat(postgrest): add the remaining filter operators

spydon:
Stack 2/7 for SDK-1741. Base: `lukasklingsbo/sdk-1741-1-column-expressions`.

Pattern, list, array, range, JSON and text search operators, completing the operator surface.

- `contains`/`containedBy`/`overlaps` come in three forms because the operand literal is chosen by the column's Postgres type, not by the operator: `{a,b}` for an array, `[1,3)` for a range, `{"a":1}` for `jsonb`. The array form is an extension on `PostgrestFilterableExpression<Row, List<E>>`, so it only exists on array columns; the range and JSON forms (`containsRange`, `containsJson`, …) have no Dart type to constrain on and live on every filterable expression. `containsJson` takes decoded JSON and encodes it, matching the string builder's `contains`.
- `inFilter` is its own node: its `in.(…)` parentheses must reach the server literal, so a group leaves them alone while still escaping the members. Named `inFilter` like the string builder and the realtime stream, since `in` is a keyword. There is no `notIn`; `.not()` covers it.
- `textSearch` lives on every filterable expression rather than on text columns only, because `supabase_typegen` maps `tsvector` to `Object`.
- `matchRegex`/`imatchRegex` keep the string builder's names.

## Summary by CodeRabbit

- **New Features**
  - Added typed filtering for `IN`, pattern matching, regular expressions, arrays, ranges, JSON containment, and full-text search.
  - Added support for configurable full-text search modes.
  - Added handling for list values, nulls, and special characters in filter expressions.

- **Tests**
  - Added coverage for new filter operators, escaping behavior, grouped filters, and text-search variants.

- **Documentation**
  - Updated the SDK compliance matrix with the new filtering capabilities.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_column_expression.dart
+++ b/packages/postgrest/lib/src/postgrest_column_expression.dart
@@ -91,6 +91,120 @@ base mixin PostgrestFilterableExpression<Row, Value extends Object>
   /// beneath `|` or [PostgrestFilter.not] unless it is valid inside `or=(…)`.
   PostgrestFilter<Row> raw(String operand) =>
       PostgrestFilter._raw(expression, operand);
+
+  /// Only rows where this expression equals one of [values].
+  ///
+  /// There is no `notIn`; negate instead: `Books.id.inFilter([1, 2]).not()`
+  /// renders `id=not.in.(1,2)`.
+  ///
+  /// An empty [values] matches no rows.
+  PostgrestFilter<Row> inFilter(List<Value> values) =>
+      PostgrestFilter._comparison(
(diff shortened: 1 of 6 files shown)


mandarini (approved):

grdsdev (approved):

grdsdev (approved):
Reviewed the diff (in/like/ilike/regex, array/range/json containment, text search operators). Operator tokens match PostgREST wire format, escaping is tested for top-level and grouped (or=(...)) cases. No issues found.

PR #1795, opened 2026-09-07
feat(postgrest): route where, order and select through column expressions

spydon:
Stack 3/7 for SDK-1741. Base: `lukasklingsbo/sdk-1741-2-filter-operators`.

```dart
final books = await client
    .table(Books.table)
    .select([Books.id, Books.title])
    .where((Books.isDone.eq(false) & Books.priority.gt(3)) | Books.id.eq(7))
    .order(Books.dueDate.desc().nullsFirst());
```

`where` takes a `PostgrestFilter<Row>`, so a whole boolean expression is one call instead of a chain of per-operator methods, and a filter that depends on runtime state is assembled as a value. `order` takes a `PostgrestOrdering<Row>`; a bare orderable expression is one with no direction, so `order(Books.title)` sends `order=title` and lets PostgREST apply its default, `.asc()`/`.desc()` are for when it matters and chain into `.nullsFirst()`/`.nullsLast()`. Repeated `order` calls merge into the single `order` parameter PostgREST reads. `select` takes a list of column expressions instead of a string, since the typed surface is about not spelling columns as strings; `select()` still means `*`.

A `PostgrestOrdering` holds a `SortDirection` and a null placement enum, so the `asc`/`desc`/`nullsfirst`/`nullslast` tokens appear only where the key is rendered.

Two `@internal` hooks on the string builders carry a rendered filter and a rendered order key to the request without going through methods that would re-render them.

`supabase_typegen` emits `PostgrestColumn<BooksRow, int>` for a `NOT NULL` column and `PostgrestNullableColumn` for a nullable one, so `isNull()` exists exactly where it can match. The generator and `where` move together because the generated code has to compile against whichever `where` exists.

`TableColumn` and the `ColumnFilter` hierarchy still exist for the realtime stream; they go in 4/7.

## Summary by CodeRabbit

- **New Features**
  - Added strongly typed column selection for queries and transformations.
  - Added composable filtering with typed expressions, including nested AND/OR logic.
  - Added fluent ordering with ascending/descending direction and null placement controls.
  - Generated Dart clients now use typed, nullability-aware column definitions.

- **Documentation**
  - Updated examples and API guidance for typed selection, filtering, and ordering.
  - Clarified nullable-column behavior and updated generated-code usage instructions.

### DIFF
--- a/packages/postgrest/lib/src/postgrest.dart
+++ b/packages/postgrest/lib/src/postgrest.dart
@@ -125,7 +125,7 @@ class PostgrestClient {
   ///
   /// Unlike [from], results are converted into the row type of [table]
   /// instead of raw `Map<String, dynamic>` data, and filters are built from
-  /// [TableColumn]s, which makes them compile-time checked.
+  /// [PostgrestColumn]s, which makes them compile-time checked.
   ///
   /// ```dart
   /// final List<Book> books = await client
--- a/packages/postgrest/lib/src/postgrest_column_expression.dart
+++ b/packages/postgrest/lib/src/postgrest_column_expression.dart
@@ -209,11 +209,32 @@ base mixin PostgrestFilterableExpression<Row, Value extends Object>
 
 /// A column expression that can be used as an `order` key.
 ///
-/// Stored columns mix this in. Expression kinds PostgREST rejects in `order`
-/// do not, so ordering by one is a compile error.
+/// On its own it is a [PostgrestOrdering] with no direction, so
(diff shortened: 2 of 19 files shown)


grdsdev (approved):

grdsdev (approved):
Reviewed the incremental diff. At first glance this looked like it dropped whereAny(), order(referencedTable:), and embedded-resource select() support, but confirmed those are staged removals: #1796 cleanly removes the old TableColumn/whereAny surface, and #1799 restores embedded select/order via the new typed column-expression system. Not permanent regressions.

PR #1796, opened 2026-09-07
refactor(postgrest): remove the TableColumn filter surface

spydon:
Stack 4/7 for SDK-1741. Base: `lukasklingsbo/sdk-1741-3-where-order-select`.

Deletes `TableColumn`, `TextTableColumnFilters`, the sealed `ColumnFilter` hierarchy (one class per operator) and `whereAny`. There is now exactly one way to write a typed filter.

- `.where(Books.id.eq(1))` is unchanged in spelling; the column is a `PostgrestColumn` now.
- `.whereAny([a, b])` becomes `.where(a | b)`.
- `.order(Books.title, ascending: false)` becomes `.order(Books.title.desc())`.

The three readers of the old surface move over:

- `SupabaseTypedStreamFilterBuilder.filter` takes a `PostgrestFilter<Row>`. A realtime stream applies one comparison at a time, so it reads the filter's single comparison through a new `PostgrestFilter.comparison` getter, a `PostgrestComparison<Row>` carrying the column expression itself, the `PostgrestFilterOperator` and the operand as it was given. The stream requires the column to be a stored `PostgrestColumn` and switches exhaustively over the operator enum, so a derived expression, an operator realtime does not support, or a composed or raw filter is an `ArgumentError` rather than a string that fails to match.
- `stream(primaryKey:)` and the stream's `order` take `PostgrestColumn`s.
- The typed `upsert(onConflict:)` takes the conflict target as columns rather than a string, so a column that does not exist is a compile error.

Not marked breaking: the typed surface is `@experimental` and has only shipped in a `-dev` prerelease.

## Summary by CodeRabbit

* **New Features**
  * Added access to a filter’s column, operator, and value through comparison details.
  * Upsert conflict targets now accept typed column lists, with validation for empty lists.
  * Realtime stream ordering, primary keys, and filtering now use updated typed query APIs.

* **Breaking Changes**
  * Removed legacy typed table-column and filter model APIs, including `whereAny`.
  * Realtime filtering does not support raw, composed, or negated filters.

* **Documentation**
  * Updated examples and SDK capability documentation for the revised typed APIs.


mandarini (approved):

grdsdev (approved):

grdsdev (approved):
Reviewed the diff. Confirmed no dangling references anywhere in the repo to the removed TableColumn/ColumnFilter/whereAny surface. sdk-compliance.yaml updated consistently. Realtime stream .filter() migration is a strict superset of prior behavior.

PR #1797, opened 2026-09-07
feat(postgrest): add casts and JSON paths as column expressions

spydon:
Stack 5/7 for SDK-1741. Base: `lukasklingsbo/sdk-1741-4-remove-table-column`.

`cast(PostgrestCastTarget.text)` renders `cost::text`; `jsonText('name')` renders `data->>name` and `jsonObject('n')` renders `data->n`. All three are declared on `PostgrestColumnExpression`, so they chain.

A cast is a `PostgrestDerivedExpression`, which is select position only: PostgREST 16.1 answers `cost::text=eq.10` with 200 and drops the cast from the SQL, so the filter silently compares the uncast column, and `order=cost::text.desc` is a 400. Both are therefore compile errors. `PostgrestCastTarget<T>` pairs the Postgres type with the Dart type the cast produces; `text`, `integer`, `doublePrecision` (spelled `float8`, since a type name with a space does not belong in a query string) and `boolean` ship, and any other type is reachable through the constructor.

A JSON path keeps the positions of what it was read from. On a stored column it is a `PostgrestJsonPath`, which filters, orders and is null-testable whatever the column's own nullability, since `data->>name` is `NULL` when the key is absent. On a cast it stays select-only, so filtering `cost::text->>k`, a 400, does not compile.

`jsonText` versus `jsonObject` is the `->>` / `->` distinction and it changes results: `->>` yields text, so `data->>n=eq.10` misses the row where `n` is the number `10`.

A single `_derive` hook on the base class places a derivation where PostgREST expects it, so the embedded columns in 7/7 can put a cast inside the embed's parentheses without redeclaring the accessors.

## Summary by CodeRabbit

* **New Features**
  * Added typed PostgreSQL column casts for text, integer, double-precision, and Boolean values.
  * Added JSON text and object path expressions that can be chained, filtered, sorted, and checked for null values.
  * Enabled casts and JSON paths in typed selections and ordering.
  * Preserved JSON value types when composing object-path expressions.

* **Bug Fixes**
  * Prevented unsupported JSON path filters on typed realtime streams by reporting an argument error.

### DIFF
--- a/sdk-compliance.yaml
+++ b/sdk-compliance.yaml
@@ -796,6 +796,27 @@ features:
       - PostgrestTypedQueryBuilder.count
       - PostgrestTypedTransformBuilder
       - PostgrestTypedTransformBuilder.count
+    supporting_symbols:
+      - PostgrestColumnExpression.cast
+      - PostgrestColumnExpression.jsonText
+      - PostgrestColumnExpression.jsonObject
+      - PostgrestColumn.jsonText
+      - PostgrestColumn.jsonObject
+      - PostgrestCastTarget
+      - PostgrestCastTarget.PostgrestCastTarget
+      - PostgrestCastTarget.sqlType
+      - PostgrestCastTarget.text
+      - PostgrestCastTarget.integer
+      - PostgrestCastTarget.doublePrecision
+      - PostgrestCastTarget.boolean
+      - PostgrestDerivedExpression
+      - PostgrestDerivedExpression.expression
+      - PostgrestDerivedExpression.jsonText
+      - PostgrestDerivedExpression.jsonObject
+      - PostgrestJsonPath
+      - PostgrestJsonPath.expression
+      - PostgrestJsonPath.jsonText
+      - PostgrestJsonPath.jsonObject
   database.query.rpc:
     status: implemented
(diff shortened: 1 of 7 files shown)


sdk-compliance.yaml
```diff
@@ -796,6 +796,30 @@ features:
       - PostgrestTypedQueryBuilder.count
       - PostgrestTypedTransformBuilder
       - PostgrestTypedTransformBuilder.count
+    supporting_symbols:
+      # Casts and JSON paths on the column-expression surface. There is no
+      # database.* id for either, so they are accounted for here as the
+      # select-list entries they are, without claiming a capability.
```
spydon:
```suggestion
```

mandarini (approved):

grdsdev (approved):

grdsdev (approved):
Reviewed the diff (casts and JSON paths as column expressions). Purely additive, well-tested. Note: jsonText()/jsonObject()/cast() are callable on any column type regardless of the actual Postgres column type, deferring a bad query to a runtime PostgREST error instead of a compile error — but this matches an existing pattern already in the codebase (containsJson/containedByJson), not a new gap. Non-blocking.

PR #1798, opened 2026-09-07
feat(postgrest): add aggregate functions as select-only expressions

spydon:
Stack 6/7 for SDK-1741. Base: `lukasklingsbo/sdk-1741-5-casts-json-paths`.

`sum()`, `avg()`, `min()`, `max()`, `count()` and `PostgrestDerivedExpression.countAll()` render `amount.sum()` and `count()` in a `select` list.

An aggregate is a `PostgrestDerivedExpression`, select position only, because PostgREST has no `HAVING`: filtering on an aggregate is not a request it can serve, and `order=children(amount.sum()).desc` is a PGRST100. So `Orders.amount.sum().gt(100)` does not compile.

`min()`/`max()` keep the column's type while `sum()`/`avg()` are `double`: a min of an `int` column is an `int`, but a sum can overflow it and an average is rarely integral.

Aggregates need `db-aggregates-enabled`, which is on for hosted Supabase and off by default in self-hosted PostgREST.

Two limitations are documented rather than solved here. PostgREST keys the response by the function name alone, so `[Orders.amount.sum(), Orders.quantity.sum()]` produces two `sum` keys and needs aliasing, tracked as SDK-1757. And `sum`, `avg`, `min` and `max` are `null` in the response when no row matches, which the `Value` type parameter of a select-only expression does not express; it describes a present value, and nothing decodes through it.

## Summary by CodeRabbit

- **New Features**
  - Added aggregate expressions for sums, averages, minimums, maximums, and counts in select queries.
  - Added support for counting all records without specifying a column.
  - Aggregate results preserve appropriate numeric and column value types.
  - Added support for casting aggregate results in queries.
  - Aggregate expressions can be selected but aren’t available for filtering or ordering.

- **Documentation**
  - Updated query capability documentation to include aggregate expressions.

### DIFF
--- a/sdk-compliance.yaml
+++ b/sdk-compliance.yaml
@@ -809,7 +809,13 @@ features:
       - PostgrestCastTarget.integer
       - PostgrestCastTarget.doublePrecision
       - PostgrestCastTarget.boolean
+      - PostgrestColumnExpression.sum
+      - PostgrestColumnExpression.avg
+      - PostgrestColumnExpression.min
+      - PostgrestColumnExpression.max
+      - PostgrestColumnExpression.count
       - PostgrestDerivedExpression
+      - PostgrestDerivedExpression.countAll
       - PostgrestDerivedExpression.expression
       - PostgrestDerivedExpression.jsonText
       - PostgrestDerivedExpression.jsonObject
--- a/packages/postgrest/lib/src/postgrest_column_expression.dart
+++ b/packages/postgrest/lib/src/postgrest_column_expression.dart
@@ -60,8 +60,63 @@ sealed class PostgrestColumnExpression<Row, Value extends Object> {
   /// chain [jsonText] or [cast] to reach a scalar.
   PostgrestColumnExpression<Row, Value> jsonObject(String path);
 
+  /// The sum of this expression across the group, typed [double] whatever
+  /// the column's type. Past 2^53 the value rounds silently.
+  ///
+  /// {@template postgrest_aggregate}
+  /// Select position only: PostgREST has no `HAVING` and rejects an aggregate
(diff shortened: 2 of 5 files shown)


sdk-compliance.yaml
```diff
@@ -797,9 +797,10 @@ features:
       - PostgrestTypedTransformBuilder
       - PostgrestTypedTransformBuilder.count
     supporting_symbols:
-      # Casts and JSON paths on the column-expression surface. There is no
-      # database.* id for either, so they are accounted for here as the
-      # select-list entries they are, without claiming a capability.
+      # Casts, JSON paths and aggregates on the column-expression surface.
+      # There is no database.* id for any of them, so they are accounted for
+      # here as the select-list entries they are, without claiming a
+      # capability.
```
spydon:
```suggestion
```

mandarini (approved):

grdsdev (approved):

grdsdev (commented):
Not approving yet — found two real gaps in the aggregate-function surface:

1. **Column collision**: selecting the same aggregate fn over two different columns (e.g. `[Orders.amount.sum(), Orders.quantity.sum()]`) produces two JSON response keys both named `"sum"` — PostgREST keys results by function name only, so one value silently overwrites the other with no error. There's no alias/rename mechanism to avoid this.
2. **Nullability**: `sum()`/`avg()`/`min()`/`max()` are typed as non-nullable (`double`), but PostgREST returns `null` when zero rows match. This breaks this codebase's own convention of a distinct `PostgrestNullableColumn` type for exactly this case — a caller trusting the declared type would hit a runtime type error on an empty result set.

Both are worth addressing (or at least documenting as known limitations) before this ships as anything beyond @experimental. Happy to approve once addressed.

PR #1799, opened 2026-09-07
feat(postgrest): add embedded relations to the column namespace

spydon:
Stack 7/7 for SDK-1741. Base: `lukasklingsbo/sdk-1741-6-aggregates`.

`PostgrestToOneRelation`/`PostgrestToManyRelation` sit on a table's namespace class and project a column of the related table through the embed: `Books.author(Authors.name)` renders `author(name)` in a `select` list. The relation is declared once, by hand for now; teaching `supabase_typegen` to emit them from the foreign key metadata it already parses is a follow-up, as it was for the Swift generator.

Cardinality reaches the type system because the server distinguishes it: `order=author(name).asc` is fine, while the to-many direction is PGRST118, so a `PostgrestToOneColumn` is orderable and a `PostgrestToManyColumn` is not. Neither is filterable; filtering inside an embed is out of scope, and `embeddedFilterName` (`author.name`) is what it will read.

An embedded projection renders `author(name)`, so a cast, JSON path or aggregate on one lands inside those parentheses through the `_derive` hook from 5/7: `Authors.books(Books.id).count()` renders `books(id.count())`, where `books(id).count()` would be a 200 with the `.count()` silently dropped. A JSON path on an embedded column stays an embedded column, keeping its position.

Projections compose, so `Orders.todo(Todos.orders(Orders.amount))` renders `todo(orders(amount))`.

## Summary by CodeRabbit

* **New Features**
  * Added typed embedded-relation support for selecting columns from related records.
  * Supports to-one and to-many relationships, including nested embeds.
  * Embedded selections support aggregates, casts, JSON paths, and chained derivations.
  * To-one embedded columns can be ordered; to-many embedded columns remain non-orderable.

* **Documentation**
  * Updated SDK capability documentation with typed relationship embedding syntax.

* **Tests**
  * Added coverage for embedded projections, nesting, filtering, ordering, and derived expressions.

### DIFF
--- a/packages/postgrest/lib/src/postgrest_embedded_relation.dart
+++ b/packages/postgrest/lib/src/postgrest_embedded_relation.dart
@@ -1,0 +1,114 @@
+part of 'postgrest_typed_builder.dart';
+
+/// A to-one embedded relation (many-to-one or one-to-one) of the table whose
+/// rows are [Row], pointing at the table whose rows are [Target].
+///
+/// Calling it with a column of the target projects that column into the
+/// parent's `select` list:
+///
+/// ```dart
+/// class Orders {
+///   static const todo = PostgrestToOneRelation<Order, Todo>('todo');
+/// }
+///
+/// Orders.todo(Todos.title).expression // todo(title)
+/// ```
+///
+/// Projections of a to-one embed can be ordered by, unlike those of a
+/// [PostgrestToManyRelation].
+@experimental
+final class PostgrestToOneRelation<Row, Target> {
+  /// Creates a relation PostgREST addresses as [name], including any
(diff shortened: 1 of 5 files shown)


mandarini (approved):

grdsdev (approved):

grdsdev (approved):
Reviewed the diff (embedded relations in the column namespace). Type hierarchy, to-one vs to-many orderability restrictions, and nesting all check out against PostgREST semantics and the tests. Two non-blocking notes: (1) this PR deletes the test 'a filter carries the row type of its table' with no replacement coverage — worth restoring elsewhere since it's the only check on PostgrestFilter's row-type generic; (2) PostgrestToOneColumn/PostgrestToManyColumn duplicate several fields/methods verbatim, could share a private mixin.

PR #1802, opened 2026-09-07
refactor(auth)!: derive ErrorCode wire values from the enum name

spydon:
## Summary
- `ErrorCode` no longer carries a manually assigned wire value. The `code` field
  becomes a getter backed by the shared `snakeCase` extension from
  `supabase_common`, the same way the other enums in the SDK encode their wire
  values.
- Six values whose name did not match their wire value are renamed so the
  derivation holds: `mfaPhoneEnrollDisabled`, `mfaPhoneVerifyDisabled`,
  `mfaTotpEnrollDisabled`, `mfaTotpVerifyDisabled`, `mfaWebauthnEnrollDisabled`
  and `mfaWebauthnVerifyDisabled` become `mfaPhoneEnrollNotEnabled`,
  `mfaPhoneVerifyNotEnabled`, `mfaTotpEnrollNotEnabled`,
  `mfaTotpVerifyNotEnabled`, `mfaWebauthnEnrollNotEnabled` and
  `mfaWebauthnVerifyNotEnabled`, matching `mfa_*_not_enabled` on the wire.
- Drops the now non-existent `ErrorCode.ErrorCode` constructor from
  `sdk-compliance.yaml`.
- Adds `error_code_test.dart`, which pins the full list of 86 wire values so a
  future rename cannot silently change the protocol, and covers the `fromCode`
  round-trip.

All 86 wire values are byte-for-byte identical to what the enum produced before,
verified by dumping `ErrorCode.values.map((e) => e.code)` on both revisions and
diffing.

## Test plan
- `dart analyze packages` is clean.
- `dart test test/src/types/error_code_test.dart` passes.
- The rest of the `supabase_auth` suite shows the same live-GoTrue failures on
  this branch as on `main` (they fail in different combinations run to run and
  pass when the files are run in isolation).

## Summary by CodeRabbit

* **Breaking Changes**
  * Updated authentication error code names for MFA enrollment and verification states from “Disabled” to “Not Enabled.”
  * Error code values are now automatically derived from enum names while preserving their wire-format strings.

* **Tests**
  * Added comprehensive coverage for error code conversion, lookup, round-tripping, and unknown values.

### DIFF
--- a/packages/supabase_auth/lib/src/types/error_code.dart
+++ b/packages/supabase_auth/lib/src/types/error_code.dart
@@ -1,276 +1,275 @@
 // All error codes from the Supabase Auth API. The whole list can be found here:
 // https://github.com/supabase/auth/blob/master/internal/api/errorcodes.go
 import 'package:collection/collection.dart';
+import 'package:supabase_common/supabase_common.dart';
 
 /// A machine-readable identifier for an [AuthApiException], returned by the
 /// Supabase Auth server as the `error_code` field.
 enum ErrorCode {
   /// An unexpected server failure, such as an internal error.
-  unexpectedFailure('unexpected_failure'),
+  unexpectedFailure,
 
   /// A request value failed validation.
-  validationFailed('validation_failed'),
+  validationFailed,
 
   /// The request body could not be parsed as JSON.
-  badJson('bad_json'),
+  badJson,
 
   /// The email address is already in use by another user.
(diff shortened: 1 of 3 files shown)


spydon:
Not worth the breaking change `mfaPhoneEnrollDisabled` -> `mfaPhoneEnrollNotEnabled` etc

PR #1805, opened 2026-09-08
feat(auth)!: let AuthClient own session persistence through one storage

spydon:
> [!NOTE]
> Stacked on #1804, which adds `persistSession` to `AuthClient`. The base flips to `main` once that merges.

Breaking refactor of session persistence, tracked in SDK-1750.

Credit to @Vinzent03: the design here is the one from #1087, opened in November 2024. A single storage interface for the session and the pkce verifiers, the client owning persistence like auth-js, `persistSession` and `storageKey` on the auth options, and dropping the no-op `CancelableOperation` around the recovery all come from that PR. It could not be rebased after the package rename and the v3 cleanups, so this is a fresh implementation of the same idea.

Session persistence lives in `supabase_flutter`: `SupabaseAuth` listens to `onAuthStateChange` and writes the session to a `LocalStorage`, while the pkce code verifiers go to a separate `AuthAsyncStorage` passed as `pkceAsyncStorage`. Customizing storage means implementing two interfaces, and the plain `supabase` package has no session persistence at all.

## What is the new behavior?

`AuthClient` owns persistence, as in auth-js:

- One `AuthAsyncStorage` (`AuthClientOptions.asyncStorage`, renamed from `pkceAsyncStorage`) holds the session and the code verifiers. Its methods take positional parameters.
- With `persistSession` the client writes the session on every change and restores it on construction. Writes are queued in order and a failed write is logged rather than thrown. `AuthClient.initialized` completes once the restore is done; `Supabase.initialize` awaits it, so `currentSession` is set when it returns, as before. An expired session is refreshed after `initialized` completes, so the wait never touches the network.
- `storageKey` names the session key, defaulting to `defaultPersistSessionKey(url)`. It also prefixes the pkce verifier keys and names the broadcast channel. `PKCEVerifierStore` still reads and cleans up verifiers under the old `supabase.auth.token` prefix so an in-flight flow completes across the upgrade.
- `LocalStorage`, `EmptyLocalStorage`, `SharedPreferencesLocalStorage` and `FlutterAuthClientOptions.localStorage` are gone. `persistSession: false` replaces `EmptyLocalStorage`.
- `SharedPreferencesAuthAsyncStorage` stays the Flutter default. On web it writes to `window.localStorage` so the session is shared with supabase-js, and it decodes a verifier that `SharedPreferencesAsync` JSON-encoded there before. On other platforms a value written by v2 through the legacy `SharedPreferences` API is moved over on first read, once per key, and a removed value retires its legacy entry so a signed-out session cannot come back.
- `SupabaseAuth` in `supabase_flutter` loses the storage handling, and the no-op `CancelableOperation` around the old recovery goes with it. `package:async` is no longer a dependency of `supabase_flutter`.
- `initialSession` is emitted to every new subscriber of `onAuthStateChange` as its first event, with the session at that moment, the way auth-js and supabase-swift do. It waits for the restore, and events that fire in the meantime are held back so the initial one stays first. The stream no longer replays its latest event or error to late subscribers; a listener attached after a sign-in gets `initialSession` with that session instead of a replayed `signedIn`. `AuthClient.initialized` remains the await point for the restore, since the event now describes the moment of subscription rather than startup.

The ticket asked for the storage write to be awaited before notifying subscribers. The write is queued instead, so `_saveSession` stays synchronous and the session version checks around refreshes keep their meaning. The write still happens at the same time, only the event no longer waits for it.

## Breaking changes

See the two new sections in `MIGRATION.md`. In short: `LocalStorage` and its implementations are removed, `pkceAsyncStorage` is `asyncStorage`, `AuthAsyncStorage` methods are positional, and the pkce verifier keys move under `storageKey`.

## Tests

- `packages/supabase_auth/test/session_persistence_test.dart`: write on sign-in, removal on sign-out, restore in a new client, custom key, non-persisting client, expired session refresh and sign-out, corrupt values, failing storage.
- `packages/supabase_auth/test/pkce_flow_test.dart`: legacy prefix fallback and cleanup.
- `packages/supabase_flutter/test/storage_migration_test.dart`, `storage_test.dart`, `storage_web_test.dart` (browser): the unified storage on VM and web, including the v2 migrations.
- The remaining `supabase_flutter` tests are ported to `asyncStorage`. The full `supabase_auth`, `supabase` and `supabase_flutter` suites pass locally, the browser ones on Chrome.

## Compliance matrix

`client.session_management.custom_storage` and `persist_session` are reconciled. Symbol and drift checks pass locally.

Closes SDK-1750

## Summary by CodeRabbit

- **New Features**
  - Sessions and PKCE verifiers now use shared, configurable asynchronous storage.
  - Added customizable storage keys and an initialization signal for session restoration.
  - Flutter uses asynchronous shared-preferences storage by default, with legacy-value migration.
  - Expired restored sessions can refresh automatically, and PKCE flows continue to support legacy stored values.

- **Breaking Changes**
  - Replaced `localStorage` and `pkceAsyncStorage` with `asyncStorage`.
  - Updated custom storage methods to use positional parameters.
  - Removed legacy local-storage APIs and related configuration options.

- **Documentation**
  - Added migration guidance and updated custom storage examples.


Tr00d (approved):

packages/supabase/lib/src/supabase_client_options.dart
```diff
@@ -64,18 +67,28 @@ class AuthClientOptions {
...
+  /// Whether the session is written to [asyncStorage] whenever it changes and
+  /// restored from there when the client is created.
   ///
-  /// The client stores nothing itself. On web a persisted session is kept in
-  /// sync across the tabs of the same project, so a sign-in or sign-out in one
-  /// tab reaches the others. Leave it false for a client that must keep its
-  /// own session, such as one created with the service role key next to the
-  /// user's client. `supabase_flutter` persists the session and defaults it to
-  /// true.
+  /// Await `AuthClient.initialized` to know when the restore is done. On web a
```
Vinzent03: Shouldn't users of the library rather wait for the `initialSession` event rather than awating that method?

packages/supabase_auth/lib/src/auth_client.dart
```diff
@@ -172,7 +191,35 @@ class AuthClient {
...
+
+  /// Completes once the session persisted by an earlier run has been restored.
+  ///
+  /// When the session is persisted, [currentSession] holds the restored
+  /// session and the [AuthChangeEvent.initialSession] event has been emitted
+  /// by then. A restored session that has expired is refreshed in the
+  /// background, which [onAuthStateChange] reports like any other refresh.
+  /// Completes right away, without emitting an event, when the session is not
+  /// persisted.
+  Future<void> get initialized => _initialized.future;
```
Vinzent03, line 218: Isn't this getter here duplicate to an initialSession event?

packages/supabase_auth/lib/src/auth_client.dart
```diff
@@ -1699,22 +1749,135 @@ class AuthClient {
...
+    _queueStorageWrite((storage) => storage.removeItem(_storageKey));
+  }
+
+  /// Runs [write] after the storage writes queued before it, when the session
+  /// is persisted.
+  ///
+  /// A failed write is logged rather than thrown: the session change it was
+  /// mirroring has already happened, and the user only loses the session at
+  /// the next start.
+  void _queueStorageWrite(
```
Vinzent03: What exactly is the reason for the write queue? At the spot, I cannot come up with an actual case where this could matter.

Vinzent03 (commented):
First of all I'm very happy this can finally land. Didn't expect it to take 2 years with even being designed non breaking in the beginning.

I'm a bit unsure about creating an initial session event when persist session is off. It has no real value. The only benefit of it I see is for a flutter application to quickly turning off the persistence without having to change their logic regarding an initial session.

packages/supabase_auth/lib/src/auth_client.dart
spydon: Kind of, but this really only serves an internal purpose (or for people building custom clients), it's there for the initialization to know when it is done. We don't really want to subscribe to the stream, filter and then unsubscribe just for that. This is how it is done in auth-js too for example.

spydon:
> I'm a bit unsure about creating an initial session event when persist session is off. It has no real value. The only benefit of it I see is for a flutter application to quickly turning off the persistence without having to change their logic regarding an initial session.

I changed it in 49714d4b so that the event now means "startup is done, here is the current state" rather than "a session was restored", so it's now always emitted instead. That matches auth-js, where INITIAL_SESSION always fires, and it also leads to that we can remove the @internal call in supabase_flutter.

packages/supabase_auth/lib/src/auth_client.dart
spydon: I changed this again, initialSession now follows auth-js and supabase-swift: every new subscriber of `onAuthStateChange` gets it as its first event with the session at that moment, after the restore has finished. So the event describes the moment of subscription, and initialized is the only thing that describes startup. Supabase.initialize awaits it so `currentSession` is set when it returns, and a pure Dart client awaits it for the same reason. The stream no longer replays its latest event either, since the per-subscriber initial event covers that case.

spydon:
> > I'm a bit unsure about creating an initial session event when persist session is off. It has no real value. The only benefit of it I see is for a flutter application to quickly turning off the persistence without having to change their logic regarding an initial session.
> 
> I changed it in [49714d4](https://github.com/supabase/supabase-flutter/commit/49714d4be787bdf160174a0b78099db3976f7109) so that the event now means "startup is done, here is the current state" rather than "a session was restored", so it's now always emitted instead. That matches auth-js, where INITIAL_SESSION always fires, and it also leads to that we can remove the @internal call in supabase_flutter.

They really should support threads for comments directly on the PRs too. 😅 

I re-did it so that the event is sent per subscriber instead, which is what auth-js and supabase-swift do. So each listener will get `initialSession` with the current session or null when it subscribes. That removes the special case in supabase_flutter, gives hand-built clients the same first event as `Supabase.initialize`, and a Flutter app that turns persistence off keeps its splash logic working.

Great review, as always!

Vinzent03:
I was not aware of the replay subject meaning sign in events would replay. Great that is fixed. 
Sounds good now, even though the initial session event has now a magnificent different meaning than before. But good if that is the same behavior now as for the other sdks.

This still means that the `initialized` future is not needed for flutter users, but may be helpful for other environments without having to resort listening to the stream. Sounds good.

spydon:
> I was not aware of the replay subject meaning sign in events would replay. Great that is fixed.
Sounds good now, even though the initial session event has now a magnificent different meaning than before. But good if that is the same behavior now as for the other sdks.

Yeah, I think it's worth it having this aligned. :)

> This still means that the `initialized` future is not needed for flutter users, but may be helpful for other environments without having to resort listening to the stream. Sounds good.

It is not needed in the public surface for the common Flutter users, but the initialization still uses it under the hood.
So if anyone wants to create their own clients I think it could be good to keep exposing it.

Issue #1174, opened 2025-05-21
[kakao login] Issue when using Supabase Kakao login oauth using Flutter on IOS


merche-o:
### Steps to reproduce

I'm implementing Kakao login in my app using Supabase and Flutter. Here’s my implementation, which I followed based on the Supabase [documentation](https://supabase.com/docs/guides/auth/social-login/auth-kakao?queryGroups=language&language=flutter&queryGroups=environment&environment=client):

```dart
await _client.auth.signInWithOAuth(
  OAuthProvider.kakao,
  redirectTo: redirectUrl,
  authScreenLaunchMode: LaunchMode.inAppWebView,
);
```
The redirectUrl is a deep link that should redirect back to my app.

### Expected results

The Kakao login should run smoothly and automatically redirect users to the app’s home page after successful authentication.

### Actual results

When using an in-app browser (which is required by Apple’s guidelines), I encounter a blank webview and the following error in my logs:

```
PlatformException(Error, Error while launching https://ldtimmwvdiqiuprrgset.supabase.co/auth/v1/authorize?provider=kakao&redirect_to=com.nectrlimited.scout%3A%2F%2Fkakaologin&flow_type=pkce&code_challenge=APN0TMO63GO3jY8a0YD0jCHkuvDd6fwn51WHaMeGFes&code_challenge_method=s256, null, null)
```

To proceed with login, you must manually dismiss the blank webpage using the Done button.

(FYI i don't have any issue when using external browser)

### Code sample

<details open><summary>kakao oauth</summary>

```dart
`  await _client.auth.signInWithOAuth(
        OAuthProvider.kakao,
        redirectTo: redirectUrl,
        authScreenLaunchMode: LaunchMode.inAppWebView,
      );
```

</details>

gagik:
Also running into this when using the Keycloak Auth provider. Seems generally related to the OAuth flow.

gagik:
Found a hacky workaround for the time being which can serve as basis for a fix. It also seems like while the deeplink does get handled by supabase (`handle deeplink uri` is logged) and the user events get called, the `auth.signInWithOAuth` Future never actually resolves.

It's possible to manually close the webview by listening to auth state changes and closing accordingly with something like:
```dart
  final listener = supabase.auth.onAuthStateChange.listen((state) async {
      if (state.event == AuthChangeEvent.signedIn) {
        await closeInAppWebView();
      }
      /** Other events */
    });

  unawaited(
[...]

PR #1807, opened 2026-09-08
feat(supabase_flutter): pluggable OAuth launcher + opt-in supabase_flutter_web_auth

spydon:
## What

Makes how `signInWithOAuth`/`signInWithSSO`/`linkIdentity` open their sign-in
URL pluggable, and ships the improved behavior as a new, separate, opt-in
package instead of a breaking change to `supabase_flutter` itself.

- Add an `OAuthLauncher` extension point
  (`FlutterAuthClientOptions.oauthLauncher`), defaulting to
  `UrlLauncherOAuthLauncher`, which preserves today's `url_launcher`-based
  behavior byte-for-byte (same launch modes, same Google-on-Android
  external-browser workaround). No breaking change, no new dependency.
- Add a new package, `supabase_flutter_web_auth`, providing
  `FlutterWebAuth2OAuthLauncher`: runs OAuth/SSO/identity-linking through a
  system web authentication session (`ASWebAuthenticationSession` on
  iOS/macOS, Custom Tabs on Android/desktop) via `flutter_web_auth_2`.

## Why

`url_launcher`'s in-app browser never dismisses itself after an OAuth
redirect returns to the app, leaving the user on a blank page after a
successful sign-in (#1174). A system web authentication session fixes this
and shares cookies with the system browser for SSO, but pulling
`flutter_web_auth_2` (and, on Linux/Windows, its own embedded-webview
dependency) into `supabase_flutter` directly would force every app to carry
that dependency weight and would be a breaking change (drops
`authScreenLaunchMode`/`launchMode`, requires Android manifest changes).

Making the launcher pluggable gets the fix to apps that want it without
touching the dependency graph or the public API of apps that don't.

## Notes

- `supabase_flutter_web_auth`'s launcher drops the hardcoded
  "force external browser for Google on Android" workaround. Verified
  against Google's OAuth policy
  (https://developers.google.com/identity/protocols/oauth2/resources/best-practices):
  it bans routing through an embedded user-agent under the app's control,
  and Custom Tabs (the only mode `flutter_web_auth_2` uses on Android) runs
  as Chrome's own sandboxed process, so it already satisfies the policy.
- Verified the dependency isolation actually holds in this pub workspace
  (shared `pubspec.lock`): `packages/supabase_flutter/example`'s own
  `.flutter-plugins-dependencies` only lists `app_links`/
  `shared_preferences`/`url_launcher` after `pub get`, while
  `packages/supabase_flutter_web_auth/example` correctly picks up
  `flutter_web_auth_2`, `desktop_webview_window`, `jni`, etc.
- The new package's example app does not check in native platform folders;
  run `flutter create .` inside it to add the ones you want to test on.

## Test plan

- [x] `flutter test` in `packages/supabase_flutter` — all existing
      OAuth/SSO/link-identity behavior unchanged (82 pre-existing tests
      pass), plus new coverage for the `OAuthLauncher` extension point.
- [x] `flutter test` in `packages/supabase_flutter_web_auth` — new coverage
      for `FlutterWebAuth2OAuthLauncher` against a fake
      `FlutterWebAuth2Platform`.
- [x] `flutter analyze` clean in both packages.
- [ ] Manual on-device verification of the native flow (not done — no
      platform folders are checked in for the new example; see Notes).

## Summary by CodeRabbit

- **New Features**
  - Added configurable launchers for OAuth, SSO, and identity-linking flows.
  - Added support for ephemeral authentication sessions and custom launch behavior.
  - Added the `supabase_flutter_web_auth` package for system authentication sessions, shared browser cookies, and automatic session dismissal.
  - Added platform-specific redirect handling for web, mobile, desktop, and universal links.

- **Documentation**
  - Added setup instructions, usage guidance, licensing information, and an example application.

- **Tests**
  - Added coverage for launcher delegation, redirects, ephemeral sessions, and authentication flows.

### DIFF
--- a/packages/supabase_flutter_web_auth/LICENSE
+++ b/packages/supabase_flutter_web_auth/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Supabase Community
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/supabase_flutter_web_auth/LICENSE
+++ b/packages/supabase_flutter_web_auth/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2026 Supabase Community
+Copyright (c) 2020 Supabase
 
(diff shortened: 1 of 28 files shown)


Tr00d (approved):

packages/supabase_flutter_web_auth/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Supabase Community
```
Vinzent03: You may have more info, but sure this shouldn't be just `Supabase` here instead of the community as well?

spydon: Ah yes, good catch. It should be the same as everywhere else!

PR #1808, opened 2026-09-08
feat(postgrest): add typed ranges to the column-expression surface

spydon:
Stack 8/8 for SDK-1741, implementing SDK-1755. Base: `lukasklingsbo/sdk-1741-7-embedded-relations`.

Range columns had no Dart type: `supabase_typegen` mapped `int4range`, `tstzrange` and the rest to `Object`, so the range operators took a Postgres literal as a `String` and existed on every column. `Books.title.rangeLt('[2,25)')` compiled and failed on the server.

**`PostgrestRange<Bound>`** is the value: `closedOpen`, `closed`, `open` and `openClosed` constructors, `null` for an unbounded side, which is always exclusive since that is the form Postgres canonicalizes an omitted bound to, `empty()`, and `parse` for the literal PostgREST returns, including the quoted timestamp bounds Postgres emits (`["2024-01-01 00:00:00+00",)`). It renders through the same value renderer filters use, so a `DateTime` bound is ISO 8601 and a bound is quoted only when the literal requires it. `parse` reads the way Postgres does: an unquoted `infinity` or `-infinity` bound is unbounded, since Dart has no infinite `DateTime`, a doubled quote inside a quoted bound is a quote, and a stray bracket inside a bound is a `FormatException`.

**The range operators move onto range columns.** `contains`, `containedBy`, `overlaps`, `rangeLt`, `rangeGt`, `rangeGte`, `rangeLte` and `rangeAdjacent` are an extension on `PostgrestFilterableExpression<Row, PostgrestRange<Bound>>` and take a `PostgrestRange<Bound>`, so they exist only on range columns and the bound type is checked; `containsElement(Bound)` covers `during=cs.2024-06-01`, which had no spelling before. The string forms `containsRange`, `containedByRange`, `overlapsRange` and the `String`-taking `rangeLt` family are removed. The JSON forms stay on every filterable expression, pending SDK-1760.

**`supabase_typegen` gains a range kind** with a bound kind (`int4range`/`int8range` to `int`, `numrange` to `num`, `daterange`/`tsrange`/`tstzrange` to `DateTime`), emits `PostgrestColumn<Row, PostgrestRange<int>>`, parses the literal in the row getter and renders it back in the insert and update types with the same timestamp rules the scalar columns use: `date` bounds date-only, `timestamp` bounds as local wall time, `timestamptz` bounds in UTC. Range arrays stay `List<String>` like the other non-scalar array elements.

A hand-written namespace that types a range column as `String` or `Object` loses the range operators, with `raw('sl.[2,25)')` as the fallback. That is the contract the array operators already impose.

## Summary by CodeRabbit

- **New Features**
  - Added typed PostgreSQL range values with inclusive, exclusive, unbounded, empty, parsing, rendering, and equality support.
  - Added dedicated range filters for containment, overlap, comparison, and adjacency operations.
  - Added range serialization and deserialization for database queries and updates.
  - Added automatic generation of typed range fields for integer, numeric, date, and timestamp ranges.

- **Documentation**
  - Documented typed range columns and range filter usage.

- **Tests**
  - Added coverage for range parsing, rendering, filtering, escaping, equality, and generated models.


mandarini (approved):

grdsdev (approved):

grdsdev (commented):
Not approving yet — found a real parsing bug:

**Infinity bounds crash on read**: `PostgrestRange.parse` never special-cases the literal `infinity`/`-infinity` bound text (only handles the whole-literal `empty` and the unbounded `,)`/`(,` shorthand). Postgres commonly represents open-ended date/timestamp ranges with an explicit `infinity` bound (e.g. `[2024-01-01,infinity)`), and PostgREST returns that string as-is. The generated typegen code for `date`/`timestamp`/`timestamptz` range bounds calls `DateTime.parse` directly on the bound text, so `DateTime.parse('infinity')` throws a `FormatException` — any app reading a row with an infinite date/timestamp range bound crashes. No test currently covers this case.

Separately (lower confidence, worth a look): `_splitBounds` may not reject some malformed literals with unbalanced trailing brackets, silently absorbing a stray character into the bound text instead of throwing.

Happy to approve once the infinity-bound case is handled (and ideally tested).

PR #1809, opened 2026-09-08
feat(supabase_typegen): emit relation members from foreign key metadata

spydon:
Stack 9/9 for SDK-1741, implementing SDK-1759. Base: `lukasklingsbo/sdk-1741-8-typed-ranges`.

`PostgrestToOneRelation` and `PostgrestToManyRelation` (8/9, #1799) had to be declared by hand. `supabase_typegen` already parsed the `relationships` array of the `GeneratorMetadata` document but only used it to mark foreign key columns; it now emits the relation members.

For every foreign key whose both ends are tables of the schema, the referencing table's namespace gets a `PostgrestToOneRelation<SourceRow, TargetRow>` named after the referenced table, and the referenced table's namespace gets a `PostgrestToManyRelation<TargetRow, SourceRow>` named after the referencing table, or a to-one relation when the metadata flags the key as one-to-one:

```dart
class Books {
  static const authors = PostgrestToOneRelation<BooksRow, AuthorsRow>('authors');
}
class Authors {
  static const books = PostgrestToManyRelation<AuthorsRow, BooksRow>('books');
}

client.table(Books.table).select([Books.id, Books.authors(Authors.name)]);
// select=id,authors(name)
```

Two rules handle names that would otherwise collide:

- Two keys from the same table to the same target make the plain table name ambiguous for PostgREST (PGRST201), so the embed carries the constraint hint, `map!postgrest_table_mood_fkey`, and the members are told apart by the key they follow: `mapByMood` on the referencing side, `postgrestTableViaMood` on the referenced side.
- A member that collides with a column, another relation or a core member name goes through the same disambiguation. `By` names a key this table holds, `Via` a key the other table holds.
- A self-referential key produces no member: PostgREST cannot tell the two directions of such an embed apart and needs a computed relationship, so a generated member could never be requested.

Keys into another schema get no member, since there is no generated row type to point at. Views take part when the metadata lists a relationship for them, as `author_stats` does in the fixture.

The type names of all tables are now claimed in one pass before any table is written, in the same order as before, so the relation types can reference each other; existing goldens are unchanged apart from the new members.

## Summary by CodeRabbit

- **New Features**
  - Dart schema generation now includes typed one-to-one and one-to-many relation members for supported foreign keys.
  - Relation names are disambiguated when multiple keys connect the same tables, including self-referencing relationships.
  - Generated relation members support embedded-column selects and count modifiers.

- **Documentation**
  - Documented generated relation members and clarified limitations for cross-schema relationships and typed functions.

- **Tests**
  - Added coverage for relationship parsing, generation, naming, and embedded select behavior.

### DIFF
--- a/packages/supabase_typegen/README.md
+++ b/packages/supabase_typegen/README.md
@@ -13,6 +13,12 @@ For every table the generator emits:
   `PostgrestNullableColumn` so `isNull()` only exists where it can match, and
   range columns typed `PostgrestRange<int>`, `PostgrestRange<num>` or
   `PostgrestRange<DateTime>` so the range operators only exist on them,
+- `PostgrestToOneRelation` and `PostgrestToManyRelation` members for every
+  foreign key between two generated tables of the selected schema, named
+  after the table on the other side and carrying the constraint hint when two
+  keys point at the same table. Keys into another schema and self-referential
+  keys get no member, the latter because PostgREST needs a computed
+  relationship to embed a table into itself,
 - Dart enums for Postgres enums, with wire-name mapping.
 
 ## Usage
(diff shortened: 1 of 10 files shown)


mandarini (approved):

grdsdev (approved):

grdsdev (approved):
Reviewed the diff (relation-member codegen from FK metadata). Traced self-reference and ambiguous-FK disambiguation against the golden fixtures — correct. Only minor codegen-time efficiency/duplication nits (e.g. _writeNamespace's collision loop could reuse _uniqueMemberNames), non-blocking.

PR #1837, opened 2026-09-14
feat(supabase_typegen): introspect the database through the Supabase CLI

spydon:
## Summary

`supabase_typegen` could only read a `GeneratorMetadata` document on stdin and relied on `supabase gen types --lang dart` to produce it, which has not shipped (supabase/cli#6230 was closed and supabase/cli#6404 carries no Dart path). This adds a Dart port of the introspection of [`@supabase/postgrest-typegen`](https://github.com/supabase/sdk/tree/main/packages/postgrest-typegen), pinned to `postgrest-typegen-v0.2.2`, so the tool can produce the document itself. The database connection is delegated to the Supabase CLI:

```sh
dart run supabase_typegen --local
dart run supabase_typegen --linked
dart run supabase_typegen --project-ref <ref>
dart run supabase_typegen --db-url 'postgresql://…'
dart run supabase_typegen --local --dump-metadata
```

Stdin stays the default when no target is given, so the eventual CLI integration is untouched. Everything new lives under `lib/src/introspection/`, `lib/introspection.dart`, `test/introspection/`, the tool scripts and the drift workflow, and the README states that it will be removed once the CLI ships Dart support.

## How it connects

The eleven introspection queries are folded into one `SELECT` with a `json_agg` sub-select per collection and run through a single `supabase db query --output json --file <statement>` call. The CLI resolves and authenticates the connection: `--local` uses the running stack, `--linked` and `--project-ref` go through the Management API with the `supabase login` credentials and need no database password, and `--db-url` is passed straight through. The package has no database driver; `postgres` is only a dev dependency for seeding the parity database.

Failures produce actionable messages: the CLI missing from `PATH` (with the install link), not logged in (`supabase login` or `SUPABASE_ACCESS_TOKEN`), no linked project (`supabase link` or `--project-ref`), and everything else surfaces the CLI's own message with its colour codes and progress line stripped, which already covers "run `supabase start`" and network restriction hints. Passing more than one target is a usage error.

## What is ported

- The eleven SQL builders, one file each, taking the schema filter alone, matching the trimmed builders of supabase/sdk#180. The view definition rewrites render from a table, matching supabase/sdk#179. The `pg-format` literal escaping is included for the schema filter.
- The view relationship expansion (cartesian product over view key dependencies).
- The ordering pass, with a stable merge sort because the TypeScript sort relies on `Array.prototype.sort` being stable for composite primary keys and tied relationships.
- `localeCompare` semantics: the ICU root collation for printable ASCII (punctuation, digits, then letters with case as the last tiebreaker, lower case first), verified against Bun and Node. Non-ASCII code units sort after ASCII by code unit, which only affects collection order for accented names.

Following 0.2.2 also brings in the changes since 0.2.0: `is_insert_enabled` and `is_update_enabled` on views, trigger and rule backed view columns marked updatable, and virtual generated columns counted as generated. The fixture and goldens are regenerated accordingly; the join view with an `INSTEAD OF INSERT` trigger now gets a `BookSubmissionsInsert` type, and the two tests asserting the old behaviour were updated.

## Verification

- Parity test (`test/introspection/parity_test.dart`): seeds a fresh `postgres:15` with `test/fixtures/seed.sql` when the database is empty, introspects it through the CLI unfiltered and restricted to `public`, and asserts the document holds the same records as `test/fixtures/generator_metadata.json`, collection by collection. Records are compared as maps, so key order is irrelevant; record order within a collection is checked since the document is sorted. It also runs the binary and checks `--dump-metadata` yields the fixture records and `--output -` reproduces `test/goldens/supabase_schema.dart`. The test skips unless `SUPABASE_TYPEGEN_PARITY_DATABASE_URL` is set; `test.yml` starts the container and installs the CLI for the `supabase_typegen` matrix entry.
- Drift guard (`tool/check_introspection_drift.ts`): fetches the TypeScript builders of the pinned supabase/sdk revision from GitHub, renders every query for three filter scenarios, and compares byte for byte with `dart run tool/dump_introspection_sql.dart`. The new `typegen-drift.yml` runs it on PRs touching the package and weekly against supabase/sdk `main` with `--latest`, so an upstream SQL change surfaces as a failing scheduled run. Bumping the pin is: run the script with `--ref`, port the printed diff, regenerate the fixture with `tool/regenerate_fixture.ts --source <sdk checkout>`, update `postgrestTypegenRevision`.
- Unit tests for the literal escaping, the schema filter and builder conditionals, the relationship expansion (ported from upstream), the sort and collation, the document assembly against a fake query runner, and the CLI runner against a fake `supabase` script (flags, JSON parsing, not installed, not logged in).
- Manually ran `--local` against the running local stack and the error paths (not logged in, not linked, CLI missing from `PATH`, local stack down, conflicting flags).

## Notes

- The direct modes require a Supabase CLI with `db query` (2.116 or newer). `--db-url` against a server without TLS needs `sslmode=disable` in the connection string, as the CLI requires TLS otherwise.
- `oven-sh/setup-bun` is a new third-party action, pinned by SHA like the others in this repo.

Closes SDK-1834.

## Summary by CodeRabbit

* **New Features**
  * Added database introspection through local, linked, project-reference, or database URL connections.
  * Added `--dump-metadata` to output raw generator metadata.
  * Added flexible output options for writing generated code to a file or standard output.
  * Improved support for view relationships and insertable views, including generated insert types.

* **Documentation**
  * Updated usage guidance for connection modes, authentication, TLS settings, and output options.

* **Bug Fixes**
  * Improved metadata ordering and relationship handling for more consistent generated output.


Tr00d (approved):

packages/supabase_typegen/lib/src/introspection/sql/views_key_dependencies_sql.dart
```diff
@@ -0,0 +1,200 @@
...
+      replace(
+      replace(
+      replace(
+      replace(
+      replace(
+      replace(
+      replace(
+      replace(
+      replace(
+        view_definition::text,
```
spydon: There must be a cleaner way to do this?

PR #1844, opened 2026-09-16
fix(auth): parse OAuth authorization details for users without an email

AndroidPoet:
## Summary

`auth.oauth.getAuthorizationDetails` throws a `FormatException` for a signed-in user who has no email, so the consent screen cannot load for users who signed up with a phone number or signed in anonymously.

Auth fills the `user` object of that response with `Email: user.Email.String()` on a field tagged `json:"email,omitempty"` ([`internal/api/oauthserver/authorize.go`](https://github.com/supabase/auth/blob/4eee58f296d9698a1c2c0ae14d7a0b379c7622d3/internal/api/oauthserver/authorize.go#L53-L57)), so `email` is left out when the user has none:

```json
{"authorization_id": "...", "redirect_uri": "...", "client": {"id": "...", "name": "..."}, "user": {"id": "39f28ab1-..."}, "scope": "email"}
```

`OAuthAuthorizingUser.fromJson` requires `email` to be a string, so it rejects the whole response. In v2 the same body parsed, because `details.user` was a `User`, whose `email` is `String?`. auth-js also types `email` as `string`, but it returns the body without validating it, so its callers get `undefined` instead of an error.

`OAuthAuthorizingUser.email` is now a `String?` and optional in the constructor. `fromJson` still throws when `email` is present but is not a string.

## Compatibility

Against v2 this is not a breaking change, since `User.email` was already nullable. The `MIGRATION.md` entry "`OAuthAuthorizationDetailsResponse.user` is an `OAuthAuthorizingUser`" says code that only reads `id` and `email` needs no change, and with this fix that holds for the types too. Code written against the v3 dev releases that treats `email` as a `String` needs a null check.

## Test

- "throws FormatException when the user email is missing", added with `OAuthAuthorizingUser` in #1663, is now "parses a user without an email" and expects `user.email` to be `null`. I reversed it because a missing `email` is what Auth sends for these users.
- The new "throws FormatException when the user email is not a string" keeps the type check covered.
- The new integration test "get authorization details for a user without an email" signs up a phone user on the local stack and fetches the details of a new authorization. On `main` it fails with `FormatException: Expected the user id and email to be strings, got String and Null`; with the fix it passes.

I also checked an anonymous user by hand against the same stack (Auth v2.192.0). The response contained `"user": {"id": "..."}` with no `email`.

Run in `packages/supabase_auth` against `supabase start` with CLI 2.109.1:

```
$ dart format lib test -l 80 --set-exit-if-changed   # 0 changed
$ dart analyze --fatal-infos .                        # No issues found!
$ dart test --concurrency=1                           # 574 tests, all passed
$ dart test -p chrome test/browser                    # 5 tests, all passed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth authorization responses now support users without an email address.
  * Phone-based sign-up flows with a null email are handled correctly.
  * Invalid non-text email values continue to produce an error.


spydon (approved):
Thanks for you contribution

PR #1845, opened 2026-09-17
refactor(auth)!: return Session from the auth methods that always issue one

spydon:
## Summary

`AuthResponse` has a nullable `session` and a nullable `user`, and every signing-in method on `AuthClient` returned it. For most of those methods a null session is not a reachable state, so callers wrote `response.session!` for an outcome that cannot happen, and three call sites in the client asserted the same invariant at runtime before using `!` themselves.

Resolves SDK-1874.

### Return `Session` directly

- `signInAnonymously`, `signInWithPassword`, `signInWithIdToken`, `signInWithWeb3`, `linkIdentityWithIdToken`, `refreshSession`, `setSession` and `recoverSession` on `AuthClient`.
- `AuthPasskeyApi.verifyAuthentication`, and with it the `supabase_flutter` extensions `signInWithPasskey` and `signInWithRestoreKey`.
- The internal refresh path (`_refreshAccessToken`, `_callRefreshToken`, `_doRefresh` and the pending refresh completers) carries a `Session` as well.

A token response without a session now throws `AuthException('The server response did not contain a session.')` from one private `_sessionFromResponse` helper, replacing the three `throw` then `!` blocks and the two methods that silently resolved with `session: null`.

### Kept on `AuthResponse`

`signUp` and `verifyOTP`, since both can legitimately complete without a session: a sign-up that needs email confirmation first, and the first step of a secure email or phone change. The `AuthResponse` dartdoc now says so.

### Server behaviour checked

Verified against the gotrue source that the tightened methods cannot get a 200 without a token: `SignupAnonymously`, the password grant and `PasskeyAuthenticationVerify` all end in `issueRefreshToken` and `sendJSON(w, http.StatusOK, token)`.

## Tests and docs

- Test call sites read the session off the return value directly; `expect(x.session, isNotNull)` assertions that the type now guarantees are dropped, and the `supabase_test` `signInTestUser` helper loses its `!`.
- `MIGRATION.md` gains a section listing the affected methods with a before and after snippet.
- The `supabase_flutter` README social sign-in snippets and the `supabase_auth` example return `Session`.

No public symbols are added or removed, so `sdk-compliance.yaml` is unchanged.

## Out of scope

Making `AuthState` a sealed class and folding `AuthChangeEvent` into it. Much larger blast radius on `onAuthStateChange`, discussed separately.

## Summary by CodeRabbit

- **API Updates**
  - Authentication methods that establish a session now return `Session` directly, including sign-in, refresh, recovery, passkey, Web3, restore credential, and session management flows.
  - These methods throw an authentication error when the server response does not include a session.
  - `signUp()` and `verifyOTP()` continue returning `AuthResponse` because they may complete without creating a session.

- **Documentation**
  - Updated authentication examples and API descriptions to reflect direct `Session` responses.


QuintinWillison (approved):
I don't really understand what I'm looking at, but is `refactor` the right nature of this change, because to my untrained (in Flutter/Dart) eyes the `AuthResponse` to `Session` rename looks breaking in nature... or is this explained by it being part of a wider refactor you're in the middle of (wider context)?

spydon:
> I don't really understand what I'm looking at, but is `refactor` the right nature of this change, because to my untrained (in Flutter/Dart) eyes the `AuthResponse` to `Session` rename looks breaking in nature... or is this explained by it being part of a wider refactor you're in the middle of (wider context)?

It is a breaking refactor, that is what the `!` in the conventional commit standard indicates.

Vinzent03:
Let's say I sign up, but get no session because confirmation is required. What would be the result of signing in with password into that user? Would I get a user back from the server, which would throw an exception now because no session exists?

spydon:
> Let's say I sign up, but get no session because confirmation is required. What would be the result of signing in with password into that user? Would I get a user back from the server, which would throw an exception now because no session exists?

`signInWithPassword` will get a 400 with `error_code: email_not_confirmed`, and the client will throw an `AuthApiException` with `errorCode == ErrorCode.emailNotConfirmed.code` and status 400. So this behavior does not change, it's the same as it was for v2.

Vinzent03 (approved):

PR #1850, opened 2026-09-17
fix(postgrest): build the typed transform builder with the row converter

spydon:
## Summary

Main does not compile since #1848 and #1849 merged within minutes of each other. #1849 replaced the typed transform builder's `PostgrestTable` field with a `RowConverter<Row>` (`_rowFromJson`), and #1848 added `stripNulls`, `dryRun`, `maxAffected` and `csv` on top of the old field. Neither PR's CI saw the other, so every check on main and on the open PRs fails at compile time in `postgrest_typed_transform_builder.dart`:

```
Error: The getter '_table' isn't defined for the type 'PostgrestTypedTransformBuilder<Row, T>'.
```

## Changes

- The four methods from #1848 pass `_rowFromJson` to the private constructor, like the rest of the class does after #1849.
- The two tests from #1848 that passed raw maps use `BookInsert` and `BookUpdate`, which #1849 made the only accepted payload types on the typed builder.
- The `dryRun` dartdoc example uses `BookInsert` as well.

No public API change. `dart analyze` on `packages` and `examples` is clean, `typed_query_test.dart` passes (49), DCM is clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved typed mutation operations so options such as dry runs, affected-row limits, CSV responses, and null stripping consistently preserve the expected result types.
  - Updated typed insert and update usage in mutation scenarios to ensure submitted data is handled with the correct typed representations.

- **Documentation**
  - Updated the dry-run example to demonstrate usage with a typed insert value.


QuintinWillison (approved):
😬

